PR #5, opened 2016-05-05
Fix missing bracket

jerguslejko:
Just add a closing bracket to the if statement in an example in the `stubs/SampleValetDriver.php`


taylorotwell:
Thanks!

PR #6, opened 2016-05-05
Add StaticValetDriver for static pages

jerguslejko:
It would be really nice if we could use _valet_ even for some static sites. This pull request adds an option for that. Assuming that you have an `index.html` file in your project, you're ready to use it with _valet_ awesomeness.

### DIFF
--- a/drivers/StaticValetDriver.php
+++ b/drivers/StaticValetDriver.php
@@ -1,0 +1,43 @@
+<?php
+
+class StaticValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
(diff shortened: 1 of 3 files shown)


taylorotwell:
I made some tweaks to this that I think make it work a tad better: https://github.com/laravel/valet/commit/c29429ba60dff237bb3d5cd33d5d3dbb97045eaa

Can you check them out and see what you think? So you can have the nested directory index.html behavior, etc.

jerguslejko:
Hmmm, I did not think about that. Yes, I definitely like your one better! thanks :)

PR #10, opened 2016-05-05
Add Katana Valet Driver

themsaid:
Not sure if you're familiar with Katana (https://github.com/themsaid/katana), it's a 2 month old Blade powered static site & blog generator.

### DIFF
--- a/drivers/KatanaValetDriver.php
+++ b/drivers/KatanaValetDriver.php
@@ -23,6 +23,6 @@ class KatanaValetDriver extends StaticValetDriver
      */
     public function serves($sitePath, $siteName, $uri)
     {
-        return is_dir($sitePath.'/public');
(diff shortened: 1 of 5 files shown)


drivers/KatanaValetDriver.php
```diff
@@ -0,0 +1,28 @@
...
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return void
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return is_dir($sitePath.'/public');
```
taylorotwell: This will be too generic I think. Is there a more specific check we can make?

themsaid: Modified, it'll now check for the existence of a `katana` file.

PR #13, opened 2016-05-05
Add Generic Driver to point to a public/ folder

svpernova09:
Adds Generic Support to a `public/` folder containing an `index.php` file.  Adds support for SlimPHP, Non-Framework PHP apps.

### DIFF
--- a/drivers/GenericPublicValetDriver.php
+++ b/drivers/GenericPublicValetDriver.php
@@ -1,0 +1,51 @@
+<?php
+
+class GenericPublicValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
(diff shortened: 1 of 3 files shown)


igorpan:
Wouldn't this block any other more strict Driver from working? Laravel one for example?

svpernova09:
Possibly, it may be the default driver will work as a generic

CHH:
I think something like this should be included in Valet. When I installed Valet, I expected plain php files to just work out of the box, but was surprised that it didn't work.

CHH:
We could also think about something like a `Valetfile`, which can configure most common options, like public directory, front controller and an override for the driver.

svpernova09:
@igorpan Turns out as long as it's in the right order it won't conflict.

PR #14, opened 2016-05-05
Add Kirby Driver

pedroborges:
Learn more about Kirby: http://getkirby.com


JonasDoebertin:
Was about to write a Kirby driver myself. Glad I check back here beforehand 😉

pedroborges:
Updated this PR at #42.

PR #15, opened 2016-05-05
Symfony2/3 Valet driver

igorpan:
This driver should work for both Symfony V2 and V3.

It could be split in two separate drivers if necessary

### DIFF
--- a/drivers/SymfonyValetDriver.php
+++ b/drivers/SymfonyValetDriver.php
@@ -1,0 +1,49 @@
+<?php
+
+class SymfonyValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return bool
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return file_exists($sitePath.'/web/app_dev.php') &&
+               (file_exists($sitePath.'/bin/symfony_requirements') ||
+                file_exists($sitePath.'/app/SymfonyRequirements.php'));
+    }
+
+    /**
+     * Determine if the incoming request is for a static file.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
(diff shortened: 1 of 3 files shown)


taylorotwell:
Nice thanks for give it a try later tonigiht.

CHH:
Would be nice to make the front controller somehow configurable. Sometimes one needs to run Symfony apps using the prod front controller.

drivers/SymfonyValetDriver.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return bool
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return file_exists($sitePath.'/web/app_dev.php') &&
+               (file_exists($sitePath.'/bin/symfony_requirements') ||
+                file_exists($sitePath.'/app/SymfonyRequirements.php'));
```
WouterJ: -1 on these 2 checks. These are just default conventions, nothing forbits you to change the Standard edition directory structure.

igorpan: But nothing forbids you from changing anything in Symfony, really. You don't need to have a web directory at all. If you go away from it's defaults, you would have to create your own driver either way.

igorpan: Maybe we could rename this to "SymfonyStandardEditionValetDriver" to be more specific. I don't think it makes sense making anything more generic than this.

taylorotwell:
Tweaks checks a bit and made it so that if "app_dev" doesn't exist it will serve from "app", so that you can at least run through the production front controller if you need to by renaming "app_dev" to "app_dev_bkp" or something. Thanks!

drivers/SymfonyValetDriver.php
ruudk: Just installed Valet on a Symfony2 project without the 2 files. Didn't work. Took me a while to find that this was causing it.

PR #16, opened 2016-05-06
Ensure domain is read from the configuration file for dnsmasq

michaeldyrynda:
Whilst you can configure the domain via `config.json`, the domain is not read from the configuration file when writing to `dnsmasq.conf`

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -27,7 +27,7 @@ class Configuration
         }
 
         if (! file_exists(static::path())) {
-            static::write(['domain' => 'dev', 'paths' => []]);
+            static::write(['paths' => []]);
         }
 
         chown(static::path(), $_SERVER['SUDO_USER']);
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -7,9 +7,10 @@ class Configuration
     /**
      * Install the Valet configuration file.
      *
+     * @param  string  $domain
      * @return void
      */
-    public static function install()
+    public static function install($domain)
     {
         if (! is_dir($directory = VALET_HOME_PATH)) {
             mkdir($directory, 0755);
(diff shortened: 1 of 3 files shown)


taylorotwell:
So I'm guessing you update domain in config and then run `valet install` again?

michaeldyrynda:
That's currently the only way to go about it. I didn't want to introduce any extra CLI arguments. I could modify this to do something like `valet install <domain=dev>` if you'd like?

michaeldyrynda:
image

Is this sufficiently zonda? 😀

adamwathan:
I think we're going to want to do something to mark any `/etc/resolver/*` files we create so we can clean them up if someone wants to use a different TLD. Right now if you do a normal `valet install` and then a `valet install local` afterwards, it leaves `/etc/resolver/dev` behind, and we probably want to delete that if we know we're the ones who created it. Pow does this [with a comment](https://github.com/basecamp/pow/blob/master/src/templates/installer/resolver.eco#L1) so that it's easy for them to know if they can safely delete that file.

It looks like this PR is not creating a new `/etc/resolver/{tld}` yet for the user specified TLD either, so will need to do that.

Ultimately I wonder if we want to also add a `valet tld {tld}` command that just swaps this stuff, seems overkill to rewrite all the launch daemons and everything unnecessarily. I do like being able to specify it on install as well though.

michaeldyrynda:
Yeah, still a WIP as I discovered some issues (like `dnsmasq.conf` being overwritten, but `config.json` isn't). I just tweaked that so that the domain is merged from the command.

Another issue i've found is that dnsmasq gets a little sticky if you change the TLDs around. Restarting dnsmasq and flushing cache doesn't do the trick; have needed to restart for a changed TLD to stick. There *must* be some command to get it to work, but I'm yet to find it. Will work some more on this tomorrow; it's late.

adamwathan:
Late? It's only 10:09 AM man 😉

taylorotwell:
Did this a little different. Fixed a few bugs and made sure that it also applies when using Valet share.

michaeldyrynda:
Thanks Taylor :)

PR #20, opened 2016-05-06
Fix Wordpress driver when you use permalink

blackburn1911:
- fix 404 not found


taylorotwell:
I fixed it. This wasn't the right fix.

PR #29, opened 2016-05-06
Jekyll Valet Driver

slashequip:
Because sometimes `localhost:4000` is ~~ugly~~ not Zonda.

Use with `$ jekyll build --watch` for all the recompiling loveliness on a beautiful valet'd url.

### DIFF
--- a/cli/drivers/JekyllValetDriver.php
+++ b/cli/drivers/JekyllValetDriver.php
@@ -15,8 +15,6 @@ class JekyllValetDriver extends ValetDriver
         //Check for several default Jekyll folders and files.
         return 
             is_dir($sitePath.'/'.$this->getServingFolderName())
-            && is_dir($sitePath.'/_sass')
-            && is_dir($sitePath.'/_includes')
             && file_exists($sitePath.'/_config.yml')
             && file_exists($sitePath.'/'.$this->getServingFileName());
     }


adamwathan:
Just testing this out with my own Jekyll blog but not catching it because my site has no `.jekyll-metadata` file. Just tried creating a new Jekyll site too (using Jekyll 3.1.1) and that file doesn't exist. Is there another way you know of to reliably determine if a site is a Jekyll site?

scrwdnet:
Even after running `jekyll build`? Sorry if that's a stupid question!

Just doing some testing myself now.

adamwathan:
Yep! Seems weird after reading the docs but its true :/

adamwathan:
image

scrwdnet:
You are right, `.jekyll-metadata` was removed in 3.0.1 of Jekyll, I was running 3.0.0

If this driver was to rely a little on the default Jekyll setup does that present an issue? e.g. reading the first line of `_config.yml` file and check for the welcome text or even just 'Jekyll'? 

On another note I was relying on the presence of `_site` this technically can be changed on build using `jekyll build -d ./_foo` so, again, it's not something that can be relied upon.

scrwdnet:
Ergh if you run `jekyll build --incremental` you'll get the `.jekyll-metadata` file.

adamwathan:
Yeah was thinking about reading the config file as well. Hmm... Maybe just check for the presence of a combination of folders/files that make it extremely likely to be a Jekyll site and stick it near the end of the driver stack?

scrwdnet:
Sounds good, I'll make a couple of changes now.

scrwdnet:
This should cover most Jekyll builds (especially the default) and it is at the end of the driver stack now.
Checking for `_site`, `_sass`, `_includes`, `_config.yml`, `index.html`. The only check that could potentially change going forward would be `_site`. As previously mentioned this can be changed by the user but I imagine most don't even look at changing that directory?!

scrwdnet:
Upon fresh eyes the fact that the `_site` directory can be changed is irrelevant as it that is changed the front controller directory is then broken so the `_site` directory check is actually a good thing!

scrwdnet:
In light of new thoughts, have merged master into the PR and removed a couple of folder checks, this should be good to go now.

kevinkhill:
:+1:  I almost started writing this driver! Glad I checked first before going forward

TuckerWhitehouse:
I don't think checking for `index.html` is valid or necessary. The page made available at the root for a jekyll site can be manipulated using the `permalink`.

itguy614:
I have a much simpler approach for a Jekyll Driver.  I wrote it up before looking for existing pull requests.  Seems to work for me okay.  It assumes that the built Jekyll site is in a _site folder, after that it just serves up that folder as all static content.

```
class JekyllValetDriver extends BasicValetDriver
{
    /**
     * Determine if the driver serves the request.
     *
     * @param  string  $sitePath
     * @param  string  $siteName
     * @param  string  $uri
     * @return bool
     */
    {
        if (file_exists($sitePath.'/_site')) {
            return true;
        }

        return false;
    }

    /**
     * Mutate the incoming URI.
     *
     * @param string $uri
     * @return string
     */
    public function mutateUri($uri)
    {
        return rtrim('/_site' . $uri, '/');
    }

}
```

mattstauffer:

Thanks!

PR #31, opened 2016-05-06
Add support for Drupal 6, 7 and 8

Cybnext:
Tested from installation phase, to creating content and navigating through the admin interfaces. All parts are functional.


taylorotwell:
Can't even get past install on Drupal 8. Please **only** submit once this is acutally tested:

http://d.pr/i/C5hK/2p4g1ebR

PR #32, opened 2016-05-07
Added Magento 2 driver

sschlein:
Just because it works (and I like Zend people at the Laravel docs looking how to install Valet) :-)

Magento composer install I've used for testing the driver:
http://devdocs.magento.com/guides/v2.0/install-gde/prereq/integrator_install.html

Magento was quite tricky because they serve their static assets via a php file - but they had a solution for this in their included phpserver.

I also updated my branch to the newest Valet version and squashed all my commits into one for easy review.

### DIFF
--- a/drivers/MagentoValetDriver.php
+++ b/drivers/MagentoValetDriver.php
@@ -1,0 +1,77 @@
+<?php
+
+class MagentoValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return void
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return file_exists($sitePath . '/bin/magento') && file_exists($sitePath.'/pub/index.php');
+    }
+
+    /**
+     * Determine if the incoming request is for a static file.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return string|false
+     */
+    public function isStaticFile($sitePath, $siteName, $uri)
+    {
+        $route = parse_url(substr($uri, 1))["path"];
+
+        if (strpos($route, 'pub/errors/default/') === 0) {
+            $route = preg_replace('#pub/errors/default/#', 'errors/default/', $route, 1);
+        }
+
+        if (
+            strpos($route, 'media/') === 0 ||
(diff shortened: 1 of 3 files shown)


drivers/MagentoValetDriver.php
```diff
@@ -0,0 +1,77 @@
...
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return void
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return file_exists($sitePath . '/bin/magento');
```
barryvdh: Perhaps also Check the pub/index.php

barryvdh:
I think you squashed to many changes in here, or used an outdated version.

sschlein:
Cleaned up the commits and also added your suggestion regarding the double-check if it's really Magento2.

barryvdh:
And make it Magento2 driver, to make it clear it's just Magento 2? (And/or add Magento1 if you want? :P)

barryvdh:
Not sure if magento2 is very different in the handling of static files. Magento1 only uses htdocs/index.php instead of pub/index.php

sschlein:
The problem with Magento 1 is that there are many ways of running the application. htdocs might be named www and the whole application structure is based on the skills of the developer team. 

I tried to run Magento1 with Valet first but the index.php rewrites the document root and does several other things... 

I'd just mention Magento 2 compatibility in the docs - maybe someone has an idea how to run Magento 1 and could upgrade this in the future.

mpociot:
Awesome seeing this happen! Setting up Magento in a VM is always a mess and slow as hell! Great.

raffw7912:
Hi, I'm using my own driver for Magento 1 and it seems to be working fine. In `serves` method I'm checking for `file_exists($sitePath.'/app/Mage.php')` then `isStaticFile` is the same as in sample class. `frontControllerPath` is just `return $sitePath.'/index.php';` and it works! I have a big, custom project running this way with no issues.

sschlein:
Hi @raffw7912 
if you simply return the normal index.php, your site should work with Valet out of the box and you don't need a custom driver. Maybe my Magento 1 project has a higher complexity and also is an EE Magento - whatever, it just doesn't work ;-)

taylorotwell:
Can't get past this screen:

http://d.pr/i/1dEXc/4WnvFkE0

taylorotwell:
Re-submit with instructions on how to run a Magento installation.

PR #33, opened 2016-05-07
Sites Manifest Page

furey:
Displays a simple sites manifest page via "http://[manifest].[domain]" (defaults to "http://valet.dev").

valet-manifest

The manifest name is configurable via the "manifest [manifest]" command.

valet-manifest-command

The manifest URL can be checked via the "manifest-url" command.

valet-manifest-url-command

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -1,0 +1,66 @@
+<?php
+
+$domain = $valetConfig['domain'];
+
+$paths = [];
+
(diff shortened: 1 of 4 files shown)


taylorotwell:
Holding off on this for now because we have something in the pipeline for this.

furey:
Ah, great, can't wait. Cheers!

PR #34, opened 2016-05-07
Update Caddyfile

abiosoft:
The regexp is can be ommited as it is a wildcard. 
`{path}` and `{path}/` ensures valid files (especially static files) and directories are served.

### DIFF
--- a/stubs/Caddyfile
+++ b/stubs/Caddyfile
@@ -5,6 +5,5 @@ fastcgi / 127.0.0.1:9000 php {
 }
 
 rewrite {
-    r .*
-    to /server.php?{query}
+    to {path} {path}/ /server.php?{query}
 }


stubs/Caddyfile
```diff
@@ -5,6 +5,5 @@ fastcgi / 127.0.0.1:9000 php {
 }
 
 rewrite {
-    r .*
-    to /server.php?{query}
+    to {path} {path}/ /server.php?{query}
```
mholt: The `to` change may not be quite what is needed in Valet's case, but I do believe the `r .*` is redundant, since by default, `rewrite` matches all requests.

taylorotwell:
I don't think we actually want Caddy to serve static files. We don't know *which* static files to serve at this point in the lifecycle. We're mainly using Caddy as a way to simple send every incoming request to FPM. The `server.php` script will then serve static assets using the same logic as PHP's own built-in web server.

taylorotwell:
Would you mind tweaking this change to only remove the regular expression check, as that is redundant as Matt noted. Thanks!

taylorotwell:
Closing this as I'm pushing a few related updates.

abiosoft:
Yeah, fine.

I didn't realise that until I had a chat with Matt.

PR #36, opened 2016-05-07
Add phpMyAdmin driver

BePsvPT:


BePsvPT:
phpMyAdmin now is already support by BasicValetDriver

PR #38, opened 2016-05-07
Add simple commands to manage caddy's server blocks

dchristophersen:
With Caddy we have the possibility to use multiple server blocks and @taylorotwell used the power of zonda to create a Caddyfile that allows us to add custom config files to ~/.valet/Caddy/. Because I wanted Valet to be my entrypoint to all local projects, I added two commands to manage custom server blocks:

`valet serve domain serverblock [-d | --data=]*`  : Uses a stub file from ~/.valet/ServerBlocks/{serverblock}.conf or {VALET_INSTALL}/stubs/ServerBlocks/{serverblock}.conf to add a new config file to the ~/.valet/Caddy/ folder. Some basic variables like "DOMAIN" or "CWD" will be replaced by default.
All other variables in the stub file can be replaced by providing the -d option with key-value data e.g. -d GIT_REPO=git@github.com/laravel/laravel.git .

and

`valet unnerve domain` : Removes .conf file from the Caddy folder and restarts Valet.

With this commands I managed to add all my bigger projects, running in my homestead box, fluently with an simple `valet serve {domain} homestead`. 

I also tested the git directive, custom caddy binary needed, and a custom fastcgi directive. That all worked perfectly for me even after changing the tld of Valet.

### DIFF
--- a/src/ServerBlock.php
+++ b/src/ServerBlock.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Valet;
+
+class ServerBlock
+{
+    public static function install()
+    {
+        if (! is_dir($serverBlocksDirectory = VALET_HOME_PATH.'/ServerBlocks')) {
+            mkdir($serverBlocksDirectory, 0755);
+
+            chown($serverBlocksDirectory, $_SERVER['SUDO_USER']);
+        }
+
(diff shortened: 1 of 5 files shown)


taylorotwell:
No plans on including this in core.

PR #39, opened 2016-05-07
Remove Automatic Linking of storage/public files

igormatkovic:
The storage/public folder is usually not symlinked to the public/ folder.
In fact it's impossible to symlink it unless you move the main index.php file.

A specific issue I'm having here is that folders in my storage/public directory actually match resource URL's

Example:

Users Controller points to: **example.dev/users**
Users uploaded files: **storage/public/users**

### DIFF
--- a/drivers/LaravelValetDriver.php
+++ b/drivers/LaravelValetDriver.php
@@ -30,10 +30,6 @@ class LaravelValetDriver extends ValetDriver
             return $staticFilePath;
         }
 
-        if (file_exists($sitePath.'/storage/public'.$uri)) {
-            return $sitePath.'/public'.$uri;
-        }
-
         return false;
     }
 


taylorotwell:
I don't think this should be removed, I just think that a ! is_dir check should be added to it. Also, I don't know what you mean by it's not usually symlinked. It's common to create a symlink at public/storage pointing to storage/app/public.

I've made the necessary changes.

igormatkovic:
I agree with what you are saying @taylorotwell, but I'm not sure you are looking at this correctly.

The check that is in the valet driver is not looking for public/storage its looking in public only.
For the code to work like you just said it should look like this then:
```php
 if (file_exists($sitePath.'/storage/public'.$uri)) {		
            return $sitePath.'/public/storage/'.$uri;		
}
```

Also. If a directory is symlinked. The first check should have already found the files in the previous check.

There is no need for the secondary assumption on how things are symlinked.

PR #43, opened 2016-05-09
Escape file paths for 'logs' command

mattdfloyd:
Use escapeshellarg to make file paths safe to use in shell command


mattdfloyd:
`tail` will choke on directories with or certain characters that are not escaped. For example, my parked directory is `~/Dropbox (Personal)/Sites/`.

PR #45, opened 2016-05-09
WordPress Driver: Fix undefined offset error in wp-admin

aaemnnosttv:
This fixes an error which is produced in wp-admin.  Although wp-admin seems to work fine already, it is silently triggering this error as WordPress does not show/log errors by default.  You can see this by adding `define('WP_DEBUG', true);` to `wp-config.php`

image

Basically, WP uses the `PHP_SELF` sever global to determine which admin page it is on, which is currently `server.php` everywhere.

`/wp-includes/vars.php`
image

This PR ensures that the global is defined in a way which should work with what WP expects.

### DIFF
--- a/drivers/WordPressValetDriver.php
+++ b/drivers/WordPressValetDriver.php
@@ -14,4 +14,19 @@ class WordPressValetDriver extends BasicValetDriver
     {
         return is_dir($sitePath.'/wp-admin');
     }
+
+    /**
+     * Get the fully resolved path to the application's front controller.
+     *
+     * @param  string  $sitePath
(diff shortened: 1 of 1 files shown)


ramon-villain:
thank you! 

had the same issue.

sixlive:
This also fixes a major issue accessing plugins that utilizes custom admin pages (sub-menu pages in particular, top-level are ok). You will end up getting stuck with `You do not have sufficient permissions to access this page.`.

kevdotbadger:
Oh thank god!

PR #47, opened 2016-05-09
Added Magento 2 driver

sschlein:
To test this driver, run the following commands. Please make sure to have an empty database "magento" which can be accessed with the root user without a password on localhost.

`composer create-project --repository-url=https://repo.magento.com/ magento/project-community-edition magento`

`cd magento`

`chmod a+x bin/magento`

```
bin/magento setup:install --base-url=http://magento.dev/ \
--db-host=localhost --db-name=magento --db-user=root --db-password= \
--admin-firstname=Magento --admin-lastname=User --admin-email=user@example.com \
--admin-user=admin --admin-password=admin123 --language=en_US \
--currency=USD --timezone=America/Chicago --use-rewrites=1
```

### DIFF
--- a/drivers/MagentoValetDriver.php
+++ b/drivers/MagentoValetDriver.php
@@ -1,0 +1,77 @@
+<?php
+
+class MagentoValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return void
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        return file_exists($sitePath . '/bin/magento') && file_exists($sitePath.'/pub/index.php');
+    }
+
+    /**
+     * Determine if the incoming request is for a static file.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return string|false
+     */
+    public function isStaticFile($sitePath, $siteName, $uri)
+    {
+        $route = parse_url(substr($uri, 1))["path"];
+
+        if (strpos($route, 'pub/errors/default/') === 0) {
+            $route = preg_replace('#pub/errors/default/#', 'errors/default/', $route, 1);
+        }
+
+        if (
+            strpos($route, 'media/') === 0 ||
+            strpos($route, 'opt/') === 0 ||
+            strpos($route, 'static/') === 0 ||
+            strpos($route, 'errors/default/css/') === 0 ||
+            strpos($route, 'errors/default/images/') === 0
+        ) {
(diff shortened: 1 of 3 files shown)


lazyguru:
You should consider naming this Magento2ValetDriver instead (just for clarity that it only handles M2 installs).

barryvdh:
+1 for renaming to Magento2 for clarity, as Magento1 is still supported until 2018.

I've tested this driver and it seems to work for the front-end + admin at least. I installed using Composer + CLI (http://devdocs.magento.com/guides/v2.0/install-gde/install/cli/install-cli-install.html), so not via /setup.

barryvdh:
Also this has conflicts, so needs to be updated anyways :)

LasseRafn:
Cool! Any status on this?

lazyguru:
For anyone still needing this, I've been using https://gist.github.com/lazyguru/686dd5673097b7db1483dc5c817bd4ee for the last few weeks for local dev with no issues.  This PR's version seems geared at working with a site in production mode (hosted from pub directory).  Mine runs from Magento root instead (for developer mode).  If you diff my version from the SampleDriver.php you'll see it's not very different from it (mostly just removes reference to the "/public" path from URLs and then logic to detect the framework.

LasseRafn:
Lovely! I'll give it a shot :-) I had to switch to Vagrant while waiting for this, so it's very much appreciated!

lazyguru:
I updated my gist to be able to handle for either developer mode or production (treats default as if it is production). @sschlein if you want to update this PR with my code that is fine, otherwise I can submit a new PR with my version.  https://gist.github.com/lazyguru/686dd5673097b7db1483dc5c817bd4ee

sschlein:
A new by @lazyguru implementation is available

beningreenjam:
Brand new install of Magento 2.1 and Valet 2.

Magento's static assets are all returning a 404 not found error.

lazyguru:
What do you have MAGE_MODE set to in app/etc/env.php?  If it is set to anything other than developer you need to run the static content deploy using bin/magento

I am using this same driver with 2.1.3 and Valet 2 with no issues.

lazyguru:
Just realized which PR this is. Which driver are you talking about? Mine or @sschlein's?

beningreenjam:
@lazyguru your driver. I've tried in all in 3 deployment modes but it's my understanding that the static content on production mode which is done automatically when switching to production mode.

beningreenjam:
I'm also using Magento 2.1 and Valet 2.

lazyguru:
You might have to adjust some of the timeouts in nginx php-fpm. I proxy my requests using ngrok (or valet share). The first time it generates the symlinks can take a long time to process. Also, lets move to #147 since this pertains to my driver

PR #49, opened 2016-05-10
Add Bedrock driver

aaemnnosttv:
Adds a driver for the [Roots Bedrock](https://github.com/roots/bedrock) modern WordPress stack.

Install:
```
composer create-project roots/bedrock
cd bedrock
cp .env.example .env
```
1. Update `.env` with DB creds
1. Set `WP_HOME` to `http://bedrock.dev` in `.env`
1. 👍


taylorotwell:
Sweet! Will give it a shot in a bit.

fishtankmike:
Having trouble with static files now that its been added to valet core eg...

cant access /web/older/file.php

Just get the WordPress 404 response.

PR #50, opened 2016-05-10
Open website in your browser

RobinMalfait:
When you are in a folder for example `/Volumes/Data/Sites/RobinMalfait` you can run `valet open` and it will open `http://robinmalfait.dev/`.

If you have another tld configured for example `.app` it will open `http://robinmalfait.app/`


DemianD:
Would be nice !

RobinMalfait:
Merged the remote-upstream so the file `valet.php` is moved to `cli/valet.php`. Can be merged normally now :)

taylorotwell:
Fixed so it works with linked sites: https://github.com/laravel/valet/commit/62ba72a77a319986bc685dbf4bc9818e9a7aa457

PR #52, opened 2016-05-10
Added a driver for single page applications.

tlaverdure:
Wanted to use valet to run a few single page apps that I've built. This mainly handles the redirection of all urls to the index.html file so the app can handle routing. Otherwise, you will always end up on the 404 page.

The serves method checks to see if the there is a package.json file, a node_modules directory, and if there is an index.html file in either the root directory or a /dist directory. The driver could also be enabled by adding this property to the app's package.json file:

``` json
"valet_driver": "spa",
```

### DIFF
--- a/drivers/SpaValetDriver.php
+++ b/drivers/SpaValetDriver.php
@@ -1,0 +1,101 @@
+<?php
+
+class SinglePageApplicationValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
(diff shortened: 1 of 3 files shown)


est73:
The Ubuntu fork of Valet has this PR merged and it was causing a fatal error.
https://github.com/cpriego/valet-ubuntu/pull/2/files

taylorotwell:
Will let this be a custom driver.

PR #56, opened 2016-05-10
Add driver for Wordpress installed as a Composer dependancy

corneliusio:
Wordpress can be installed as a dependance using the "johnpbloch/wordpress" package. This driver should account for the idiosyncrasies involved with that setup (and likely variations).

EDIT: I noticed after I created this that there is already pull request #49 for a Bedrock driver, this driver will now account for a bedrock install as well as more generic setups.

### DIFF
--- a/cli/drivers/WordpressComposerInstallValetDriver.php
+++ b/cli/drivers/WordpressComposerInstallValetDriver.php
@@ -13,7 +13,6 @@ class WordpressComposerInstallValetDriver extends ValetDriver
     public function serves($sitePath, $siteName, $uri)
     {
         if (file_exists($composerFile = $sitePath .'/composer.json')) {
-
             if (file_exists($sitePath .'/wp-config.php')) {
                 return true;
             }
@@ -61,7 +60,7 @@ class WordpressComposerInstallValetDriver extends ValetDriver
     {
         if (is_dir($staticFilePath = $sitePath . $uri)) {
             return $sitePath . $uri . '/index.php';
-        } else if ($this->fileIsPHP($uri)) {
+        } elseif ($this->fileIsPHP($uri)) {
             $_SERVER['PHP_SELF'] = $uri;
             return $sitePath . preg_replace('/\/$/', '', $uri);
         }
--- a/cli/drivers/WordpressComposerInstallValetDriver.php
+++ b/cli/drivers/WordpressComposerInstallValetDriver.php
@@ -14,7 +14,8 @@ class WordpressComposerInstallValetDriver extends ValetDriver
     {
         if (file_exists($sitePath . '/composer.json')
             && (file_exists($sitePath . '/wp-config.php')
-                || is_dir($sitePath . '/vendor/johnpbloch/wordpress-core-installer'))
+                || $this->isJohnpblochInstall($sitePath)
+                || $this->isWebrootInstall($sitePath))
         ) {
             return true;
         }
@@ -37,8 +38,16 @@ class WordpressComposerInstallValetDriver extends ValetDriver
             $sitePath .= '/web';
         }
 
-        if (file_exists($staticFilePath = $sitePath . $uri) && !$this->fileIsPHP($uri) && !is_dir($staticFilePath)) {
-            return $staticFilePath;
+        if ($this->getStaticFilePath($sitePath, $uri)) {
+            return $this->getStaticFilePath($sitePath, $uri);
+        }
+
+        if ($this->isWebrootInstall($sitePath)) {
+            $sitePath = $this->getWebrootInstallPath($sitePath);
+
+            if ($this->getStaticFilePath($sitePath, $uri)) {
+                return $this->getStaticFilePath($sitePath, $uri);
+            }
         }
 
         return false;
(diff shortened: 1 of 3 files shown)


cli/drivers/WordpressComposerInstallValetDriver.php
```diff
@@ -0,0 +1,75 @@
...
+     * Determine if the incoming request is for a static file.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return string|false
+     */
+    public function isStaticFile($sitePath, $siteName, $uri)
+    {
+        if (file_exists($staticFilePath = $sitePath . $uri) && !$this->fileIsPHP($uri) && !is_dir($staticFilePath)) {
```
KennedyTedesco: Could be:

```
if ($this->isActualFile($staticFilePath = $sitePath.$uri) && ! $this->fileIsPHP($uri)) {
```

cli/drivers/WordpressComposerInstallValetDriver.php
```diff
@@ -0,0 +1,75 @@
...
+     * Get the fully resolved path to the application's front controller.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return string
+     */
+    public function frontControllerPath($sitePath, $siteName, $uri)
+    {
+        if (is_dir($staticFilePath = $sitePath . $uri)) {
```
KennedyTedesco: Could be:

```
if (is_dir($path = $sitePath.$uri)) {
	return $path.'/index.php';
```

cli/drivers/WordpressComposerInstallValetDriver.php
corneliusio: Tried `$this->isActualFile()` originally. For whatever reason, it throws a Fatal error: Call to undefined method for me.

Also, the `!is_dir($staticFilePath)` check is needed for any path that doesn't explicitly load the index.php such as as the dashboard landing (i.e. example.com/wp-admin/)

cli/drivers/WordpressComposerInstallValetDriver.php
corneliusio: I did miss this. Thanks for pointing it out.

cli/drivers/WordpressComposerInstallValetDriver.php
KennedyTedesco: Got it.

cli/drivers/WordpressComposerInstallValetDriver.php
```diff
@@ -0,0 +1,105 @@
...
+     * @return void
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        if (file_exists($composerFile = $sitePath . '/composer.json')) {
+            if (file_exists($sitePath . '/wp-config.php')) {
+                return true;
+            }
+
+            $composerJson = json_decode(file_get_contents($composerFile));
```
aaemnnosttv: `json_decode()` on every request seems a bit inefficient to me.

How about `is_dir($sitePath . '/vendor/johnpbloch/wordpress-core-installer/')` instead?

corneliusio: Excellent idea. Tried to avoid it by looking for wp-config.php first, but this makes way more since.

The only catch I see with that is the rare case where someone might define a custom vendor directory for composer AND put wp-config.php in a subdirectory. But probably not a common enough scenario to worry about.

Checking for wp-config.php or wordpress-core-installer should hopefully account for like 99.99% of setups.

corneliusio:
Had to do a fair amount of additions and some refactoring, but this driver should now deal with Wordpress installs that use fancyguy/webroot-installer.

cli/drivers/WordpressComposerInstallValetDriver.php
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return string
+     */
+    public function frontControllerPath($sitePath, $siteName, $uri)
+    {
+        // Look in public web directory
+        if ($this->isBedrockInstall($sitePath)) {
```
aaemnnosttv: As a generic driver, this should probably not handle specific cases like Bedrock, IMHO.
Specific implementations are better off with their own driver.

corneliusio: Yeah, really started to consider that during this most recent commit.

Idk, perhaps it would be better to close this pull request and create generic and webroot installer drivers separately. Then you can roll with the your bedrock implementation. But at the same time, they'll still be repeating a lot of the same logic.

I could go either way, shouldn't be any more widely used patterns to account for that I can tell. Open to more input.

corneliusio:
After giving it some thought, I've decided @aaemnnosttv is right and this driver is a little too broad. Going to close this pull request as to not muddy things up with all the commits and create separate drivers for webroot installs and a generic (johnpbloch/wordpress & index.php at root) install. #49 can handle Roots Bedrock instances.

PR #59, opened 2016-05-11
Add fancyguy & johnpbloch flavor Wordpress drivers

corneliusio:
[johnpbloch/wordpress](https://github.com/johnpbloch/wordpress) and [fancyguy/webroot-installer](https://github.com/fancyguy/webroot-installer) are the two most common (only two I've ever seen) non-opinionated methods (vs something such as [roots/bedrock](https://github.com/roots/bedrock)) for installing Wordpress with composer.


taylorotwell:
I'd rather leave these as community drivers and just focus on Bedrock since that seems pretty popular.

corneliusio:
Sounds good. Figured I needed to write it for myself anyway, might as well offer it up to be pre-packaged. 😃

PR #60, opened 2016-05-12
Consistency changes

unfunco:
Some simple changes for consistency and IDE warning suppression.

Let me know if you would prefer these commits squashed.


unfunco:
Solving the important problems. 👍

PR #62, opened 2016-05-13
Fix spaces in path when using 'valet link'

jpscharf:
- Adds helper function `quote ( $string, $quote_character = '"' )`
- Uses `quote()` on `$target` when running `symlinkAsUser($target, $link)`


taylorotwell:
Should we just use this? http://php.net/manual/en/function.escapeshellarg.php

jpscharf:
I removed the custom function and used escapeshellarg(), it appears to work.

PR #64, opened 2016-05-13
Allow HTTPS

Keoghan:
Add :443 definition to Caddyfile, create SSL Cert upon install. Requires a `valet install` to get going and openssl available.

### DIFF
--- a/cli/Valet/Caddy.php
+++ b/cli/Valet/Caddy.php
@@ -27,9 +27,23 @@ class Caddy
      */
     function install()
     {
+        $this->installSSLCertificate();
         $this->installCaddyFile();
         $this->installCaddyDirectory();
         $this->installCaddyDaemon();
+    }
+
+    /**
+     * Make the certificate we're going to use when serving over HTTPS
+     * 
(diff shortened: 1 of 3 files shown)


cli/Valet/Caddy.php
```diff
@@ -27,12 +27,26 @@ function __construct(CommandLine $cli, Filesystem $files)
...
     function install()
     {
+        $this->installSSLCertificate();
         $this->installCaddyFile();
         $this->installCaddyDirectory();
         $this->installCaddyDaemon();
     }
 
     /**
+     * Make the certificate we're going to use when serving over HTTPS
```
GrahamCampbell: Missing fullstop.

taylorotwell:
This doesn't address several issues with SSL we are already working on, such as refreshing all certificates when changing domains. Trusting the certificates etc. We are working on this.

Keoghan:
Fair enough, I agree that those will add to the feeling of the tool, but, from my user standpoint, while the certificate doesn't change when changing domain, it still works, and while it's not a trusted cert, it does work for dev purposes (as it does on Homestead).

PR #65, opened 2016-05-13
Detect WordPress by configuration file

tillkruss:
WordPress supports being placed in [it's own directory](https://codex.wordpress.org/Giving_WordPress_Its_Own_Directory), but in both setups the `wp-config.php` file will be in the site's root directory.

### DIFF
--- a/cli/drivers/WordPressValetDriver.php
+++ b/cli/drivers/WordPressValetDriver.php
@@ -12,7 +12,7 @@ class WordPressValetDriver extends BasicValetDriver
      */
     public function serves($sitePath, $siteName, $uri)
     {
-        return is_dir($sitePath.'/wp-admin');
+        return file_exists($sitePath.'/wp-config.php');
     }
 
     /**


scrwdnet:
The file check makes sense but WordPress also supports the `wp-config.php` file being one level above the serving folder as well - perhaps worth a second check for this to?

sixlive:
I would also make it:

```
{
    return file_exists($sitePath.'/wp-config.php') ||  file_exists($sitePath.'/wp-config-sample.php');
}
```

This way it will allow you to use the configuration gui to setup the new site.

sixlive:
@scrwdnet In that case, I don't think there is really way to detect if its a WordPress site and you'll likely be better serving through the `valet link` functionality.

tillkruss:
Testing for `wp-config.php` works fine with standard WordPress setups:

```
~/Sites/WordPress/wp-config.php
```

And it works with WordPress in sub-directories:

```
~/Sites/WordPress/wp/wp-config.php
```

See: https://github.com/WordPress/WordPress/blob/master/wp-load.php

In all other cases `valet link` is a better choice.

sixlive:
@tillkruss right, but we can't guarantee that everyone is going to use `wp/` as the directory that they nest inside of.

tillkruss:
It doesn't matter what the folder is called:

```
~/Sites/WordPress/wp/wp-config.php
~/Sites/WordPress/wp/wp-config.php
```

sixlive:
@tillkruss I just tested it out, I see what you're saying now.

tillkruss:
It doesn't matter what the folder is called, because you can adjust the ABSPATH in you `wp-config.php` file:

```php
if ( ! defined( 'ABSPATH' ) ) {
	define( 'ABSPATH', __DIR__ . '/wp/' );
}

require_once ABSPATH . 'wp-settings.php';
```

```
~/Sites/WordPress/wp/wp-settings.php
~/Sites/WordPress/wordpress/wp-settings.php
~/Sites/WordPress/yomama/wp-settings.php
```

sixlive:
@tillkruss yeah, once I fired it up to take a look, I saw what you meant.

abdusfauzi:
Great fix! all my work is this setup: WordPress in its own folder as `core`.

Btw, I seems to have issue with Permalink structure, where it still display `../index.php/..` even though I can remove them using the Custom Structure.

Where should I open a ticket?

taylorotwell:
If you have fix it crack open a PR.

ramon-villain:
@abdusfauzi this index.php is about the ```got_url_rewrite()``` right?

Do you have a fix?

abdusfauzi:
@ramon-villain i guess so. No, I don't have a fix. I stumbled this on the Settings > Permalink page. I'm not familiar with how Valet Driver works. Tried to read and modified stuff, but end up error. 😅

tillkruss:
@abdusfauzi Are permalinks working fine and it's just the settings that are weird?

abdusfauzi:
@tillkruss Everything is working fine (permalink). However, previously when using Vagrant based setup (LEMP), the `/index.php/` does not appear between the Common Settings.

So, I'm wondering if this is because how Valet works or Caddy.

PR #70, opened 2016-05-18
Support expected PHP version : 7.x

lucasmichot:
As of https://github.com/laravel/valet/blob/master/cli/scripts/install.sh#L17-L18

Even if it not clear to me why PHP 7.0 is a requirement

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: php
 
-php:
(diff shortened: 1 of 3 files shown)


taylorotwell:
That's just what we install by default. But you can actually use it with lesser versions.

PR #48, opened 2016-05-09
Drupal 7 & 8 Driver

Cybnext:
This driver adds support for Drupal 7 and 8.

PR #79, opened 2016-05-27
Add Support for drupal sites

McGo:
I added a DrupalValetDriver that checks if the drupal js api core file is present either in core/misc (Drupal 8) or misc (Drupal 7 and below).


McGo:
Uups. Did not see, that this is a duplicate of #48

PR #81, opened 2016-05-29
Add support for custom sub-domain for valet share.

ajmalmsali:
Current version uses ngrok only, where custom sub-domain is a paid feature,
  Introduced http://localtunnel.me as an option.
  Usage e.g: valet share subdomain

This feature happened to be a must-have for our project where we wanted the url to be static.

### DIFF
--- a/bin/lt
+++ b/bin/lt
@@ -1,0 +1,62 @@
+#!/usr/bin/env node
+var lt_client = require('../client');
+var open_url = require('openurl');
+
+var argv = require('yargs')
+    .usage('Usage: $0 --port [num] <options>')
+    .option('h', {
+        alias: 'host',
+        describe: 'Upstream server providing forwarding',
+        default: 'http://localtunnel.me'
+    })
+    .option('s', {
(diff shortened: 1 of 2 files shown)


irazasyed:
Perhaps we could also add support for ngrok paid users?

I have an account that supports custom sub-domain and I was in fact, looking for a solution for it.

taylorotwell:
Better to add support for Ngrok paid users.

vesper8:
excellent PR! Too bad it got rejected but I'm definitely going to use it so thanks for posting it!

shiroamada:
+1 for this PR.
It is good to have option choose between ngrok, ngrok paid and localtunnel.me

AshboDev:
@ajmalmsali Nice work, going to start using this! Perhaps you can fork it and people can install valet from your source instead?

Love that you can choose between tunnels, and that it's so easy to set up...

srikat:
Does this solve the redirect loop issue with `valet share`ing secure non-subdomain sites like example.test?

PR #82, opened 2016-05-31
Add support for CS-Cart e-commerce software

drahosistvan:
Added support for [CS-Cart](http://www.cs-cart.com/), which is a very popular e-commerce software.


taylorotwell:
Holding off on this. Think it should be a custom driver.

PR #83, opened 2016-06-01
Changed communication to unix socket

fzoske:
Changed the communication between caddy and php-fpm to unix sockets for better performance and lesser overhead.


taylorotwell:
8ms on default laravel response is already really fast.

PR #87, opened 2016-06-09
Added valet support for Symfony < 3.0

xyNNN:
This pull request adds support for Symfony versions which are lower than 3.0, by the way.... awesome work for **valet** - it work's like a charm :)


taylorotwell:
That's slightly generic of a check. Does Symfony 2.x have a more specific file we could look at?

xyNNN:
I think the ```app/AppKernel.php``` is a very specific file for a Symfony application. So we could improve the condition, what do you think @taylorotwell ?

taylorotwell:
Yeah that sounds possibly more specific.

xyNNN:
I've updated the condition. I think now this driver should support each Symfony 2.x/3.x application ;)

PR #92, opened 2016-06-15
Adds the option to view Caddy access- and error-log

kapersoft:
This PR adds the option to view the Caddy access- and error-log using 
```valet server-log```

The both logfiles are stored in the ~/.valet/Log directory.


rugaard:
I'd love this! +1 👍

PR #96, opened 2016-06-27
WordPressValetDriver: Redirect wp-admin to wp-admin/

tlaverdure:
When attempting to access `/wp-admin` while unauthenticated, WordPress redirects to login `/wp-login.php`. After completing the login form you are then redirected to `wp-admin` instead of `wp-admin/` which causes issues with all the links in the admin until you add the trailing slash.


johnjamesjacoby:
Chiming in to confirm the issue. Solution seems fine enough, and works as expected.

PR #98, opened 2016-07-08
Support xip.io feature

Gounlaf:
W.I.P Support xip.io feature

* Support IPv4


Gounlaf:
I've made mistake: I will create new pull-request on a branch instead of master ><

PR #101, opened 2016-08-01
Create Magento1ValetDriver.php

LasseRafn:
A driver for Magento 1.X (tested with latest 1.9)

Made in a hurry (5 mins max) as I'm leaving the office. Please provide any valid feedback, this is absolutely first time I attempt to create a Valet driver, and I did so by looking at a Magento 2 driver + the documentation.

### DIFF
--- a/cli/drivers/ValetDriver.php
+++ b/cli/drivers/ValetDriver.php
@@ -45,7 +45,7 @@ abstract class ValetDriver
         $drivers = static::driversIn(VALET_HOME_PATH.'/Drivers');
 
         $drivers[] = 'LaravelValetDriver';
-
+        $drivers[] = 'Magento1ValetDriver';
         $drivers[] = 'WordPressValetDriver';
         $drivers[] = 'BedrockValetDriver';
         $drivers[] = 'SymfonyValetDriver';
--- a/drivers/Magento1ValetDriver.php
+++ b/drivers/Magento1ValetDriver.php
@@ -1,0 +1,47 @@
+<?php
+
+class Magento1ValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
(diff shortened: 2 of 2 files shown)


LasseRafn:
I have tested everything I know of (cronjobs, some rest api calls, clean install/installer, uploading files and such) Nothing failed, and no reports created by Magento either. I tested with a fresh copy from Magento, and a copy that I was using with Vagrant (Homestead) before. Both worked smoothly.

LasseRafn:
However, little note: It will break with Php7. Some changes has to be made. I simply decided to switch php version for testing.

LasseRafn:
Remember to switch php version to 5 (5.6) as Magento 1.X does not support PHP 7.
````
brew unlink php70
brew install php56
valet restart
````

nur79:
"valet share" doesn't working and 302 redirecting!

LasseRafn:
@nur79 good catch - I'll look into it right away. My thoughts are that either it's Magento that's set to the valet URL (and hence Magento does not accept the ngrok url) or I missed something in the Valet settings. Will do some testing

LasseRafn:
@nur79 A solution to this is that you go here: ````Admin -> System -> Configuration -> Web```` and set ````Auto-redirect to Base URL```` to "No"

This worked for me 👍 Maybe a better solution is possible, so I'll keep digging

nur79:
@LasseRafn thanks for your suggestions. I already tried that which only allow to load the home page, but the links are not formatted with the shared domain. One possible solution is to override the baseUrl processing of Magento Core which yet to be tested. Once I've a successful test, will inform you.

LasseRafn:
Ah well, I see.. I'll do some testing too. Thanks a lot! Hoping to find a solution

taylorotwell:
Going to let this be a custom driver.

PR #128, opened 2016-08-27
Change documentation link in README

tjb328:
The documentation link should point to master for the most up to date documentation.


tjb328:
it looks like this is a duplicate of #127

PR #151, opened 2016-09-18
Add TYPO3 driver

fschwaiger:
This driver is able to serve TYPO3 projects from version 7 and up. If you grab the newest version and install it to the "web/" subdirectory, the driver should work out of the box. Adjust $documentRoot for a different setup.


taylorotwell:
It seems there are conflicts. Feel free to resubmit when those are resolved.

PR #160, opened 2016-09-20
Add composer cache to travis config

KorvinSzanto:
This makes it so that travis doesn't have to redownload the composer vendor files each time it runs. This helps prevent download inflation on packagist (Might be a negative 😄) and makes travis tests run a lot faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ sudo: false
 before_install:
   - travis_retry composer self-update
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 install:
   - if [[ $setup = 'basic' ]]; then travis_retry composer install --no-interaction --prefer-dist; fi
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey @KorvinSzanto, cool! Is it all still smart enough to redownload the vendor files if the dependencies change? If so, happy to merge.

KorvinSzanto:
Yep, it's just the same as how composer caches on your local machine just
persisted across testing environments.
There might be danger of it ballooning if we switch dependencies a lot so I
think they ask that you clear it if that's the case.

adamwathan:
Cool, thanks!

KorvinSzanto:
Just to revisit this with observations now that it's in place:

New: https://travis-ci.org/laravel/valet/jobs/164053329
Old: https://travis-ci.org/laravel/valet/jobs/164053329

If you expand the composer install task they are all installing from cache now so definitely not inflating packagist downloads anymore. It looks like about 15 seconds saved on downloading those assets so not a huge improvement there.

PR #186, opened 2016-10-05
LaravelValetDriver update, so that supports both public and public_html

jpcursor:
Handy in pre-configured servers e.g. cPanel

Add in bootstrap/app.php before the App return:

$app->bind('public.path', function (){
    return base_path('/public_html');
});

And then you're ready to change your public folder to public_html

### DIFF
--- a/cli/drivers/LaravelValetDriver.php
+++ b/cli/drivers/LaravelValetDriver.php
@@ -12,8 +12,8 @@ class LaravelValetDriver extends ValetDriver
      */
     public function serves($sitePath, $siteName, $uri)
     {
-        return file_exists($sitePath.'/public/index.php') &&
-               file_exists($sitePath.'/artisan');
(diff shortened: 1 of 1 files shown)


KorvinSzanto:
How is this pull request different than #187? Looking at the changes side by side they are the same. Can you close one of em?

jpcursor:
Hello Korvin,

Sorry, i'm a one man army, so i'm not very experienced with open source...
(branches etc)

So I made PR in the master and sierra braches..

Should have done differentely??

Soon i'm schedueling more time to open source, packages etc, will perform
better :

Sure, tell me which should i close?

KorvinSzanto:
Oh I see, disregard what I said :)

taylorotwell:
Would prefer this just be a separate driver. Laravel typically is designed to work with a public directory.

PR #187, opened 2016-10-05
LaravelValetDriver update, so that supports both public and public_html

jpcursor:
Handy in pre-configured servers e.g. cPanel

Add in bootstrap/app.php before the App return:

$app->bind('public.path', function (){
return base_path('/public_html');
});

And then you're ready to change your public folder to public_html


taylorotwell:
Would prefer to have this be a separate driver.

PR #206, opened 2016-11-01
Issue #198: Check if command being run requires sudo access

willvincent:
This will do a 'fuzzy match' on the command being run to determine whether it needs to be proxied through sudo.  Without this, _all_ commands get proxied through sudo on macOS Sierra.  This change seems to resolve that, and should not have adverse effects on earlier versions of OSX.

Since it's a 'fuzzy match' `start` will be a match for `start` and `restart`, etc.

### DIFF
--- a/valet
+++ b/valet
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 SOURCE="${BASH_SOURCE[0]}"
+SUDOCMDS="install start stop secure share"
 
 # If the current source is a symbolic link, we need to resolve it to an
 # actual directory name. We'll use PHP to do this easier than we can
@@ -23,10 +24,13 @@ fi
 # If the command is one of the commands that requires "sudo" privileges
 # then we'll proxy the incoming CLI request using sudo, which allows
 # us to never require the end users to manually specify each sudo.
-if [[ "$EUID" -ne 0 ]]
+if [[ $SUDOCMDS =~ $1 ]]
 then
-    sudo $SOURCE "$@"
-    exit
+    if [[ "$EUID" -ne 0 ]]
+    then
+        sudo $SOURCE "$@"
+        exit
+    fi
 fi
 
 # If the command is the "share" command we will need to resolve out any


taylorotwell:
The reason you are being prompted for password is probably because your sudoers file may be incorrect. Sierra itself does not need this work around.

willvincent:
I find that hard to believe having never edited my sudoers file, and everything worked as expected prior to upgrading to Sierra

willvincent:
Also, I think you're misunderstanding the problem here.. It is _not_ that it's prompting for a password when it shouldn't be, but rather that it's running all commands as sudo, when it need not do so. 

So, valet open is running _as root_ which causes the site to open in the wrong browser. If you always use safari, you'll probably never notice. 

I did a full uninstall of previous valet version, including removing the valet file from /etc/sudoers.d followed by a fresh install of the newest version 2.0.1, and continued to experience the same behavior until I made the changes this PR makes. Now, once again it's functioning as I would expect it should -- and as it did on the earlier version of the OS.

I still am very much convinced this is a sierra issue, if you refuse to merge it, I guess I'll just have to manually apply the fix with any successive update.. which is a bit annoying but not the end of the world.

joshmanders:
So what exactly should our sudoers file look like, because closing with "Your file is probably wrong" isn't helpful... I am getting this same issue, and the only way around it is to apply this patch...

vmitchell85:
Manually applying this fixed my issue on Sierra also.

@taylorotwell If there is a change to the sudoers file that can fix this without the PR would you mind sharing? I haven't touched my sudoers file at all.

EHLOVader:
I am curious why this was closed. 

Did it get fixed in another way? I am still experiencing issues with valet open using sudo, opening in safari instead of chrome, and can't get past that as a serious problem.

mattstauffer:
The reason this was closed is because there are aspects of multiple other commands that still *expect* things to be `sudo` but don't expose that transparently. This PR would need a more thorough vetting of the entire codebase and possibly some changes made (or possibly giving up) in order for it to work fully.

One example Adam gave me was this line, which is used to define which user is running a given command and which requires `sudo`:

https://github.com/laravel/valet/blob/master/cli/Valet/CommandLine.php#L61-L64

As such, I'm working on an alternative PR that will not fix this, but provide a workaround.

PR #222, opened 2016-11-23
Run brew list as user

### LINKED ISSUE #221: Installation fails on newer versions of brew fail because it now complete blocks root from running any commands.
Yes _any_. Including `brew list | grep <formula>` which returns a lovely message of 

```
array(4) {
  [0]=>
  string(79) "Error: Running Homebrew as root is extremely dangerous and no longer supported."
  [1]=>
  string(76) "As Homebrew does not drop privileges on installation you would be giving all"
  [2]=>
  string(41) "build scripts full access to your system."
  [3]=>
  string(0) ""
}
```

I'll prod around locally and submit a PR if I can work out what I'm supposed to do.

Lavoaster:
As brew no longer allows running itself as root, we need to run it as the current user. This resolves #221.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -30,7 +30,7 @@ class Brew
      */
     function installed($formula)
     {
-        return in_array($formula, explode(PHP_EOL, $this->cli->run('sudo -u '.user().' brew list | grep '.$formula)));
+        return in_array($formula, explode(PHP_EOL, $this->cli->runAsUser('brew list | grep '.$formula)));
     }
 
     /**
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -30,7 +30,7 @@ class Brew
      */
     function installed($formula)
     {
-        return in_array($formula, explode(PHP_EOL, $this->cli->run('brew list | grep '.$formula)));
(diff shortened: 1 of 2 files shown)


Lavoaster:
The new blockage on running brew as root doesn't appear to block the `brew service` command.

cli/Valet/Brew.php
```diff
@@ -30,7 +30,7 @@ function __construct(CommandLine $cli, Filesystem $files)
      */
     function installed($formula)
     {
-        return in_array($formula, explode(PHP_EOL, $this->cli->run('brew list | grep '.$formula)));
+        return in_array($formula, explode(PHP_EOL, $this->cli->run('sudo -u '.user().' brew list | grep '.$formula)));
```
KorvinSzanto: This could use ` $this->cli->runAsUser('brew list | grep ' . $formula)));`

Lavoaster: ✅  Done!

jerguslejko:
👍

sebastiandedeyne:
This resolved some installation issues for me, so +1!

dbpolito:
I'm running into the same problem, this PR fixes my issue too.

OwenMelbz:
Has this made it back into a release? as running things as sudo is completely messing up our setups as users do not have sudo/root access?

PR #235, opened 2016-12-09
Handle PATH_INFO and SCRIPT_FILENAME correctly

sidneywidmer:
The fastcgi_params stub is missing the PATH_INFO and the SCRIPT_FILENAME is wrong. I fixed the issue according to the official example: https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/


taylorotwell:
Can you let us know what problems this is causing currently so we replicate and see that this fixes it?

PR #246, opened 2016-12-10
add notes when upgrading from v1

dodyagung:
When upgrading from v1, since it's semantic version, `composer global update` will not update anything. You need to run `composer global require laravel/valet` again.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,7 +27,7 @@ valet stop
 valet uninstall
 ```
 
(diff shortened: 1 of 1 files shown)


rogervila:
It's better to `composer remove laravel/valet` first and restart the mac after installing v2

juukie:
My steps were:
`valet uninstall`
`composer global require laravel/valet`
`valet install`

Didn't have to restart my mac.

taylorotwell:
I noted this in Laravel Valet docs on Laravel website.

PR #248, opened 2016-12-11
Valet for Linux

TheNodi:
I felt in love with homestead a lot of time ago but on my laptop the resources aren't enough so I decided to try to port valet to Linux systems.

As of now I've seen this working on Ubuntu 16.04 and on Fedora 23-25 (with minor changes to SELinux settings). It should be capable of working on every linux system using apt-get or dnf as package manager and that has `service` command available. The structure I've used is expandable for virtually any package/service manager which offers the needed packages (nginx, php, dnsmasq).

Unfortunately I don't own a Mac book (or any other apple product) and I can't test my changes to see if I've broken something. I know the brew class might be refactored a bit, but I hope someone with a mac can help me with that.


taylorotwell:
We don't plan **official** support for Linux. You are welcome to maintain it yourself as a package.

PR #273, opened 2016-12-20
Added TYPO3 Driver

steffenbrand:
Mostly a copy of fschwaigers approach, but this one should merge.
I also allowed myself to add .idea/ to the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 vendor/
 composer.lock
 error.log
+.idea/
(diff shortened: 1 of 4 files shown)


steffenbrand:
Merge it while it's hot ;-)

airlst:
It works pretty well, thanks for the work!

steffenbrand:
Better go with fschwaigers solution: #278

Issue #250, opened 2016-12-12
valet 2.0 ,brew check nginx but not nginx-full,etc


nickfan:
i installed nginx-full instead of nginx via brew,but valet 2.0 only check if `nginx` have installed

Fuitad:
+1

I'm having issues because I have nginx installed via nginx-full :(

Fuitad:
Pull request #286 created. I tested it and it worked fine :).

Fuitad:
Here we go. Pull request #287. Very sorry about that.

PR #287, opened 2017-01-04
Fix for issue #250 - brew package nginx-full not properly recognized as a valid nginx installation

Fuitad:
Fix for issue #250 - brew package nginx-full not properly recognized as a valid nginx installation

Now using the correct nginx service name for restart/stop :)

### DIFF
--- a/cli/Valet/Nginx.php
+++ b/cli/Valet/Nginx.php
@@ -132,7 +132,7 @@ class Nginx
      */
     function stop()
     {
-        $this->cli->quietly('sudo brew services stop nginx'. $this->brew->nginxServiceName());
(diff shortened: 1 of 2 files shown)


cli/Valet/Nginx.php
```diff
@@ -130,7 +132,7 @@ function restart()
      */
     function stop()
     {
-        $this->cli->quietly('sudo brew services stop nginx');
+        $this->cli->quietly('sudo brew services stop nginx'. $this->brew->nginxServiceName());
```
drbyte: I'm guessing you meant to remove `nginx` from "stop nginx" here?

Fuitad:
@drbyte Good catch! I fixed that typo. Thanks :)

huglester:
This one helps. Thank you so much

PR #310, opened 2017-02-02
Add support for Craft 3 Beta

scottwakefield:
Currently the Craft 3 Beta has the front controller stored in a `/web` directory; Craft 2 stored it in the `/public` directory.

As Craft 3 Beta has a `craft` cli bootstrap file, we can use this as our check in the `serves()` method. We can also use this to determine whether to load the front controller from the `/web` or `/public` directory.


sjelfull:
@adamwathan Would be great if could merge this lil thing, since I bet there is some people eagerly awaiting Craft 3 support

jalenconner:
@adamwathan can this please get merged? Would love to have native Craft 3 support in Valet.

PR #336, opened 2017-03-06
Added check to ensure brew options are installed

### LINKED ISSUE #332: [Proposal] Valet Secure Check
I just spent several hours trying to figure out why valet broke after I ran `valet secure`. Everything just stopped working. So I reversed it, but to no avail. I finally came across a thread telling me to try `brew install nginx --with-http2`, which finally fixed it.

My proposal is that within the `secure` command, a check should be run to see if `http2` exists on the nginx server. If it does not, then either issue an info message stating that it should get installed with the above mentioned brew command, or just run the command from within valet.

I looked through the code and in `Valet/Nginx.php` line 40, it is ensuring that nginx is installed and `--with-http2` is passed in as an option. However in `Valet/Brew.php` line 59, the if statement is only checking that nginx is installed, but not checking that the `--with-http2` option is also installed.

I'm not proficient at CLI, but doing something like `brew info nginx | grep -- "--with-http2"` works.

brandonferens:
This proposal ensures that any options to a brew package are installed when initially installing valet. It isn't enough to just check that nginx is installed. Attempting to secure a site without HTTP2 breaks Valet until it is installed.
  
Fixes #332

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -26,11 +26,40 @@ class Brew
      * Determine if the given formula is installed.
      *
      * @param  string  $formula
+     * @param  array  $options
      * @return bool
      */
-    function installed($formula)
+    function installed($formula, $options = [])
     {
-        return in_array($formula, explode(PHP_EOL, $this->cli->runAsUser('brew list | grep '.$formula)));
+        if (! in_array($formula, explode(PHP_EOL, $this->cli->runAsUser('brew list | grep '.$formula)))) {
+            return false;
+        }
+
+        if (! $this->optionsInstalled($formula, $options)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Determine if the given formula's options are installed.
+     *
+     * @param  string  $formula
+     * @param  array  $options
+     * @return bool
+     */
+    function optionsInstalled($formula, $options = [])
+    {
+        foreach ($options as $option) {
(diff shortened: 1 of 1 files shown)


mattstauffer:
@brandonferens Are there any cases in which this can provide a false negative? e.g. Where the install options are different in any way from the check options? Or are they all purely boolean and this will never make a false negative?

Thanks!

brandonferens:
Ha, honestly not sure. It has been awhile...

mattstauffer:
@brandonferens I figured. :) I saw this repo has been a bit dusty and asked if I could go clean up the issues and PRs--if you have some time to take a look I'd love it, but if not just let me know. Thanks!

brandonferens:
I wouldn't be able to look at it until after the 17th.

drbyte:
In looking at this, I think for the proposed code to actually have any effect then the `hasInstalledNginx()` method would also need to pass the necessary options. But then options are defined in 2 separate files, so it'd be better to pass them as a variable, which adds additional complexity.

https://github.com/laravel/valet/blob/49509b0636a5224ac82f19af4946382cdc83d65c/cli/Valet/Brew.php#L66-L70

I haven't run into this being an issue, and thus haven't needed any of this. Granted, I'm not in the habit of intentionally manually installing nginx with a bunch of custom options, so maybe haven't run into the "use case" scenario where this sort of change would be desired.

It feels like unnecessary complexity to me. Just my 2cents.

mattstauffer:
Brandon, thanks so much for your PR!

PR #345, opened 2017-04-01
Adds CodeIgniter Driver

refringe:
I know, I know... But some of us still need to use it. 😏 

This adds a working driver for the CodeIgniter framework. Given how dynamic the folder structure can be in CI, I’ve resorted to searching the `$sitePath` directory for files. This obviously isn’t as efficient as some of the other drivers, but I have yet to notice any slowdown.

Let me know what you think.

### DIFF
--- a/cli/drivers/CodeIgniterValetDriver.php
+++ b/cli/drivers/CodeIgniterValetDriver.php
@@ -1,0 +1,86 @@
+<?php
+
+class CodeIgniterValetDriver extends ValetDriver
+{
+    /**
+     * Determine if the driver serves the request.
+     *
+     * @param  string  $sitePath
+     * @param  string  $siteName
+     * @param  string  $uri
+     * @return bool
+     */
+    public function serves($sitePath, $siteName, $uri)
+    {
+        // The CodeIgniter system path is configurable. This makes checking for a system file difficult. Instead, let's
+        // attempt to find the index file (predictable location), open it, and search for an identifiable string.
+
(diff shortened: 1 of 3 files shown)


adamwathan:
Man I would be really surprised if this doesn't have an impact on performance for sites served by the `BasicValetDriver` 😕 Wouldn't this have to recursively search *every* folder in a project before ever hitting the last driver? Including insane stuff like the `node_modules` folder and `vendor` folder?

Def would be nice to get CodeIgniter support, but would love if we could come up with another way to detect them 😬

refringe:
Completely understandable.

I have a few other ideas that would be faster. I'll update the pull when I've tested them out.

refringe:
This should be a lot faster now.

The problem is that the default CI folder structure has the system directory placed within the public folder (the root of the project). Best practice is to separate the private and public files, but there's no standard amongst CI projects. 

I guess the easiest way to identify a CI project is with the `index.php` file. I've updated the driver to check a few common places. If the file is found it's searched for an identifiable string.

This won't serve every CI project under the sun, but it should easily serve the majority.

Let me know. 👍

joshmanders:
@refringe I would just check for best practices on the CodeIgniter structure. Anyone who doesn't follow that can throw their own custom driver into their project to detect how they want.

refringe:
I made a mistake when I was initially testing with one of my CI projects. The `BasicValetDriver` works well with CodeIgniter!

I'll see myself out.

PR #349, opened 2017-04-04
Adding directory listing option

revvi:
I have created directory listing option in this Laravel Valet. This will simulate "autoindex on" option that is usually found in nginx configuration. Instead of showing 404 (file not found) error message, it can display all files & directories in site folder.

How to use:

`valet directory-listing` is to show current configuration
`valet directory-listing on` is to enable directory listing
`valet directory-listing off` is to disable directory listing

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -250,6 +250,26 @@ if (is_dir(VALET_HOME_PATH)) {
             output('NO');
         }
     })->descriptions('Determine if this is the latest version of Valet');
+
+    /**
+     * Determine directory listing option
+     */
+    $app->command('directory-listing [option]', function ($option = null) {
+        $key = 'directory-listing';
+        $config = Configuration::read();
+
+        if (is_null($option)) {
+            if (!isset($config[$key])) {
+                output('Directory listing is not enabled. Set to on or off. Default is off');
+            } else {
+                output('Directory listing is ' . $config[$key]);
+            }
+        } else {
+            $config[$key] = $option;
+            Configuration::write($config);
+            output('Directory listing configuration is saved');
+        }
+    })->descriptions('Determine directory listing behavior. Default is off');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


rdeutz:
@taylorotwell any chance to add this to the next release?

drbyte:
Some useful suggestions have been posted at https://github.com/laravel/valet/issues/610 which may provide similarly without the complexity of merging this.

navinweb:
Thank you very much, @revvi

yosydz:
work thanks

orionrush:
@drbyte helpful suggestion re #610 notwithstanding, I just need the contents directories listed. I'm a bit surprised as I just assumed this would be a trivial feature that a server tool for site development would expose.

I also do a lot of agency work, but in many cases, we don't have the full project spooled up in a test environment due to the complexity of the sites. Instead, for testing markup/js/CSS we work with a page templating engine which can produce dozens of output page views, which are standalone, with no 'site' structure or interlinking to speak of.

In these cases I love how quickly I can spool up valet, as it carries a lot of cognitive load I would have to carry in setting up an alternative (vagrant etc.) == it makes my life easier. 

So though it's not a big deal to [to drop an index.php](https://css-tricks.com/snippets/php/display-styled-directory-contents/)  file in there that will do the trick, its just another step outside the project work which I have to do to make the server do what servers do.

Looking at the proposed changes, it really doesn't seem (to me) to add a terrible amount of complexity to the project. It's a bit more complex than one user dropping in a custom index.php file into one project, but certainly, less than many of us using index.php in lots of projects.

Complexity has to live somewhere, in this case, I'm +1 for increasing valet's utility.

I've been poking about, just to see if there was an alternative hook were custom directives could be bootstrapped to produce the same results, it dosesn't seem to be an ideal use of a [custom-driver](https://laravel.com/docs/5.8/valet#custom-valet-drivers).

drbyte:
@revvi Are you willing to rebase this PR? (or at least resolve the merge conflicts)

PR #353, opened 2017-04-11
Unlink renamed links

antonioribeiro:
Today, if you give a different name to your link:

```bash
➜  valet git:(master) valet link valet-renamed
A [valet-renamed] symbolic link has been created in [/Users/antoniocarlos/.valet/Sites/valet-renamed].
```

And try to unlink it, it says it removed (a wrong site name), but nothing really happens:

```bash
➜  valet git:(master) valet unlink
The [valet] symbolic link has been removed.

➜  valet git:(master) valet links
+-------------------+-----+------------------------------+---------------------------------------------------------------+
| Site              | SSL | URL                          | Path                                                          |
+-------------------+-----+------------------------------+---------------------------------------------------------------+
| bureau            |     | http://bureau.dev            | /Users/antoniocarlos/code/bureau/bureau.io                    |
| dusk-test         |     | http://dusk-test.dev         | /Users/antoniocarlos/code/laravel/dusk-test                   |
| laravel           |     | http://laravel.dev           | /Users/antoniocarlos/code/laravel/laravel                     |
| nucleoconstelar   |     | http://nucleoconstelar.dev   | /Users/antoniocarlos/code/nucleoconstelar/nucleoconstelar.com |
| parlamentojuvenil |     | http://parlamentojuvenil.dev | /Users/antoniocarlos/code/alerj/parlamentojuvenil             |
| pragmarx          |     | http://pragmarx.dev          | /Users/antoniocarlos/code/pragmarx/pragmarx.com               |
| transparencia     |     | http://transparencia.dev     | /Users/antoniocarlos/code/alerj/transparencia                 |
| valet-renamed     |     | http://valet-renamed.dev     | /Users/antoniocarlos/.composer/vendor/laravel/valet           |
+-------------------+-----+------------------------------+---------------------------------------------------------------+
```

This PR will fix this behaviour

```bash
➜  valet git:(improveUnlink) valet unlink
The [valet-renamed] symbolic link has been removed.
```


drbyte:
Probably good to fix.

But I've always specified the domain I wanted to unlink:

`valet unlink valet-renamed`
instead of just 
`valet unlink`

antonioribeiro:
@drbyte another "fix" is to remove the ability to *not* send the name. 

But the fact is that this way it is confusing, because I rarely set a new name to the link and always do "valet unlink", until now.

drbyte:
Ya, I agree it's a bit confusing.

I'm in the habit of always using shorter domains that don't get confused with the real one, so when I start typing URLs I don't have to type out the entire thing to get to which one I want.  

'abc.dev' is much faster to get to than `amalgamated-business-coffeemakers.dev`  :)

adamwathan:
How does this behave if someone has multiple links pointing to the same project? I know a bunch of folks do that for Wordpress multi-site installs or apps with subdomain routing.

antonioribeiro:
I'll test this scenario and get back to you

antonioribeiro:
When you tell it to unlink a renamed link, which has other aliases to the very same directory, it removes only the one you asked:

image

adamwathan:
Perfect! What do we think the behavior should be in that case if you don't provide a link name?

antonioribeiro:
Well, I think it should remove all links related to the current directory, but that may be too harsh, so the current behaviour, more pleasant and acceptable IMO, is to unlink only the very first link related to the current directory.

```
➜  site valet unlink
The [site-renamed-1] symbolic link has been removed.
➜  site valet unlink
The [site-renamed-2] symbolic link has been removed.
```

adamwathan:
The more I think about it I think it should fail with a warning stating there are multiple links, and force you to specify which one to remove?

antonioribeiro:
I like it. Doesn't leave room for wrongly deleted links.

antonioribeiro:
What about this?

image

antonioribeiro:
@adamwathan, to clarify: I already pushed some more code to throw this exception

drbyte:
I've just tested this out and am satisfied that it provides a useful benefit.

However, it relies on the `use DomainException;` declaration at the top of `Sites.php`, which was removed in a recent PR. I had to add it in manually, since it doesn't show in the diff here. Not sure how github will handle that since it's not showing a merge conflict and it's not showing it in the diff.
Nevertheless it's a simple fix which can be reviewed after a merge.

PR #354, opened 2017-04-11
Secure option for link command

mihaliak:
Added `--secure` option for `valet link` command. This can save some time when people want to link specific folder to domain and then generate certificate.


vesper8:
+1

I would also like to see an option to set the --secure as a default parameter

PR #372, opened 2017-05-06
Support another "wildcard dns provider": nip.io

Gounlaf:
* Rename method to more generic name
* Support nip.io provider

### DIFF
--- a/server.php
+++ b/server.php
@@ -20,11 +20,11 @@ function show_valet_404()
 /**
  * @param $domain string Domain to filter
  *
- * @return string Filtered domain (without xip.io feature)
+ * @return string Filtered domain (without wildcard dns feature (xip.io/nip.io))
  */
-function valet_support_xip_io($domain)
+function valet_support_wildcard_dns($domain)
 {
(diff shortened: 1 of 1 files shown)


joshmanders:
Even better, put the wild card dns providers into a config section, then add a method to allow you to add/remove them, so then you can be like "I want my own wildcard dns provider" and then you can be like `valet dns-like-whatttttt dns.my.shizzle` and now it's donezo. Feel me?

Gounlaf:
Hi @joshmanders,

When I made a first PR for xip.io support (https://github.com/laravel/valet/pull/126), @adamwathan express this (https://github.com/laravel/valet/pull/126#discussion_r80365906)

> One other thing I just realized, we try really hard not to define functions that could cause collisions in `server.php` because this file is included on every single request, and if the site being served tries to define a function we've already defined, the user will get an error.

I don't know (and don't found) a lot a "wildcard dns providers", but yeah, I'm in favor of doing it in a configurable way:

* A CLI cmd to add/remove a provider
* Configuration stored in user's valet configuration folder: a simple array

PR #421, opened 2017-08-24
Bump version constraint on titenco/collect package

tomeh:


adriaanzon:
Shouldn't this be in format `5.3.*` rather than `^5.3`?

I switch PHP versions a lot. When I install Valet with PHP 7 linked, it will install `tightenco/collect` version 5.5, which isn't compatible with PHP 5.6. So when I switch back to PHP 5.6, Valet will fail with syntax errors.

PR #436, opened 2017-09-18
Dev => Localhost

LasseRafn:
Related to #434

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -111,7 +111,7 @@ class Configuration
     function writeBaseConfiguration()
     {
         if (! $this->files->exists($this->path())) {
-            $this->write(['domain' => 'localhost', 'paths' => []]);
+            $this->write(['domain' => 'test', 'paths' => []]);
         }
     }
 
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -111,7 +111,7 @@ class Configuration
     function writeBaseConfiguration()
     {
         if (! $this->files->exists($this->path())) {
-            $this->write(['domain' => 'dev', 'paths' => []]);
+            $this->write(['domain' => 'localhost', 'paths' => []]);
         }
     }
 
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -33,7 +33,7 @@ class DnsMasq
      *
      * @return void
      */
-    function install($domain = 'localhost')
+    function install($domain = 'test')
     {
         $this->brew->ensureInstalled('dnsmasq');
 
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -33,7 +33,7 @@ class DnsMasq
      *
      * @return void
      */
-    function install($domain = 'dev')
+    function install($domain = 'localhost')
     {
         $this->brew->ensureInstalled('dnsmasq');
 
--- a/tests/DnsMasqTest.php
+++ b/tests/DnsMasqTest.php
@@ -39,10 +39,10 @@ class DnsMasqTest extends PHPUnit_Framework_TestCase
         $dnsMasq->configPath = __DIR__.'/output/dnsmasq.conf';
         $dnsMasq->resolverPath = __DIR__.'/output/resolver';
 
-        $dnsMasq->install('localhost');
+        $dnsMasq->install('test');
 
-        $this->assertSame('nameserver 127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/resolver/localhost'));
-        $this->assertSame('address=/.localhost/127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/custom-dnsmasq.conf'));
+        $this->assertSame('nameserver 127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/resolver/test'));
+        $this->assertSame('address=/.test/127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/custom-dnsmasq.conf'));
         $this->assertSame('test-contents
 
 conf-file='.__DIR__.'/output/custom-dnsmasq.conf
--- a/tests/DnsMasqTest.php
+++ b/tests/DnsMasqTest.php
@@ -39,10 +39,10 @@ class DnsMasqTest extends PHPUnit_Framework_TestCase
         $dnsMasq->configPath = __DIR__.'/output/dnsmasq.conf';
         $dnsMasq->resolverPath = __DIR__.'/output/resolver';
 
-        $dnsMasq->install('dev');
+        $dnsMasq->install('localhost');
 
         $this->assertSame('nameserver 127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/resolver/dev'));
-        $this->assertSame('address=/.dev/127.0.0.1'.PHP_EOL, file_get_contents(__DIR__.'/output/custom-dnsmasq.conf'));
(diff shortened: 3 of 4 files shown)


GaPhil:
Nice one!

drbyte:
Looking good.

A couple points to bring the docs in line with the code:
- There's a reference to *.`dev` in the `readme` (line 14) 
- and in comments for `createDomainResolver` in `cli/Valet/DnsMasq.php`

drbyte:
My bad. Given that the documentation is maintained elsewhere, I guess the mention of other TLDs should not go into this readme, but rather to the laravel docs. I've submitted a PR for that separately: https://github.com/laravel/docs/pull/3652

leewillis77:
Just for reference `.localhost` [is defined](https://tools.ietf.org/html/rfc6761#section-6.3) to have some non-standard behaviour. Of relevance, it's defined to *always* resolve to the local loopback address, so using it precludes you from mapping anything under the TLD to an address other than 127.0.0.1.  The TLD also specifies (same RFC) that any DNS request other than an address query should fail which again may cause issues somewhere (think querying for TXT, or MX records etc.) 

`.test` would seem more flexible as [it's defined](https://tools.ietf.org/html/rfc6761#section-6.2) to act "normally".

barryvdh:
@leewillis77 
> Of relevance, it's defined to always resolve to the local loopback address, so using it precludes you from mapping anything under the TLD to an address other than 127.0.0.1. 

But in the case of Valet, wouldn't you always want to map it to localhost?

leewillis77:
I'm not invested in the final decision on which to use (I don't use Valet personally and ended up here through some other route). However I do think it's important that whoever makes the decision does so with the understanding that .localhost isn't *just* a reserved TLD, it also has other limitations on functionality which may (or may not) be relevant.

Those limitations aren't shared by .test.

adamwathan:
> Just for reference .localhost is defined to have some non-standard behaviour. Of relevance, it's defined to always resolve to the local loopback address, so using it precludes you from mapping anything under the TLD to an address other than 127.0.0.1.

I think given this it's probably smart to use `.test` even though I hate it :)

@LasseRafn Do you mind updating the PR to `.test`? I know I was the one who asked for `.localhost` originally but I was wrong, heh.

I'm probably going to use `.app` personally until Google ruins that too, but I'll take too much heat if I make that the default for everyone :)

LasseRafn:
Hey @adamwathan, I will update it :) no problem

LasseRafn:
Updated 👍 I kind of hate the `.test` TLD.. but I guess we will have to get by so far. Hopefully Safari won't enforce the rule, so that I can continue using `.dev`.

aaemnnosttv:
What about using something super minimal like `.d` ?

barryvdh:
There are a few reserved TLDs like `.test`, `.example` and `.localhost` you can use. The rest is subject to registration.

aaemnnosttv:
Yeah, I understand that. From what I've found though, the chances of a single letter TLD seem pretty slim. None currently exist either.

adamwathan:
Thanks for this @LasseRafn!

fridzema:
Maybe just `.valet` ?

OwenMelbz:
annoying thing about `.test` is that it resembles nothing to what it is - need something that can express, `.buildingit` - its almost like... `.dev` is needed...

crabmusket:
Just to elaborate on the undesirable behaviour of `.localhost` - in my team some of us use Homestead on Linux, and some use Valet on OSX. `.localhost` is incompatible with Homestead because the site isn't served from localhost, but from the virtual box's local IP. It'd be nice to have a uniform experience from all developers (which we can canonize in the readme, etc.) and not have to change e.g. webpack config for the hot reload server depending on OS.

zaknes:
`.app`

CydeWeys:
Thanks for handling this.  I concur that `.test` is the best choice.

OwenMelbz:
@cydeweys that is more descriptive to a test though right? Doesn’t remotely suggest that the environment your on is actually local or development. It literally says it’s a test. When it’s not.

Personally I say stop complaining about installing SSL and keep .dev

CydeWeys:
@OwenMelbz .dev is a real top-level domain.  You should not be using fake domain names on real TLDs.  That will cause huge problems down the line when your fake domain names start experiencing name collisions with real domain names that people actually own.  The big advantage `.test` has is that is guaranteed to *never* be a real TLD, so there will never be name collisions with real domain names.

fridzema:
@OwenMelbz https://laravel-news.com/chrome-63-now-forces-dev-domains-https it seems there is no workaround to use .dev in chrome, thats a pretty big problem then...

tomck:
@zaknes Google has already announced they will own .app soon. Please consider sticking with the TLDs which have been listed as never-to-become-real. https://www.registry.google/about/domains.html

OwenMelbz:
@fridzema they work absolutely fine, all you need to do is have a locally signed SSL - all chrome does is redirect non-ssl to ssl

So if you use something like valet to do “valet secure” then you have no issue anymore.

@CydeWeys it’s no problem at all, you know google has owned .dev for literally years right? So unless you’ve hit an issue in the past few years - chances are it will still be absolutely fine

fridzema:
Maybe we can start an petition or need to start fundraising to claim the .dev domain back from Google for all of us. Who is in?

OwenMelbz:
Just go for .develop then :p descriptive and basically the same, and shorter than .localhost!

CydeWeys:
@OwenMelbz Trust me, I am very, very well aware that Google has owned `.dev` since 2014 ;)

`.dev` is not yet launched, so there are currently only a handful of operational domain names on it.  Thus you're unlikely to have experienced a name collision on it yet.  Given that it hasn't launched yet, though, saying "We haven't had problems yet" isn't a useful predictor for lack of future problems.

And don't go for `.develop` either.  It doesn't happen to exist yet but it may well be acquired in a future round of gTLD expansion, so you're just kicking the problem down the road.  Best to switch to something now that is guaranteed to never exist.

drbyte:
Seriously, why are we still discussing this here?

lukasjuhas:
Annoyingly if you are using google api, and you need to enter a callback url, you are not allowed to use `.test`. In that case `.app` as a temp. alternative works for me too @adamwathan 

screen shot 2017-12-19 at 22 20 17

CydeWeys:
Don't use `.app` either; like `.dev`, it's also a real TLD owned by Google.

At the point where you're testing integration with other services, the best practice is to use a real domain or subdomain that does not have potential name collisions on the Internet.  If, e.g., you happen to own lukasjuhas.com, then spin up a development subdomain at dev1.lukasjuhas.com or similar.  You can get a valid free SSL certificate for that subdomain using Let's Encrypt, and you can also configure DNS to be 127.0.0.1 for that subdomain if you wouldn't like for it to be reachable from the Internet.

lukasjuhas:
Thanks @CydeWeys, that's probably what I'll end up doing! It's really annoying though..

sheldonkotyk:
I ended up buying dev.rip if anyone feels like using it for their dnsmasq. Trust me, nothing of importance will end up there :)

CydeWeys:
Add dev.rip to the HSTS preload list ;)

sheldonkotyk:
Ha. 

Honestly, I valet secure for everything I build locally anyway. It's just when google starts using .dev for things that I will actually need that I decided to make the switch. I just can't believe Google doesn't allow .test for their oauth.

lukasjuhas:
I've submitted a request to allow `.test` domains..

https://issuetracker.google.com/issues/70856355

ryahpalma:
> Maybe just `.valet` ?

I'm using .valet and it works like a charm in Safari, If I try to insert '.test' in the search bar he always does a search on Google.

PR #438, opened 2017-09-20
Driver for frontend projects with a build step

morrislaptop:
Often the index.html and associated assets will be placed into a "build", "dist", or "public" directory. Common with tools like webpack, gulp, browserify, gulp etc..


mattstauffer:

Thanks!

PR #452, opened 2017-10-04
Port improvements for Magento2 driver from Valet Plus project

lazyguru:
This ports improvements for the Magento2 driver from the Valet Plus project (https://github.com/weprovide/valet-plus) back to this project (which is sort of an upstream)

This fixes an issue with Magento 2.2 that was reported to me outside of this repo by @piotrekkaminski (Product Manager at Magento)

### DIFF
--- a/cli/templates/magento2.php
+++ b/cli/templates/magento2.php
@@ -1,0 +1,47 @@
+<html>
+<head>
+    <title>Valet - Not Found</title>
+
+    <style>
+        body {
+            font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            text-align: center;
+            font-size: 25px;
+            font-weight: 200;
+            line-height: 1.7;
+        }
+
+        h1 {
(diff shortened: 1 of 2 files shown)


cli/templates/magento2.php
```diff
@@ -0,0 +1,47 @@
+<html>
```
timneutkens: This can be removed

lazyguru: Actually it’s not Valet+ specific. It’s used on line 113 of the driver

timneutkens: Ah alright 👍 🙌

drbyte: ... but is it really needed? A simple `echo` could accomplish the same purpose.

lazyguru: My intent was to keep it as close as possible to the downstream version so any future improvements from either side would be easy to merge. Plus I like the display of the page with a clickable link to start the web installer.

mattstauffer:
@lazyguru Can you describe in simple terms what these improvements are? We definitely want to merge in fixes, but we don't necessarily have the same perspective on how many features to add, so we want to be able to parse out the differences and merge or not merge them each separately.

Thanks!

lazyguru:
Actually @mattstauffer, I think I'm going to withdraw this PR and suggest people use valet-plus if they need this.

mattstauffer:
@lazyguru 👍

PR #458, opened 2017-10-14
Silver Stripe valet driver

adrhumphreys:
This is for the SilverStripe CMS, the biggest change in this that is off putting would be the buffer size modification to the nginx config in valet.conf which increases the buffer size above the default values, from my understanding this shouldn't effect any existing applications but is required for the silver stripe admin to work. 

An alternative if we don't want to modify the valet.conf would be to add a "Drivers" resource doc and link to various drivers for valet so that people can easily find and install them


mattstauffer:

Thanks!

PR #467, opened 2017-11-02
[2.0] Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/tests/BrewTest.php
+++ b/tests/BrewTest.php
@@ -3,9 +3,10 @@
 use Valet\Brew;
 use Valet\Filesystem;
 use Valet\CommandLine;
+use PHPUnit\Framework\TestCase;
 use Illuminate\Container\Container;
 
-class BrewTest extends PHPUnit_Framework_TestCase
+class BrewTest extends TestCase
 {
     public function setUp()
     {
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -24,7 +24,7 @@
         }
     },
     "require": {
-        "php": ">=5.6",
+        "php": ">=7.0",
```
drbyte: This probably isn't a good idea since Valet itself still supports PHP 5.6 as a courtesy to the many developers required to support older environments. 

That said, you could update `.travis.yml` to add `7.2` since `nightly` actually builds PHP 7.3 currently.

gabriel-caruso: I only dropped support was because the tests were falling. What do you suggest me to do?

drbyte: In `composer.json` use:

```json
        "phpunit/phpunit" : "^5.7|~6.2"
```

gabriel-caruso: @drbyte I didn't know I can require one `or (I)` other version. I'll work on that!

eberkund:
Looks good to me

Issue #473, opened 2017-11-24
[Feature] Set server environment variables per site


MartinPeverelli:
Some sites or frameworks depend on server environment variables to be set on the server configs (at the Apache vhost or nginx.conf levels), and do not provide a way for those variables to be declared on the code.

A quick example of this, would be specific to my line of work: Magento.
A Magento application that hosts two `websites` needs for the server to have two variables set: `MAGE_RUN_CODE` and `MAGE_RUN_TYPE`. These variables are read on the `index.php` file and its the way the application knows what `website` to serve.
I imagine this is not an uncommon situation...

Currently, there is no way for me to set these environment variables with Valet without some ugly hack, or editing my `index.php` file to hardcode those values in.

My proposal is to create a `.env.valet` file at the root of the application (`$sitePath`), and have Valet load those for me.

Here is a sample file:
```
<?php

return [
    'website1' => [
        'MAGE_RUN_CODE' => 'usa',
        'MAGE_RUN_TYPE' => 'website',
    ],
    'website2' => [
        'MAGE_RUN_CODE' => 'arg',
        'MAGE_RUN_TYPE' => 'website',
    ],
];
```
The idea is quite simple: the file will be read by the Valet Driver, and if the top level `key` matches the requested site, then the key-value pairs inside will be loaded as server environment variables.

In the example above, when `website1.dev` is requested, then this will be loaded:
```
$_SERVER['MAGE_RUN_CODE`] === 'usa' // true
$_SERVER['MAGE_RUN_TYPE`] === 'website' // true
```

I've submitted a proposal for this feature, but I'm sure it could use further discussion. If its liked, then it would require a little documentation to explain the functionality and the `.env.valet` file structure.

PR #474, opened 2017-11-24
#473 Add feature to load server environment variables from special file

MartinPeverelli:
Proposal for https://github.com/laravel/valet/issues/473


Lloople:
Hi @adamwathan @mattstauffer, is there any reason for not merge this pull request? I found that I need to set `env` variables per site and can't find a proper way. Adding this `.env.valet` file into the project looks pretty neat.

Thank you

mattstauffer:
@Lloople I don't see anything particularly wrong with the PR. What I did was close obvious "no" PRs, shut down all driver requests for now, and leave PR's like these open for Adam and Taylor to merge when they have some time to look over them and decide whether they fit within the spirit/intent of Valet. I think this being open is not a "no" but just a "taylor and adam haven't had time to dig into it yet".

iboldurev:
Please add this implementing

osrecio:
Hi, I think is a good approach for some kind of projects to have env variables and load it inside a project.

bramus:
Nice work! Would love to see this one get merged.

cli/drivers/BasicValetDriver.php
```diff
@@ -44,10 +44,12 @@ public function isStaticFile($sitePath, $siteName, $uri)
      */
     public function frontControllerPath($sitePath, $siteName, $uri)
     {
+        $this->loadServerEnvironmentVariables($sitePath, $siteName);
```
mattstauffer: Since this seems to be happening in every `frontControllerPath` method, is it possible to instead just run this in `server.php` right before it calls `frontControllerPath`? (I don't know the code super well so there may be a good reason why not...)

drbyte: Technically it probably could.

https://github.com/laravel/valet/blob/0323607850ec051ad4746409ed0992c5f24c37d9/server.php#L134-L147

MartinPeverelli:
@mattstauffer @drbyte I took both your advices and refactored this idea.
The file is now `.valet-env.php`, which makes it obvious that it should contain a PHP array (instead of the key=value pairs of regular dotfiles).
And I changed the method to be `public` and called it from `server.php` as requested, which allowed me to revert all the changes to all the Drivers.

Let me know what you think!

If this gets merged into master, I would also love to help with the documentation of this feature, how can I do that?

cli/drivers/ValetDriver.php
```diff
@@ -177,4 +177,28 @@ protected function isActualFile($path)
...
+            return;
+        }
+
+        $variables = include $varFilePath;
+        if (! isset($variables[$siteName])) {
+            return;
+        }
+
+        foreach ($variables[$siteName] as $key => $value) {
+            $_SERVER[$key] = $value;
```
Lloople: Why not using `putenv($key.'='.$value)` here along with the already existing `$_SERVER`? 🤔

MartinPeverelli: I was only trying to mimic Apache's way of loading env vars on the vhosts files, and if I'm not mistaken, that only sets values on`$_SERVER`.
I don't know if duplicating the values on both `$_SERVER` and `$_ENV` is useful or desirable, but as you say, we can add that single line to the loop if we do.

What say the rest of us?

drbyte:
@MartinPeverelli have you considered a way to allow for wildcard domain-names with this? 

I have a use-case where simply dropping in a non-domain-specific `.valet-env.php` file would make environment overrides easier. But I don't want to have to edit the file to set domain name/s inside it.
It's admittedly a little different use-case than your original need which was to do domain-specific custom overrides.

drbyte:
I've used this proposed PR on a project. It worked well.

I do think altering it to allow for being domain-agnostic would be nice too.

drbyte:
Waiting for @MartinPeverelli to merge [my PR to his valet fork](https://github.com/MartinPeverelli/valet/pull/1/files), which will update this PR, and then this could be merged.

MartinPeverelli:
@drbyte merged your PR into my fork! Sorry for the delay, hectic RL.

drbyte:
Thanks @MartinPeverelli 

I've been using this code for the past 3 weeks, and haven't run into any problems. It's definitely been helpful to keep project-specific environment variable configs in a separate `.valet-env.php` file in my app, instead of having to touch the valet-generated nginx configs.

@mattstauffer IMO this is safe to merge.

mattstauffer:
I think I got mixed up and thought this was adding `.env.valet` to each application, not `env-valet.php` to the root. Is there a reason we went this way? I was about to release but this feels weird.

MartinPeverelli:
The change to the file name was because the contents are a PHP array, instead of the key=value pair on the other .env files.

mattstauffer:
@MartinPeverelli sure... and you had to do PHP so you can differentiate them per domain. I remember now. Thanks!

drbyte:
Note: The filename is `.valet-env.php`

PR #475, opened 2017-11-29
Add support to Symfony Flex directory structure

IndraGunawan:
reference: https://symfony.com/doc/current/setup/flex.html#upgrading-existing-applications-to-flex

### DIFF
--- a/cli/drivers/SymfonyValetDriver.php
+++ b/cli/drivers/SymfonyValetDriver.php
@@ -13,7 +13,7 @@ class SymfonyValetDriver extends ValetDriver
     public function serves($sitePath, $siteName, $uri)
     {
         return (file_exists($sitePath.'/web/app_dev.php') || file_exists($sitePath.'/web/app.php')) &&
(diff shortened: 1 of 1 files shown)


jerguslejko:
Wouldn't this remove support for previous versions of Symfony?

IndraGunawan:
@jerguslejko current driver using old directory structure of Flex, so i just change to new directory structure of Flex.
ref: http://fabien.potencier.org/symfony4-directory-structure-updates.html

simensen:
On checking this again, I think I see what @IndraGunawan did here. The driver had previously supported the old Flex structure. I hadn't realized that. :)

So the only thing that was changed is that the `index.php` for Flex applications (which weren't there at all for older Symfony applications) is in `public` instead of `web`.

simensen:
👍 LGTM

simensen:
@adamwathan I believe this one should be ready to go as well.

taylorotwell:
Thanks

PR #492, opened 2017-12-19
Remove duplicated forceTrailingSlash

vinkla:
This pull request makes the Bedrock driver extend the WordPress driver. This way we don't have to redefine the `forceTrailingSlash` method.

### DIFF
--- a/cli/drivers/BedrockValetDriver.php
+++ b/cli/drivers/BedrockValetDriver.php
@@ -1,6 +1,6 @@
 <?php
 
-class BedrockValetDriver extends BasicValetDriver
+class BedrockValetDriver extends WordPressValetDriver
 {
     /**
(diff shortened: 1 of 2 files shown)


mattstauffer:
Hey @vinkla, we're going to be making some changes to how we handle drivers soon, and while this change looks good, it actually would break the upcoming changes--so I'm going to close it for now. Thanks for your PR!

vinkla:
@mattstauffer no worries! What are the changes you're working on?

mattstauffer:
@vinkla Providing a more extensible and modular structure (hopefully!) around drivers :)

vinkla:
@mattstauffer great! Would be awesome if drivers could be installable with Composer. The local driver feature is great but copying/pasting isn't.

PR #493, opened 2017-12-20
Valet secure, unsecure PhpFpm restart remove

pavinthan:
Valet secure, unsecure PhpFpm restart remove, 
It works without restart PhpFpm

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -129,8 +129,6 @@ if (is_dir(VALET_HOME_PATH)) {
 
         Site::secure($url);
 
-        PhpFpm::restart();
-
         Nginx::restart();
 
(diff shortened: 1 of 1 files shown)


drbyte:
Agreed. 
No need to restart php-fpm when only updating nginx site configs for secure/unsecure.

My only hesitation initially was that if the PHP version was also being changed and thus the `.sock` file was being re-specified then it might have a problem. BUT running `secure` or `unsecure` is not related to that. 
Worst case one can always safely run `valet restart` or `stop` and then `start` if they really needed to kick php-fpm.

Tested it in my environment with no ill side-effects.

Thus I think this is safe to merge.

PR #494, opened 2017-12-20
Fix valet doesn't recognize PHP version correctly

smartlara:
Valet loops over supported PHP versions, but passes index of the array instead of version which makes it think PHP is installed although it is a fresh installation.


smartlara:
Sorry, I had to update dependencies.

PR #507, opened 2018-01-05
Added Craft multi-locale driver supporting sites with subfolder locale

ryanfoote:
This driver adds support to Valet for Craft sites with different locales electing to use subfolder pathing (`domain.ca/fr`, for example) for one or all of their various locales. It supports Craft versions 2 & 3.


mattstauffer:
@ryanfoote Thanks for the PR! I know I was the one who brought it up to up, but I actually just went around to all the PR's for adding new drivers yesterday and closed them with this message, so I will do that here too--but trust me that this isn't the end. :) Thanks for your work packaging the other one I linked into a PR!

----

Thanks!

PR #514, opened 2018-01-29
Set files path as a static path for concrete5 driver

KorvinSzanto:
Previously, testing a concrete5 site that had tons of 404s and required file based sessions would force each request to load sequentially causing the site to load pretty slowly. This change makes it so that the driver is aware of the files directory and markes anything within as a static file.


mattstauffer:
@KorvinSzanto Could you ask a few other concrete5 folks to check this out for you, and give us emoji responses so we can verify it's working for multiple folks? None of us use concrete5 so we can't check this.

JohntheFish:
What is the specification on $uri ? Is it site relative, or host relative? If host relative, testing ===0 could have issues with installations in sub-folders.

biplobice:
This helps me a lot for sites which have lots of 404s.
It works fine for me, and the installation in sub-folders didn't work, without this code as well.

mattstauffer:
If we can get more concrete5 folks to check it, we could merge this, but it's been four years since I asked for more input so I'm gonna close it.

mattstauffer:
@KorvinSzanto To be honest, I was looking for comments and didn't even check the emoji responses :)  Please feel free to re-make the PR if you'd like.

KorvinSzanto:
I've restored the branch so this PR should be reopenable. If not I can go ahead and make a new PR

mattstauffer:
@KorvinSzanto Thanks! i just did a merge resolution; can you verify it still works correctly after that?

PR #516, opened 2018-02-01
Craft driver web or public

khalwat:
The previous method simply returned `web` if there was a file named `craft` in the $sitePath (which means we’re running Craft 3).

The problem is that many sites, especially upgraded ones, will be serving out of `public`. This is also in line with the default Forge/ServerPilot default dirs.

So we test to see if either `/web` or `/public` exist, and return them if so. Otherwise default to the current method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 vendor/
 composer.lock
 error.log
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 vendor/
 composer.lock
 error.log
+.idea
--- a/cli/drivers/CraftValetDriver.php
+++ b/cli/drivers/CraftValetDriver.php
@@ -23,6 +23,14 @@ class CraftValetDriver extends ValetDriver
      */
     public function frontControllerDirectory($sitePath)
     {
+        $dirs = ['web', 'public'];
+
+        foreach ($dirs as $dir) {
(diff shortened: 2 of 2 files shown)


mattstauffer:
@khalwat Does this conflict with other drivers that also look for public?

khalwat:
@mattstauffer It's certainly possible. Maybe what it should do is check `file_exits('craft');` to see if a file or folder named `craft` exists in the project root, and then if so, check for `public/` and `web/` and return the one it found.

khalwat:
Ah, I just checked the code @mattstauffer and it will not conflict with any other drives that detect `public/` because my modifications are only to the `frontControllerDirectory()` which doesn't affect any of the Valet detection, which is done in in the `serves()` method.

I didn't modify that, it still does exactly what I mentioned in my previous post:

```
    {
        return file_exists($sitePath.'/craft');
    }
```

.gitignore
```diff
@@ -2,3 +2,4 @@
 vendor/
 composer.lock
 error.log
+.idea
```
mattstauffer: Please drop this from the commit; we're not putting vendor-specific files in here. 

You can globally ignore the `.idea` directory from git:
https://stackoverflow.com/a/7335487

mattstauffer:
@khalwat Great! Just one quick change then we can merge this in.

khalwat:
@mattstauffer done

Issue #513, opened 2018-01-29
Enable SSI by default?


khalwat:
I have a number of websites that use [Nginx SSI](http://nginx.org/en/docs/http/ngx_http_ssi_module.html) (server side includes). Is there a chance that the default Valet Nginx config could have:

```
ssi on;
```

In the directives? Seems like it'd be a nice thing to have enabled by default in `local` dev

drbyte:
In case this suggestion is not deemed something that gets merged, one can easily set this up for themselves:

Simply manually edit either or both of these files to add `ssi on;` in the position you wish (whether that's for the `server` or `location` blocks:
`~/.composer/vendor/laravel/valet/cli/stubs/secure.valet.conf`
`~/.composer/vendor/laravel/valet/cli/stubs/valet.conf`
This will take effect in all your configs if you run `valet domain test` (replace `test` here with whatever TLD you're using with valet), as the `valet domain` command will rebuild all your configs using those stubs.

Alternatively you could do it in `~/.composer/vendor/laravel/valet/cli/stubs/nginx.conf` 
and run `valet stop && valet remove && valet install`
... or manually edit `/usr/local/etc/nginx/valet/valet.conf` and restart nginx.

Whichever of these approaches you take, you'll need to re-edit the stubs manually anytime you upgrade valet to a newer version.

khalwat:
Thanks @drbyte -- we do have this up and running locally, just hoping to get it merged into the defaults setup so that we don't have to repeat the setup for each developer/environment.

khalwat:
So a quick explanation of [SSI](http://nginx.org/en/docs/http/ngx_http_ssi_module.html). SSI is Server Side Includes, which allows the server to do some (albeit limited) processing on the files that it serves.

Apache has an SSI module, and Nginx [has SSI baked in](http://nginx.org/en/docs/http/ngx_http_ssi_module.html).

To enable it in Nginx, the following directive is needed:

```
ssi on;
```

...in any of the following blocks: `http, server, location`

The default is `ssi off;` because if you're not using it, there's no reason to have it enabled and the (nominal) overhead that comes with it to parse for SSI statements.

SSI directives in Nginx are just specially formatted HTML comments, so they don't affect any browser rendering if somehow they are served up unprocessed.

So we can do things like change what is returned in the response based on the setting of a cookie:

```
<!--# if expr="$HTTP_COOKIE=/fonts\-loaded\=1/" -->
[...]

PR #517, opened 2018-02-01
Add `ssi on;` to the default config

khalwat:
Defaulting to having `ssi on;` in the `http` block will allow those of us doing Nginx Server Side Includes to use Valet in local development.

A longer explanation of SSI, why we want it, and the impact is here: https://github.com/laravel/valet/issues/513

### DIFF
--- a/cli/stubs/nginx.conf
+++ b/cli/stubs/nginx.conf
@@ -18,6 +18,7 @@ http {
     gzip_min_length 256;
     gzip_proxied any;
     gzip_vary on;
+    ssi on;
 
     gzip_types
     application/atom+xml


drbyte:
👍 

While I don't need this myself, I have no objection to it being in Valet. 
And this is a proper way to implement it.

dennisfrank:
I would love to use SSI exactly the way @khalwat has described in #513. Any chance this PR could get merged soon?

khalwat:
Maybe @mattstauffer has some magic pull on this one :)

mattstauffer:
@khalwat Looks fine by me--any thing you can point to help alleviate the fear this may introduce speed or compatibility issues, since it's something none of us are familiar with?

drbyte:
@mattstauffer I don't see any compatibility issues or negative side-effects of this, particularly in a development environment.

mildlygeeky:
I’d also love to see this - no downside, and if there is any extra processing it must be a millisecond or two. As is, I need to code around this for local development.

khalwat:
@mattstauffer All that `ssi on;` does is enable server-side includes. 

http://nginx.org/en/docs/http/ngx_http_ssi_module.html

I can't imagine there would be any measurable speed issues (I've certainly never encountered any) in local development, and SSI doesn't introduce any compatibility issues, it just _adds_ the ability to do things like:

```
<!--# if expr="$HTTP_COOKIE=/fonts\-loaded\=1/" -->
<html class="fonts-loaded" lang="{{ craft.app.language |slice(0,2) }}" prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb#">
<!--# else -->
<html lang="{{ craft.app.language[0:2] }}" prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb#">
<!--# endif -->
```

The comment-looking things are the SSI code.

drbyte:
I've been using this simple one-line PR change in my own Valet instance.
No negative side-effects at all.

I suggest merging it.

mildlygeeky:
Ditto.

On Sep 18, 2018 at 3:52 PM, Chris Brown <notifications@github.com> wrote:


I've been using this simple one-line PR change in my own Valet instance.
No negative side-effects at all.

I suggest merging it.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/laravel/valet/pull/517#issuecomment-422528127>, or mute
the thread

PR #535, opened 2018-03-10
Preserve compatibility with PHP 5.6

### LINKED ISSUE #530: php 5.6 support
Hi,

I have valet working perfectly for php71 on my osx.

When I switch to php56 and try to run my legacy project I get the following error:

Parse error: syntax error, unexpected '->' (T_OBJECT_OPERATOR) in /Users/waveridingimac/.composer/vendor/tightenco/collect/src/Collect/Support/Collection.php on line 155

Call Stack:
    0.0035     301136   1. {main}() /Users/waveridingimac/.composer/vendor/laravel/valet/cli/valet.php:0
    0.0040     302744   2. require('/Users/waveridingimac/.composer/vendor/autoload.php') /Users/waveridingimac/.composer/vendor/laravel/valet/cli/valet.php:10
    0.0045     318648   3. ComposerAutoloaderInit5d002e971497ad1863ab7c5d29971201::getLoader() /Users/waveridingimac/.composer/vendor/autoload.php:7
    0.0127     524632   4. composerRequire5d002e971497ad1863ab7c5d29971201(???, ???) /Users/waveridingimac/.composer/vendor/composer/autoload_real.php:56
    0.0131     529144   5. require('/Users/waveridingimac/.composer/vendor/tightenco/collect/src/Collect/Support/alias.php') /Users/waveridingimac/.composer/vendor/composer/autoload_real.php:66
    0.0146     537624   6. class_alias(???, ???) /Users/waveridingimac/.composer/vendor/tightenco/collect/src/Collect/Support/alias.php:18
    0.0146     537992   7. spl_autoload_call(???) /Users/waveridingimac/.composer/vendor/tightenco/collect/src/Collect/Support/alias.php:18
    0.0146     538048   8. Composer\Autoload\ClassLoader->loadClass(???) /Users/waveridingimac/.composer/vendor/tightenco/collect/src/Collect/Support/alias.php:18
    0.0147     538216   9. Composer\Autoload\includeFile(???) /Users/waveridingimac/.composer/vendor/composer/ClassLoader.php:322

I decided to reinstall valet while being on php56 and this is where I think the issue is:

sing version ^2.0 for laravel/valet
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - illuminate/container v5.6.7 requires php ^7.1.3 -> your PHP version (5.6.33) does not satisfy that requirement.
    - illuminate/container v5.6.7 requires php ^7.1.3 -> your PHP version (5.6.33) does not satisfy that requirement.
    - illuminate/container v5.6.7 requires php ^7.1.3 -> your PHP version (5.6.33) does not satisfy that requirement.
    - Installation request for illuminate/container (locked at v5.6.7) -> satisfiable by illuminate/container[v5.6.7].

Installation failed, reverting ./composer.json to its original content.

Should valet be compatible with 5.6, please?

Thanks!

aaemnnosttv:
Valet is already compatible with PHP 5.6, but if it is installed/updated using PHP 7 or higher, Composer will install a few dependencies at newer versions that don't support 5.6. This causes Valet to break if the linked version of PHP is then switched to 5.6. The current work around is to do something like deleting Composer's global vendor directory and then run `composer update` with PHP 5.6.

This PR solves this problem by simply limiting the upper bounds of a few dependencies so that PHP 5.6 compatible versions are always installed, regardless of the version of PHP it is installed with.

Resolves #530

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,11 +25,11 @@
     },
     "require": {
         "php": ">=5.6",
-        "illuminate/container": "~5.1",
+        "illuminate/container": ">=5.1 <5.5",
         "mnapoli/silly": "~1.0",
         "symfony/process": "~2.7|~3.0|~4.0",
         "nategood/httpful": "~0.2",
-        "tightenco/collect": "^5.3"
+        "tightenco/collect": ">=5.3 <5.5"
     },
     "require-dev": {
         "mockery/mockery": "0.9.*",


drbyte:
👍

aaemnnosttv:
As a work around for this in the meanwhile, I found out you can actually set the target platform php version for Composer via your global config (works in local config too of course, but Valet is installed globally). This will ensure that all packages installed are compatible with it, even if it's not the version of the php binary you're currently using.

To do this, run 

```
composer global config platform.php 5.6.36
composer global update
```

Now you're free to switch PHP versions all over the place and your installed (global) dependencies will stay compatible with php 5.6.36 (or whatever version you want to use there).

eberkund:
Couldn't this also be done with the `platform` key? https://getcomposer.org/doc/06-config.md#platform

    "config": {
        "preferred-install": "dist",
        "sort-packages": true,
        "platform": {
            "php": "5.6"
        }
    },

drbyte:
Both @aaemnnosttv and @eberkund are suggesting essentially the same thing, although I think they envision it being done in different places (one in one's own global composer config, the other in Valet's composer.json).

I have no objection to it "being done" ... however, my preference is that a person manually imposes it upon themselves, not on everyone.

Backward compatibility has its place, but the further we get away from the 5.6 expired-support-date the less I want to be having that old junk cluttering my dev environment.

aaemnnosttv:
Closing this out as I don't think the changes are necessary or preferable anymore.

Regarding the use of `config.platform` - to @drbyte 's point - it can only be used at the root/project level; dependencies cannot impose this as a requirement. In other words, even if Valet defined this in its `composer.json`, it would have no effect when installing it as a dependency.

Using `config.platform` is the solution though, it's just that it can only be implemented on the user end. See [my comment](https://github.com/laravel/valet/pull/535#issuecomment-421621081) above for steps to set this up in a way that won't impose this requirement on any other globally installed dependencies.

PR #542, opened 2018-03-27
Static file serving for subfolders

chejaziuhin:
This allows static file serving of 'index.php' and 'index.html' files from subfolders when you have a root 'public' folder.

### DIFF
--- a/cli/drivers/BasicValetDriver.php
+++ b/cli/drivers/BasicValetDriver.php
@@ -25,9 +25,9 @@ class BasicValetDriver extends ValetDriver
      */
     public function isStaticFile($sitePath, $siteName, $uri)
     {
-        if (file_exists($staticFilePath = $sitePath.'/public'.$uri.'index.html')) {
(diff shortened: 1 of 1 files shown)


eberkund:
How will this handle different separators, like `example.test/subdir/` vs `example.test/subdir`

chejaziuhin:
@eberkund I made a change to account for those two different use cases - [26d5d71](https://github.com/laravel/valet/pull/542/commits/26d5d715a8fb680ad402bac8cab793cda05b6643). Both urls should work the same now.

kclayton925:
Any update on this? A designer I work with would benefit from this.

drbyte:
This makes sense. 
And it wouldn't conflict with existing drivers that already have a method for detecting static files, so it's not risking breaking those common configs.

davidhsianturi:
This is very useful, especially if you use Laravel API generator https://github.com/mpociot/laravel-apidoc-generator

mattstauffer:
Has anyone been using this regularly without it interrupting the serving of other static sites/etc.?

drbyte:
I haven't had any problems serving Laravel and non-Laravel sites with this change in place.

PR #545, opened 2018-03-30
fix for new php versions/paths from php installed via brew

james2doyle:
Previous versions of brew installed PHP like so:

```
/usr/local/opt/php71/bin/
```

Now they look like this:

```
/usr/local/opt/php@7.1/bin/
```

These changes just tweak the version check to remove `@` and `.` from the paths. So this should be backwards compatible with older installations as well.

*I was unable to get the tests running, but I applied these patches locally and went from not being able to install to a valid installation.*

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -182,7 +182,7 @@ class Brew
         $resolvedPath = $this->files->readLink('/usr/local/bin/php');
 
         return $this->supportedPhpVersions()->first(function ($version) use ($resolvedPath) {
-            return strpos($resolvedPath, "/$version/") !== false;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Is this still needed after merging: https://github.com/laravel/valet/pull/545

james2doyle:
Created #552

jimrubenstein:
What are the chances I composer update during this..lol. Dug in and fixed the syntax error on my own :|

Line 185 of Brew.php contains: 

`return strpos(preg_replace('/([@|\.])/', '', $resolvedPath, "/$version/")) !== false;`

needs to be

`return strpos(preg_replace('/([@|\.])/', '', $resolvedPath), "/$version/") !== false;`

If someone happens across this before a fix is pushed out.

james2doyle:
@jimrubenstein should be good with #552

PR #546, opened 2018-04-01
Fixes #521

### LINKED ISSUE #521: php72 stops after valet secure
When I run `valet secure` nginx stops working and shows a missing certificates error. 
In order to fix it, I had to use a workaround suggested by @zhse in #321.

matrunchyk:
This PR fixes inability to run `valet secure` (see #521) and nginx work. 
The problem was because valet config is stored under `/var/root/.valet` folder while valet tries to create certificates under `/Users/johndoe/.valet/Certificates` which obviously doesn't work.
Also, it fixes `error_log` path which was located under `/var/root/.valet` folder which is not accessible because `nginx` runs under `johndoe` user.

### DIFF
--- a/cli/Valet/CommandLine.php
+++ b/cli/Valet/CommandLine.php
@@ -25,7 +25,7 @@ class CommandLine
      */
     function quietlyAsUser($command)
     {
-        $this->quietly('sudo -u '.user().' '.$command.' > /dev/null 2>&1');
+        $this->quietly('sudo '.$command.' > /dev/null 2>&1');
     }
 
     /**
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -25,7 +25,7 @@ server {
     }
 
     access_log off;
-    error_log VALET_HOME_PATH/Log/nginx-error.log;
+    error_log /usr/local/var/log/nginx/error.log;
 
     error_page 404 VALET_SERVER_PATH;
 


matrunchyk:
@taylorotwell Please take a look if you have a moment.

cli/stubs/secure.valet.conf
```diff
@@ -25,7 +25,7 @@ server {
     }
 
     access_log off;
-    error_log VALET_HOME_PATH/Log/nginx-error.log;
+    error_log /usr/local/var/log/nginx/error.log;
```
drbyte: meh -- I don't like this change. 

My valet properly writes nginx logs to `VALET_HOME_PATH/Log/nginx-error.log`, which makes it very quick and easy to do troubleshooting, because I can quickly find that file from my home directory, without having to traverse into system directories.

I'm using Valet 2.0.8 without any of the symptoms you describe in 521 or this PR.

matrunchyk: Hey @drbytem thanks for your comment!
I'm using 2.0.8 as well, however, even when having `/var/root/.valet` owned by my user, I can't read `/var/root/.valet/Log/nginx-error` file (permission error). Nginx, valet, php has been reinstalled from scratch (fresh install), so I'm not sure what else I have to do to make it working w/o proposed changes... 
Do you have `.vallet` under `/var/root`?

drbyte: No, my `.valet` is in my own user directory: `~/.valet`, hence no permission issues.

matrunchyk: @drbyte How did you install it under your user directory? I run `valet` and it tries to install under `/var/root` because bash script does `sudo valet`. Did you do manual steps while installing 2.0.8?

drbyte: I've installed and re-installed so many times when doing writing and testing of various valet PRs I don't remember what I might have done differently: as far as I can recall it was nothing special.

drbyte: Also: never install valet while in sudo or as root.

cli/Valet/CommandLine.php
```diff
@@ -25,7 +25,7 @@ function quietly($command)
      */
     function quietlyAsUser($command)
     {
-        $this->quietly('sudo -u '.user().' '.$command.' > /dev/null 2>&1');
+        $this->quietly('sudo '.$command.' > /dev/null 2>&1');
```
jmarcher: The method signature does not make sense anymore.

matrunchyk: Why?

cli/stubs/secure.valet.conf
matrunchyk: @drbyte running `valet` in fact does `sudo` within a shell script. That's because it fails.

matrunchyk: Can anybody test valet by reinstalling from scratch? Because it just doesn't work.

cli/Valet/CommandLine.php
jmarcher: Well as the method signature says: quietlyAsUser you just deleted the code that executes the code as a user, therefore, is the signature no longer valid for its functionality.

matrunchyk: Ahh, yeah

PR #548, opened 2018-04-02
View realtime nginx error log

gahlawat:


drbyte:
This could be handy.  I would use it, as it would save me remembering what path to `tail`.

However, there are additional logs that one might wish to see, related to valet, so it would be fair to not limit this to nginx logs.

I'd like to see `valet logs` support passing a parameter of which log to `tail`, and then it could offer any of:
- nginx-error.log
- nginx access log
- php-fpm error log
- dnsmasq log

mattstauffer:
Closed because https://github.com/laravel/valet/pull/757

PR #549, opened 2018-04-02
Update Brew.php - Homebrew\Core Package Names

epreston:
Update supportedPhpVersions to include the new names used by homebrew now that the php packages have migrated to 'core'.   `php71` has become `php@7.1` etc

The new names are put at the front of the list to optimise the loops that use these collections.

Anyone who installs or updates 'homebrew php' from will start using the new packages names which breaks parts of valet.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -52,7 +52,7 @@ class Brew
      */
     function supportedPhpVersions()
     {
-        return collect(['php', 'php72', 'php71', 'php70', 'php56']);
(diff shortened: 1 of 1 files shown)


JesseLeite:
I can't valet restart without `php@7.1` being in list since [homebrew's php stuff got moved to core](https://github.com/Homebrew/homebrew-php/blob/master/README.md) with new naming convention of `php@version`.  Added it locally and valet runs fine now 👍

Issue #558, opened 2018-04-06
2.0.10 - valet restart, no longer restarts PHP fpm process


epreston:
This release no longer finds and restarts php when running `valet restart`

I had this all working in #549 now its not.

RobvH:
```sh
╭─rob.vanhoose@endor: brstack/RobvH/valet  master
╰─▶ ps -ax | grep php-fpm
35289 ??         0:00.07 /usr/local/opt/php/sbin/php-fpm --nodaemonize
35310 ??         0:00.00 /usr/local/opt/php/sbin/php-fpm --nodaemonize
35311 ??         0:00.00 /usr/local/opt/php/sbin/php-fpm --nodaemonize
╭─rob.vanhoose@endor: brstack/RobvH/valet  master
╰─▶ valet restart        
Restarting php...
Restarting nginx...
Valet services have been restarted.
╭─rob.vanhoose@endor: brstack/RobvH/valet  master 
╰─▶ ps -ax | grep php-fpm
36589 ??         0:00.06 /usr/local/opt/php/sbin/php-fpm --nodaemonize
36596 ??         0:00.00 /usr/local/opt/php/sbin/php-fpm --nodaemonize
36597 ??         0:00.00 /usr/local/opt/php/sbin/php-fpm --nodaemonize
```

Restarted perfectly. Can you provide more information about your environment or any errors you're getting?

epreston:
It restarted perfectly only because your php is called “php”.  Link php@7.1 or php@7.2.  You’ll notice the “restarting php” message go away. Also any function like isphpinstalled will return false.

epreston:
I don't mind how we get a solve on this.  Use the regular expression fix in all places where the array is referenced and my test environments will work.  That would be a modification to `hasInstalledPhp` as well introduce a regular expression that works in all cases.

The history on this seems to be both of us quickly putting up patches to get back to work.  Due to the time differences, nobody seeing what everyone is doing.  I haven't helped the situation with my lack of confidence introducing changes to the unit tests.  It also looks like Taylor and Adam didn't have much comms in the state it was left between releases and versions.

How can i help to have this sorted over the weekend?

RobvH:
If you want to use php@7.1 you would normally unlink 7.2 and force link the cask, no? Is there a better way? I always need php and php-fpm to be called by their proper names for my scripts, ides, etc to work. 

[...]

PR #559, opened 2018-04-06
Revert "Fix (really) changes where brew php paths now include @ and semver-st…"

epreston:
Reverts laravel/valet#553

We should be updating the supported versions array.  This change reverted the fix in place already.

Doing a regex here does not address the other place where the array is used (hasInstalledPhp) causing valet not to restart php.  It may leave people who haven't or can't update the brew package source in a bad place.  Also, we want the display string to represent the php "flavour" installed and detected for that user path (ie. php71 or php@71) so it is cleared what is linked in the context of the web server.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -52,7 +52,7 @@ class Brew
      */
     function supportedPhpVersions()
     {
-        return collect(['php', 'php72', 'php71', 'php70', 'php56']);
+        return collect(['php', 'php@7.2', 'php@7.1', 'php@7.0', 'php@5.6', 'php72', 'php71', 'php70', 'php56' ]);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


epreston:
See #558

epreston:
Unit tests need to be updated to include a handler for 'php@7.2'

RobvH:
@epreston Great find! It’s interesting that no test is breaking because of this.

RobvH:
@epreston that error from Mockery means the test failed bc you failed the expectation’s signature. It doesn’t mean the test needs a handler. It’s a bit confusing in its wording. 

Hope this helps.

epreston:
Ok. So am I to read that as the Travis builds are not respecting the environment variable to install specific versions of php and failing when they find php@7.2 instead of the environment they are testing?

epreston:
See #561

PR #564, opened 2018-04-09
Fix Regression from 2.0.9 and Modernise Unit Tests

epreston:
Remove the commit made to satisfy the unit tests.

Ensure support for environments using the `homebrew\core` tap and those still on packages from the deprecated and mostly deleted `homebrew\homebrew-php` tap.

Ensure support for pinned and unpinned versions of php at 7.2.

Modernise unit tests to mock commands that could be run today in the command line as confirmation.  The exception to this are the "tap" commands that have always been illustrative.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -52,7 +52,7 @@ class Brew
      */
     function supportedPhpVersions()
     {
-        return collect(['php', 'php72', 'php71', 'php70', 'php56']);
+        return collect(['php', 'php@7.2', 'php@7.1', 'php@7.0', 'php@5.6', 'php72', 'php71', 'php70', 'php56' ]);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


mattstauffer:
@epreston Sorry for the delay; taking a look now. I looked at that issue and it seems like it was resolved.

I'll admit, I haven't fully followed the changes here. Could you tell me a little bit more about this?

Thanks!

drbyte:
@epreston I think some of these core changes are already integrated. However, some of these tests could be useful.  Do you mind addressing the merge conflicts and then we can look at what's left?

epreston:
@drbyte  The merge conflicts come from the replacement of lookup tables (arrays) with some regular expressions.  I feel that they are not clear, slow, and inflexible.  At the time, pull requests where being merged by maintainers so quickly without confirmation, it was chaos.  I was not familiar with one of the unit test tools and all my test machines were setup differently.  It took time for a quality solve.

@mattstauffer there was a name standard change and repo change on homebrew php at the same time.  There were a number of targets and setups to maintain compatibility with.  There were a few configuration states (combinations of php versions and installation sources) to add support for.  Taylor had done this with a clear table lookup that could be extended.  Information was unclear because a fix for some was a breaking change for others depending on what pieces they had updated.  It took these changes, unit tests, setting up a mirror of the Jenkins, and a few test macs to confirm.  I still worry that I missed something.

mattstauffer:
OK. We have finally settled on and modernized how our stack handles detection of PHP version.

@epreston If you'd like to fix up this PR I'd love to review it again, although it might make more sense as a new PR.

However, it's a year old. If you'd like to just let it die, please feel free to just close it. Your call--either way thanks for your work on it!

PR #565, opened 2018-04-09
Fix return string with php-fpm config path for linked PHP

jesusgoku:


jesusgoku:
@adamwathan It has been useful to refactor the method. It was not necessary to go through the entire list of PHP versions supported.

gonzunigad:
👏

eaguad1337:
Thanks you Jesus Christ!

PR #573, opened 2018-04-23
Allow http2 server push for secured domains [BREAKING]

corneliusio:
This is a very small addition that allows sites running on valet to test out preloading static assets.

However, `http2_push_preload` was not added into NGINX until 1.13.9 (released February 20th, 2018) which would mean this directive breaks secured sites on any older version of NGINX.

Figured this could be merged into the next major release?

### DIFF
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -10,6 +10,7 @@ server {
     root /;
     charset utf-8;
     client_max_body_size 128M;
+    http2_push_preload on;
 
     location /VALET_STATIC_PREFIX/ {
         internal;


drbyte:
I have no objection to this.

If people are in the habit of running `brew update` and `brew upgrade` then they'll have a modern version of nginx which should support this by now. (Mine's 1.15.3 today.)

drbyte:
I've been using this in my daily environment since the PR was posted. No negative side effects.

I'd say it's safe to merge.

mattstauffer:
@drbyte Sounds good! I don't think we have the concept of a version constraint on Nginx right now so I wonder whether we need to add that. Will try to take a look a bit later if I can

drbyte:
Meh - as for version constraints on Nginx, right now I don't think Valet should care if someone's using an Nginx version older than Feb 2018 now that we're into Sept 2019.
Particularly since Homebrew can only install 1.17.x, and Nginx only makes updates to the current version. (While Nginx will offer paid commercial support for 2 years after a given release, if someone's engaged in a support contract like that then if they're using Valet then they're already going to be heavily customizing their own Valet to accommodate that anyway.)
Hence I think it's moot.

drbyte:
NOTE for anyone upgrading to 2.4.0 where this was released: it will take effect automatically on each new domain you run `valet secure` on.  You can optionally re-secure all your domains by running `valet tld test` (where `test` equates to your .test tld in use; substitute accordingly)

PR #587, opened 2018-05-18
Added support for Directory Listing

RoyVoetman:


drbyte:
Hi @RoyVoetman ,
In this era of web security and minimizing exposure of server-side information to end users, this seems like an odd request. Most ops/devops people are intentionally disabling directory listings on servers.

I realize Valet is not intended as a production tool, so in very rare cases maybe a directory listing might be useful for debugging, but IMO that'd be better done as a specific request rather than an always-on thing like your PR is proposing.

What's the rationale, ie: business problem you're trying to solve with this PR?

RoyVoetman:
I am not tying to solve any business logic, I just thought it was an useful thing to have for debugging purposes.

I could add a command line command to toggle the Directory Listing and set the default to Off?

drbyte:
The discussion in #610 offers an alternate approach which may support what you were thinking of.

PR #590, opened 2018-05-22
Update the tap list, current taps deprecated.

robjstanley:
On performing `valet install` the current version of valet will error due to deprecated taps, previous taps moved to `homebrew/homebrew-core`.
```
> valet install
Stopping nginx...
Installing nginx configuration...
Installing nginx directory...
Installing php71...
Error: homebrew/dupes was deprecated. This tap is now empty as all its formulae were migrated.
Error: homebrew/versions was deprecated. This tap is now empty as all its formulae were migrated.
Error: homebrew/php was deprecated. This tap is now empty as all its formulae were migrated.```

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -11,7 +11,7 @@ class PhpFpm
     var $brew, $cli, $files;
 
     var $taps = [
-        'homebrew/dupes', 'homebrew/versions', 'homebrew/homebrew-php'
+        'homebrew/homebrew-core'
     ];
 
     /**


mattstauffer:
@robjstanley Good catch. Do you know whether we'll need to check someone's version of Homebrew in order for this not to break it for folks on older versions? Or does version of Homebrew not matter for this?

drbyte:
@mattstauffer I'm pretty sure it doesn't matter.

mattstauffer:
@drbyte <3

PR #591, opened 2018-05-31
Recursively search for custom valet drivers

morrislaptop:
This PR searches directories inside `~.valet/Drivers` so that users can easily bring in custom drivers from other packages. 

Before:

    Clone the directory and then copy CustomValetDriver.php into your ~/.valet/Drivers directory

After:

    git clone https://github.com/DCzajkowski/DistValetDriver.git ~/.valet/Drivers/DistValetDriver`

### DIFF
--- a/cli/drivers/ValetDriver.php
+++ b/cli/drivers/ValetDriver.php
@@ -113,12 +113,14 @@ abstract class ValetDriver
 
         $drivers = [];
 
-        foreach (scandir($path) as $file) {
-            if ($file !== 'ValetDriver.php' && strpos($file, 'ValetDriver') !== false) {
-                require_once $path.'/'.$file;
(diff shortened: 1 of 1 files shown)


morrislaptop:
No?

drbyte:
Tests good. 
Great idea -- adds simplicity!

mattstauffer:
Whoops, not sure how I missed this one entirely. Thanks @morrislaptop, @drbyte, and @morrislaptop's glorious mustache.

drbyte:
TECH NOTE: Since Valet 2.1.0 in Sept 2018, the config dir is now `~/.config/valet` and not `~/.valet` as mentioned in the comments above.
Other than that, the concept still works as intended.

PR #601, opened 2018-06-19
Remove port from url

joinsito:
Hi,

I've found an issue when trying to access my application through valet specifying the port in the address, valet is not able to find the right folder in this case.

My motivation for specifying the port is that I'm accessing my application through a reverse proxy in a remote server listening on a port different from 443 or 80.

 I've fixed it by removing port from $_SERVER['HTTP_HOST'].

### DIFF
--- a/server.php
+++ b/server.php
@@ -65,7 +65,7 @@ $uri = urldecode(
 
 $siteName = basename(
     // Filter host to support wildcard dns feature
-    valet_support_wildcard_dns($_SERVER['HTTP_HOST']),
+    valet_support_wildcard_dns(explode(':',$_SERVER['HTTP_HOST'])[0]),
     '.'.$valetConfig['domain']
 );
 
--- a/server.php
+++ b/server.php
@@ -65,7 +65,7 @@ $uri = urldecode(
 
 $siteName = basename(
     // Filter host to support wildcard dns feature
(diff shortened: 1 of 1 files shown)


server.php
```diff
@@ -65,8 +65,8 @@ function valet_default_site_path($config)
 
 $siteName = basename(
     // Filter host to support wildcard dns feature
-    valet_support_wildcard_dns($_SERVER['HTTP_HOST']),
-    '.'.$valetConfig['tld']
+    valet_support_wildcard_dns(explode(':', $_SERVER['HTTP_HOST'])[0]),
+    '.'.$valetConfig['domain']
```
drbyte: Hi, `domain` here should be changed to `tld`

joinsito: Noted, thanks.

drbyte:
@joinsito What are your thoughts on whether it's best to remove the port before passing it into the `valet_support_wildcard_dns()` function, vs doing it inside the `valet_support_wildcard_dns()` function just before it returns the parsed domain, as in #630 ?

joinsito:
@drbyte, makes sense to do it inside the function as in #630. Happy to see It merged!

drbyte:
Great!  Thanks for your contribution @joinsito !

PR #606, opened 2018-07-10
Use `isActualFile` instead of `file_exists`

imacrayon:


cli/drivers/BasicValetDriver.php
```diff
@@ -25,7 +25,7 @@ public function serves($sitePath, $siteName, $uri)
      */
     public function isStaticFile($sitePath, $siteName, $uri)
     {
-        if (file_exists($staticFilePath = $sitePath.'/public'.$uri)) {
+        if ($this->isActualFile($staticFilePath = $sitePath.'/public'.$uri)) {
```
mattstauffer: isActualFile also checks whether it's not a directory; do you know whether this currently should be serving directories? If so, that would make this a bad change.

imacrayon: hmm - that's a good point. It looks like most of the other drivers use `isActualFile` and I assumed that `isStaticFile` only cared about files, but maybe the Basic Driver is intended to be more flexible?

mattstauffer: Since neither of us know, are you OK if we close it until someone else comes along and has a strong opinion? :)

imacrayon: According the the Valet docs for custom drivers: 
> The `isStaticFile` should determine if the incoming request is for a file that is "static", such as an image or a stylesheet. If the file is static, the method should return the fully qualified path to the static file on disk. If the incoming request is not for a static file, the method should return  false.
Based off that it sounds like it's not suppose to be a directory.

drbyte: (note: I don't have a bunch of valet sites without a provided driver to do more aggressive testing on, so my comments below aren't based on actual tests.)

Like @imacrayon I think the namings of `isStaticFile` and `isActualFile` should both reflect the actual operation of the methods; however, in practice, `isStaticFile` appears to be regularly defined to deal with both custom directory names as well as files. 
That might be an error in initial implementation, but this `file_exists` check has been here since the earliest versions of Valet, even before it was refactored into its current state.

Some drivers, including the [LaravelValetDriver](https://github.com/laravel/valet/blob/master/cli/drivers/LaravelValetDriver.php#L30) specifically add an `is_file` constraint/check. But other drivers add a bunch of additional logic to filter out a bunch of directories ... which implies also that perhaps they're also glad to serve some directories if mentioned. (Lots of other ways to interpret this too, but no need to write a book here.)
That said, I'm glad that other drivers do specifically use `isActualFile`, since that provides appropriate specificity for their use-cases.

Because of this, and that the change is probably BC for edge cases and lesser-maintained drivers, I'm inclined to say this should be canceled. I do wish the naming was more semantically reflective, but in many ways it's trivial too.

mattstauffer:
Closing based on our concerns for BC. Thanks for the PR anyway @imacrayon!

PR #612, opened 2018-08-13
load drivers dynamically

buddhaCode:
Load drivers dynamically. No need to keep require and driver list up to date manually.


mattstauffer:
Four years later.. thanks for this PR! I believe it was manually listed out on purpose in order to be able to make sure we don't have any race conditions.

PR #615, opened 2018-08-18
Add missing imports for helpers

aaemnnosttv:
This PR fixes a bug introduced in #602 which namespaced all helper functions, but didn't import them in `valet.php`.

E.g.
```
PHP Fatal error:  Uncaught Error: Call to undefined function info() in /Users/aaemnnosttv/.composer/vendor/laravel/valet/cli/valet.php:234
```


mattstauffer:
Great catch--thanks!

Issue #620, opened 2018-08-28
Domain command no longer outputs bare TLD 


aaemnnosttv:
As of #241 `valet domain` no longer outputs the plain TLD used for the domain, but a more informational message.

This breaks a package that I maintain that uses this command as part of a scripted install. See aaemnnosttv/wp-cli-valet-command#39

With the addition of the `tld` command, can we keep `domain` outputting the bare TLD for backwards compatibility?

If this is the wrong way to get configuration values from Valet, what's the best/safest future-proof way to do so?

aaemnnosttv:
Thank you for the PR, that's much appreciated.

I'm still interested to know what the maintainers have to say in regard to the questions above though as it goes beyond just the issue with my package.

mattstauffer:
@aaemnnosttv From my understanding, the maintainers are open to PRs that offer functionality that makes sense, doesn't overly complicate the codebase, is easy to grok and test and implement and maintain, and isn't more drivers. So if you want to build a feature that exposes data for consumption by a third-party tool, I can't guarantee they'll say yes, but I can say that's probably going to be your best option. That, or just merge @drbyte's PR.

FWIW, while he doesn't have commit rights, @drbyte is by far the most active helper on this repo, so him offering help in this way is one of the best things we have going for us.

aaemnnosttv:
Thanks Matt. I wasn't thinking that building something for getting configuration values is really necessary.

I took a look at if it might be possible to keep the `domain` command the way it was without listing it as a normal command but it doesn't look like a simple change. Preserving BC would require changing the `tld` command as well and I'm not suggesting that since it has already been accepted as-is.

Everything I need is in `~/.valet/config.json`. As long as I can rely on that not moving, then I think reaching for it there is probably the way to go going forward (this is what [Lambo](https://github.com/tightenco/lambo) does).

Thanks to @drbyte and everyone else who helps maintain this!

mattstauffer:
@aaemnnosttv :) ❤️

drbyte:
@aaemnnosttv wrote:
> Everything I need is in ~/.valet/config.json. As long as I can rely on that not moving, then I think reaching for it there is probably the way to go going forward (this is what Lambo does).

[...]

PR #628, opened 2018-09-14
Remove extra message text from TLD output, for valet share cmd

drbyte:
The `valet share` command relies on the `tld` command outputting only the TLD and not a status message ... much like was discussed in #620.
This removes the status message that was added with the recent tld update in #241. 

(Other alternatives were considered including special treatment if the "domain" alias was detected etc, but in the end I know that the preference is to keep the code simpler, and removing the status message felt like the best approach after all.)

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -64,7 +64,7 @@ if (is_dir(VALET_HOME_PATH)) {
         }
 
         if ($tld === null) {
-            return info('Valet is configured to serve for TLD: .'.Configuration::read()['tld']);
+            return info(Configuration::read()['tld']);
         }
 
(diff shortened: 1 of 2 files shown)


mattstauffer:
So if I am reading this right, the change broke both an external package and an internal API; this reverts it and will fix both? Thanks!

drbyte:
Yes, that's the goal of this PR.

I was testing `valet share` for someone yesterday and ran into this issue on `master`. This PR resolves that, and has the side benefit of removing the issue reported by the other package.

mattstauffer:
Thanks @drbyte!

PR #629, opened 2018-09-14
Fix errors from relocation of config dir

drbyte:
This PR fixes the PHP errors reported in https://github.com/laravel/valet/issues/620#issuecomment-417365462 which were a result of merging #449 

From the tests I've been doing with both "old" and "new" locations of the config files, this PR seems to resolve those.

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -41,14 +41,14 @@ class Configuration
      */
     function createConfigurationDirectory()
     {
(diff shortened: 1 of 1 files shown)


mattstauffer:
I started hacking around on this but I'm so so so glad you were able to get to it. I'll merge and we can all test around a bit; if we have this fixed and your other PR then I think we're ready for a tag, finally!

PR #656, opened 2018-10-17
Only serve statically cached files on GET requests for Statamic

jasonvarga:
Statamic's Live Preview feature does a POST request to an item's actual URL with updated data so you can preview the item. If the item had already been statically cached, the same html file would load every time, and not reflect your changes.

We ship Statamic with sample config files for this situation. An htaccess for Apache, an nginx conf for Nginx, and this PR is the equivalent for Valet users.


mattstauffer:
❤️  Thanks!

jasonvarga:
Thank you!

PR #658, opened 2018-10-24
Remove deprecated testcase implementation

samnela:


samnela:
Because Travis did not trigger, I closed and reopened this PR

mattstauffer:
As far as I know, this is only relevant for PHPUnit 6+, and this project uses PHPUnit ~5.7 at the moment. I assume you ran into this when using your system phunit instead of when using the project PHPUnit.

If you want to pair this with an update to the version of PHPUnit in the composer.json (and fixing anything that may break) that would be helpful. Other wise we can close this.

drbyte:
Note also, PHPUnit 6 requires PHP 7.0+, but valet intentionally supports PHP 5.6 at this time.

PR #668, opened 2018-11-11
Pre-create the parent .config dir if needed

drbyte:
On a fresh OSX install, the ~/.config/ dir doesn't always exist. 
This improves the installation process by checking and creating it if needed.

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -41,6 +41,8 @@ class Configuration
      */
     function createConfigurationDirectory()
     {
+        $this->files->ensureDirExists(preg_replace('~/valet$~', '', VALET_HOME_PATH), user());
+
         $oldPath = posix_getpwuid(fileowner(__FILE__))['dir'].'/.valet';
 
         if ($this->files->isDir($oldPath)) {
--- a/tests/ConfigurationTest.php
+++ b/tests/ConfigurationTest.php
@@ -28,6 +28,7 @@ class ConfigurationTest extends PHPUnit_Framework_TestCase
     public function test_configuration_directory_is_created_if_it_doesnt_exist()
     {
         $files = Mockery::mock(Filesystem::class.'[ensureDirExists,isDir]');
+        $files->shouldReceive('ensureDirExists')->once()->with(preg_replace('~/valet$~', '', VALET_HOME_PATH), user());
         $files->shouldReceive('ensureDirExists')->once()->with(VALET_HOME_PATH, user());
         $files->shouldReceive('isDir')->once();
         swap(Filesystem::class, $files);


cli/Valet/Configuration.php
```diff
@@ -41,6 +41,8 @@ function install()
      */
     function createConfigurationDirectory()
     {
+        $this->files->ensureDirExists(preg_replace('~/valet$~', '', VALET_HOME_PATH), user());
```
mattstauffer: it took me a minute to figure out what was going on here. Is there any simpler way to trim this off the end? If not, that's OK, but just for future folks who get lost like me lol

tests/ConfigurationTest.php
```diff
@@ -28,6 +28,7 @@ public function tearDown()
     public function test_configuration_directory_is_created_if_it_doesnt_exist()
     {
         $files = Mockery::mock(Filesystem::class.'[ensureDirExists,isDir]');
+        $files->shouldReceive('ensureDirExists')->once()->with(preg_replace('~/valet$~', '', VALET_HOME_PATH), user());
```
mattstauffer: I wish we could test this--I almost wonder whether we should hard code the path, even if that means having to re-write this test if it ever changed

cli/Valet/Configuration.php
drbyte: I considered just str_replace, but the regex with `$` forcing the end of the string felt safer.

Granted, it's unlikely there'd be multiple `/valet` instances in it, so str_replace is probably fine.
Which direction would you like us to take?

tests/ConfigurationTest.php
drbyte: Ya, I was trying to remain consistent between the code and the test, as well as with the code on the next line.
A str_replace would be more readable here too.

cli/Valet/Configuration.php
mattstauffer: ughghghghgh i want the clearer option that's less precise but i feel like it will bite us later. Let's keep it. <3

tests/ConfigurationTest.php
mattstauffer: yah, i think the bummer here is that we're missing a test that the preg_replace is functioning the way we want. I think if I were writing it I would hard code the string here and let this be a test that just breaks later if we change the directory, and then we can just change this test later if we do that. :) Thoughts?

drbyte: The downside to hard-coding it here is that here's how the constant is derived:

https://github.com/laravel/valet/blob/25d6e47eb1506f7763d2acde5705ca07d3d56836/cli/includes/helpers.php#L13

I guess it's shorter if I just extract that out. 

But my thinking has been that then it's very disconnected from the more important next line.

mattstauffer: Ahhhh ok. That makes a bit sad but I get it . :) Thanks!

PR #669, opened 2018-11-11
Sanity check for tld config

drbyte:
In some cases of mixed older versions and configs, the 'tld' config entry was not set.
This now updates (sets) the `tld` config entry if it is not set.

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -63,13 +63,16 @@ $app->command('install', function () {
  */
 if (is_dir(VALET_HOME_PATH)) {
     /**
+     * Upgrade helper: ensure the tld config exists
+     */
+    if (empty(Configuration::read()['tld'])) {
+        Configuration::writeBaseConfiguration();
+    }
+
+    /**
      * Get or set the TLD currently being used by Valet.
      */
     $app->command('tld [tld]', function ($tld = null) {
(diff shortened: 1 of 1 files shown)


drbyte:
Those are fair questions. 

My observations:
- as to other config/upgrade checks, this should only fire if it's determined that valet is already installed. All the other checks done for `valet install` are related to ensuring certain directories exist, so I'm not seeing any negative side effects.
- yes it does force an extra hit to the filesystem, but calls to `valet` are not high-traffic, so the impact is minimal
- calling `secure`, `unsecure`, and `open` depend on the `tld` entry being present. Only the commands for `tld`/`domain` and `install` where ensuring its presence, which meant upgrades were likely to miss it (unless the person upgrading actually ran `valet install`)
  I didn't want to copy/paste this "if" statement into those other 3 functions (thinking DRY here), so opted to make it global, as the only side-effect is that it runs anytime the `valet` command is called once valet is installed. That said, we can go that route instead. 

I guess another alternate strategy would be to edit the `read()` method to automatically provide a fallback if no `tld` entry is found when requested.

mattstauffer:
I think those are great answers and definitely enough for now. We may choose to re architect how all this validation works soon but this looks solid for now. Thanks Chris!!

PR #670, opened 2018-11-11
Change valet share to use port 60

drbyte:
Since the current TCP port 88 clashes with Kerberos, changing it to port 60 which IANA shows as unassigned, and other databases show as not being used in general practice.

Several alternates were researched, but show (although limited use, nevertheless possible) clashes with other systems, including ports 47, 81-90, 8080-8090, and 100. Various sources were referenced, and the overall conclusion is that 60 seems to be the safest and least likely to conflict.

Tested against a few limited firewall configs without issue.
Tested with both fresh new valet installs and an older install upgraded from 2.0.x.

### DIFF
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -43,7 +43,7 @@ server {
 }
 
 server {
-    listen 88;
+    listen 60;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
--- a/valet
+++ b/valet
@@ -41,11 +41,11 @@ then
         fi
     done
 
-	# Decide the correct PORT to use according if the site has a secure 
+	# Decide the correct PORT to use according if the site has a secure
 	# config or not.
 	if grep --quiet 443 ~/.config/valet/Nginx/$HOST*
 	then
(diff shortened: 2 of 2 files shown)


drbyte:
In my experience this fixes the concern raised by @simonhamp and @Danny-G-Smith in:
https://github.com/laravel/valet/pull/630#issuecomment-437593570

mattstauffer:
@simonhamp and @Donny-G-Smith: do either of y’all know how to test this directly From this branch? I plan to test it myself but if either of y’all have the ability to let us know if it fixes your issues that would awesome. Thanks!

@drbytezc thanks as always for your amazing work!!!

Danny-G-Smith:
Turning on screen sharing will do it every time, in my experience, so it should be easy to check, at least for me.

mattstauffer:
@Danny-G-Smith Would you point your global Valet record to `dev-master`, update it, and check this out? Let me know if that's something you're comfortable doing. Thanks!

Danny-G-Smith:
It is still working for me after updating.  Good work!

simonhamp:
@mattstauffer Yesssss! This absolutely works. Thanks for getting this merged so rapid folks!

PR #676, opened 2018-11-23
Disable xdebug

driesvints:
This should speedup builds a bit and makes it more consistent with other Laravel libraries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
 sudo: false
 
 before_install:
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks @driesvints!

driesvints:
@mattstauffer thanks for merging!

FYI: there's a stale branch still open on the repo which could be deleted I think: https://github.com/laravel/valet/branches

Issue #678, opened 2018-12-04
Home directory using spaces breaks installation.


sahibalejandro:
Let's say my home directory is like this `/Users/John Doe/` this causes valet installation to fail when running the `sudo` commands like this `sudo -u John Doe ...`.

The solution is just to quote the username, I already did it on my local valet installation and I can send a PR.

drbyte:
This can be closed

PR #679, opened 2018-12-04
Surround username with quotes (in case it has spaces)

sahibalejandro:
Fix issue #678


drbyte:
Thanks @sahibalejandro !

PR #683, opened 2018-12-07
Support PHP 7.3

tillkruss:
I was getting a "502 Bad Gateway" error after upgrading to PHP 7.3. This seems to fix it for me.

The error message was:

```
connect() to unix:/Users/<user>/.config/valet/valet.sock failed
```

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -7,8 +7,21 @@ use DomainException;
 
 class Brew
 {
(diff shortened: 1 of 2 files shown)


mattstauffer:
Once the tests go green on CI, I'll merge and test it out a bit, and then we can release if it seems to work OK for me (you can update yours to dev-master, too, if you're up for it)

PR #693, opened 2018-12-27
Add driver for Directus

Log1x:
This adds a simple driver for [Directus](https://github.com/directus/directus), an open-source headless CMS & API.


mattstauffer:
We made a policy a while ago to not add any new drivers, with the intention of building a new system for adding drivers. That hasn't happened yet, but I'm still avoiding adding more drivers.

flayks:
Directus v8 is outdated anyway and v9 went with full NodeJS so this is not relevant anymore

mattstauffer:
Great, thanks!

Issue #533, opened 2018-03-08
In Brew.php line 187: Unable to determine linked PHP.                                    


orangeShadow:
I try methods: composer global update, install php 72, brew upgrade,  but error repeat and repeat ? what can i do, to run service ? valet version 2.0.8

drbyte:
@donaldallen wrote:
> It would be nice if Valet had some suggestions on what to do when "Unable to determine linked PHP." comes up.

Two code references:
https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L178-L179
In this one it's looking to ensure that the `php` binary is linked. So, if you've unlinked all the installed PHP versions, then you'll get that error message.

Next it moves on to verify that the linked php binary is a supported one, per this list:
https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L53-L56

uruloke:
Getting some error here, tried to follow my symlink and it looks like this `../Cellar/php/7.2.2_14/bin/php`. 

[Looking at how it matches,](https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L185) it looks for a full path containing one of the supported versions.

The supported php version looks for strings named like this `php72`, which of course is not in the path.

scherii:
@uruloke Should this not be caught [by the first item](https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L55) `php`?

orangeShadow:
uruloke, Yes, I commented locally this checking, and all is working, but it is not good decision

uruloke:
@scherii You are right, it should be caught by that, didn't see it. Gotta test this out a little locally and see if I can find an great solution.

uruloke:
After cloning the repo I can't reproduce the error, lol. 
If OP or someone else could tell how to reproduce the error, I will gladly try to fix it.

orangeShadow:
Try brew update php veriosn

uruloke:
@orangeShadow Just uninstalled php72 then installed php71, linked to php71, upgraded to php72 and linked to php72, then ran valet start and everything seems to work.

orangeShadow:
@uruloke, Can you write your file path and name for php link

lesaff:
Brew has promoted `php7.2` to core, so it is called `php` now. 
https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L185 will always throw exception until the logic is revised.

martijnvdven:
@scherii,

[...]

PR #699, opened 2019-01-08
Change default PHP install pkg to 'php'

drbyte:
When no homebrew-installed PHP version is found, Valet attempts to install one.
The old default was `php71`. This PR changes it to `php`, which will install the latest Homebrew PHP version.
The problem with installing an older version is that Homebrew no longer symlinks it automatically. Rewriting Valet to do the symlinking is a little more involved since we should check for "which" version to link, vs other versions installed, etc.
Simply changing the default to the primary supported PHP version allows Valet to rely on Homebrew symlinking it properly ... ie: currently that means it'll install PHP 7.3

This should resolve recent complaints in old issue #533 related to "new installs of Valet on a fresh OSX"

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -37,7 +37,7 @@ class PhpFpm
     function install()
     {
         if (! $this->brew->hasInstalledPhp()) {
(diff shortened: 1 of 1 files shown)


mattstauffer:
Gonna merge, test, and release as soon as Travis builds.

PR #700, opened 2019-01-08
Rework PHP version resolution

drbyte:
Last year's Homebrew's PHP packaging changes altered their version numbering strategy.
Now that their changes appear to have stabilized, Valet also needs some updates to match.

The `linkedPhp()` function was parsing the symlinked directory name for where the php binaries are stored, but that numbering strategy has morphed over time.

This PR changes the logic to accommodate the most common directory naming strategies I can find, including those of older installs.
I've included some examples of these names in code comments for future reference since finding a variety of them can be complicated.


tests/BrewTest.php
```diff
@@ -174,9 +174,21 @@ public function test_linked_php_returns_linked_php_formula_name()
...
+        $files->shouldReceive('isLink')->once()->with('/usr/local/bin/php')->andReturn(true);
+        $files->shouldReceive('readLink')->once()->with('/usr/local/bin/php')->andReturn('/test/path/php/7.2.9_2/test');
+        swap(Filesystem::class, $files);
+        $this->assertSame('php@7.2', resolve(Brew::class)->linkedPhp());
+
+        $files = Mockery::mock(Filesystem::class);
+        $files->shouldReceive('isLink')->once()->with('/usr/local/bin/php')->andReturn(true);
+        $files->shouldReceive('readLink')->once()->with('/usr/local/bin/php')->andReturn('/test/path/php72/7.2.9_2/test');
+        swap(Filesystem::class, $files);
+        $this->assertSame('php@7.2', resolve(Brew::class)->linkedPhp());
```
mattstauffer: SHould we also add one for "../Cellar/php@7.2/7.2.13/bin/php" per your note above? Or am I not actually correctly following this?

Thanks!

drbyte: Good catch. Will do.

drbyte: Pushed.

PR #701, opened 2019-01-11
getPrefix() added to retrieve user's path , replacement of  hard coded /usr/local path setting

xajayanthi:


drbyte:
Hi,
Can you explain what problem this is trying to solve?  ie: what system configuration makes this necessary?

xajayanthi:
> Hi,
> Can you explain what problem this is trying to solve? ie: what system configuration makes this necessary?

Hi,

 The changes have been done to handle the error ,when homebrew is installed other than the default path /usr/local . As homebrew was installed in /usr/****/local, the error message”Unable to determine linked PHP.” was received.

To resolve this error,the following changes have been done

1.Added a command line execution “ $this->prefix = trim($this->cli->runAsUser('brew --prefix’));” in the Brew class constructor function
2.Function getPrefix is written to return the prefix .
3.getPrefix function is called in the places where path configuration is to be done.

However, A further change will be done to move the Command line execution “ $this->prefix = trim($this->cli->runAsUser('brew --prefix’))  to the Function getPrefix()

public function getPrefix()
{
  return $this->prefix = trim($this->cli->runAsUser('brew --prefix’);
  
}

Now,PHP Unit testing is being carried out .By mistake,a pull request was created before complete preparation is done.Sorry if there is any inconvenience caused.Please suggest us what needs to be done.Thank you!

xajayanthi:
> @xajayanthi wrote:
> 
> > when homebrew is installed other than the default path /usr/local . As homebrew was installed in /usr/****/local
> 
> Okay. Can you explain "why" you did that?
> (I'd like to understand your reasons.)

As Multiple user is using the same system, environment is being created for each user. As explained in brew documentation ,Homebrew multiple installation method is being used.Hence Homebrew is installed for every user and Installation path is selected according to each user set up.

xajayanthi:
> Hmmmm ... There are mixed opinions of Homebrew's implementation of multi-user installation and functionality.
> But then there's also the challenge of serving valet nginx configs, which are locked/restricted from other users due to permission masks.
> Further, your approach (as per your PR) suggests installing multiple copies of PHP and nginx and dnsmasq. But Valet expects PHP and nginx and dnsmasq to run only one instance, as root. How have you sorted out having the right instance running for the current user, when you login as someone else?


In our system,Multi-users using the same machine but in different times with the different set of packages .When it comes to Valet, We understand that there might be challenges having multiple nginx,dnsmasq installations.

Also,We understand that there is a possibility to have homebrew installation other than the default path /usr/local.And there might be users who installed homebrew in different location other than /usr/local for various reasons.In these cases, will it not be helpful having prefix option  instead of having hard-coded(/usr/local) path value ?As it has resolved the PHP linking issue(Unable to determine PHP) in our case, Will there be other challenges incorporating this change? If there is difficulty in incorporating this change ,we will cancel our merge request.would like to hear your recommendation. Thank you!

drbyte:
1. I'm not asking you to cancel. I'm seeking to understand. I haven't had an opportunity to test this.
I suspect 95% of Valet users will not need this.
From the perspective of code maintenance I wish this didn't mean losing the visual recognition points in the code. For example, where `'/usr/local/etc/php'` is really easy to recognize and spot, the new isn't so visually friendly: `$this->brew->getPrefix().'/etc/php'`

2. A quick scan of your PR reveals a couple more places where you didn't change `/usr/local` to insert the brew prefix dir. Was this intentional?
 - createSudoersEntry()
 - DnsMasq class:  var $configPath = '/usr/local/etc/dnsmasq.conf';
 - were there any other occurrences that you chose not to change?

3. I like that you honored Homebrew's own nomenclature in your method naming: they call `/usr/local` a "prefix", as did you. I just wish they had chosen something more accurate such as "installation directory", rather than "prefix". Still, I agree with your choice there.

xajayanthi:
> 1. I'm not asking you to cancel. I'm seeking to understand. I haven't had an opportunity to test this.
>    I suspect 95% of Valet users will not need this.
>    From the perspective of code maintenance I wish this didn't mean losing the visual recognition points in the code. For example, where `'/usr/local/etc/php'` is really easy to recognize and spot, the new isn't so visually friendly: `$this->brew->getPrefix().'/etc/php'`
> 2. A quick scan of your PR reveals a couple more places where you didn't change `/usr/local` to insert the brew prefix dir. Was this intentional?
> 
> * createSudoersEntry()
> * DnsMasq class:  var $configPath = '/usr/local/etc/dnsmasq.conf';
> * were there any other occurrences that you chose not to change?
> 
> 1. I like that you honored Homebrew's own nomenclature in your method naming: they call `/usr/local` a "prefix", as did you. I just wish they had chosen something more accurate such as "installation directory", rather than "prefix". Still, I agree with your choice there.

Thank you for understanding and support!

createSudoersEntry() was untouched as not to disturb the valet installation as defined in package.

Would like to understand that 
1.Is it recommended  to get valet installation done  in /usr/local/bin/valet ?  

2.Can it be supported if valet gets installed in other location as homebrew's location ( considering the case when homebrew is  installed other than default location) ?

It  makes us to think further that our change may not serve  the purpose as we thought,if valet is recommended to install in usr/local/bin/valet .

drbyte:
You asked about relocating `/usr/local/bin/valet`. There are 2 things involved with changing this. First, it must be in a directory on the User's PATH. Second, Valet must know where it is in order to update it. If it's not in the location that Valet expects, then Valet can't make sure it's symlinked properly. So, if you change its location, you'll need maintain it yourself.

The more I look at this, the more I think this is the wrong direction to take for solving the "multi-user" problem, at least as far as trying to fix it in Valet. I think as far as the Valet project goes, solving the multi-user issue shouldn't involve installing multiple copies of Homebrew apps and configs. Thus I think this PR is not suitable.

You are of course welcome to customize the code to suit your needs.
Thanks for exploring the bigger picture with me, with my many questions!

xajayanthi:
> You asked about relocating `/usr/local/bin/valet`. There are 2 things involved with changing this. First, it must be in a directory on the User's PATH. Second, Valet must know where it is in order to update it. If it's not in the location that Valet expects, then Valet can't make sure it's symlinked properly. So, if you change its location, you'll need maintain it yourself.
> 
> The more I look at this, the more I think this is the wrong direction to take for solving the "multi-user" problem, at least as far as trying to fix it in Valet. I think as far as the Valet project goes, solving the multi-user issue shouldn't involve installing multiple copies of Homebrew apps and configs. Thus I think this PR is not suitable.
> 
> You are of course welcome to customize the code to suit your needs.
> Thanks for exploring the bigger picture with me, with my many questions!

Thank you very much for your time and support ! We are closing this PR .

xajayanthi:
We are closing this PullRequest now and will explore the feasibility.Thank you very much for your time and support !

PR #705, opened 2019-01-17
Add JSON:API support for Drupal

samgranger:
This change fixes issues with using the JSON:API in Drupal.

### DIFF
--- a/cli/drivers/DrupalValetDriver.php
+++ b/cli/drivers/DrupalValetDriver.php
@@ -57,7 +57,7 @@ class DrupalValetDriver extends ValetDriver
     {
         $sitePath = $this->addSubdirectory($sitePath);
 
-        if (!isset($_GET['q']) && !empty($uri) && $uri !== '/') {
+        if (!isset($_GET['q']) && !empty($uri) && $uri !== '/' && strpos($uri, '/jsonapi/') === false) {
           $_GET['q'] = $uri;
         }
 


mattstauffer:
@samgranger Ping--anyone you know who can test it who's a Drupal dev using Valet?

cli/drivers/DrupalValetDriver.php
```diff
@@ -57,7 +57,7 @@ public function frontControllerPath($sitePath, $siteName, $uri)
     {
         $sitePath = $this->addSubdirectory($sitePath);
 
-        if (!isset($_GET['q']) && !empty($uri) && $uri !== '/') {
+        if (!isset($_GET['q']) && !empty($uri) && $uri !== '/' && strpos($uri, '/jsonapi/') === false) {
```
e0ipso: 👋 I'm the JSON:API maintainer in Drupal.

Bear in mind that the `/jsonapi` path is configurable. So is anyone changes the defaults this may break.

chrfritsch:
This fix only works for the contrib module. There is still https://github.com/laravel/valet/issues/771

cli/drivers/DrupalValetDriver.php
arshad: @e0ipso oh yeah. I completely missed that. Funny thing is, IIRC, I wrote the patch to make this configurable. 😄

mattstauffer: Thanks for sharing, y'all! Sooo.. is there any way to fix this driver or is that just not going to be possible since it's totally variable?

mattstauffer:
@chrfritsch Thanks for bringing that back up! I'd love for anyone who knows how this all works to PR in a fuller fix.

cli/drivers/DrupalValetDriver.php
vijaycs85: I have a custom prefix and good to have this issue to know where should I look for in valet. thanks!

Issue #709, opened 2019-01-20
If running latest version of php formula (current: php@7.3) it will not correctly start the service


jrbarnard:
### To recreate

brew install php@7.3
brew unlink {current php version or php}
brew link php@7.3

If you call valet start you will then only see:
```
Restarting nginx...
Valet services have been started.
```

The reason is because when calling Brew::restartLinkedPhp() it passes the result of Brew::linkedPhp() to Brew::restartService(), the call the Brew::installed() then returns false, when you are running the latest version of php though this comes back as php@7.3 rather than just php.

### Suggested fix

Use the constant: Brew::LATEST_PHP_VERSION to change the php service to start
```
// In Valet/Brew.php
function restartLinkedPhp()
{
    $version = $this->linkedPhp();
    $this->restartService($version === self::LATEST_PHP_VERSION ? 'php' : $version);
}
```
I'll put in a PR for this

jrbarnard:
I created a PR however after some discussion on #710 this wouldn't cover people with older versions of php who installed them when it was the latest on homebrew.

In order to update the PR and re open it relies on some code from #710 

Ongoing branch for this fix: https://github.com/jrbarnard/valet/tree/bugfix/php-latest-%23709

jrbarnard:
Created PR here: #720

Issue #708, opened 2019-01-20
When running valet stop it runs stop on all installed php versions, this can be slow


jrbarnard:
E.g if you have php@5.6, php@7.0, php@7.1, php@7.2 all installed it will call stop on all even though say on php@7.2 is running. 

We could instead look at the currently running services before deciding what to pass to Brew::stopService().

As part of the work for #378 I've got code that would allow us to do this quite easily, so we could potentially swap the PhpFpm::stop() with PhpFpm::stopRunning();

PR for this new functionality: #710 

I'm happy to link this with the PR #378 and then update the PhpFpm::stop calls to stopRunning in a new PR if the other one gets approved.

drbyte:
#710 has been merged.

jrbarnard:
@drbyte  @mattstauffer  The underlying issue still exists, the PR that was merged included code that can be used to fix the issue but didn't fix the issue.

Here's a screen shot showing the issue:
<img width="545" alt="Screen Shot 2019-09-27 at 08 30 27" src="https://user-images.githubusercontent.com/6758366/65750958-770bf500-e101-11e9-82c7-f5585cbd3fe0.png">

I'll put a PR together to update this to look for the currently running version of PHP before stopping?

PR #378, opened 2017-05-15
PHP version switch

timneutkens:
Built upon the work in #313.

`valet use 7.1`
`valet use 7.0`
`valet use 5.6`

If not installed it will install the version of PHP if it's in the supported versions list 👌

PR #710, opened 2019-01-20
switch php version 

jrbarnard:
A currently working version of #378 which also adds some functionality that would support #708 .

Usage:
```
valet use php@7.2
valet use php7.2 // We will convert to php@7.2
valet use php72 // We will convert to php@7.2
```
All passed php versions are checked against the supported versions in the Brew class.

Couple of things this has brought up, there is a current bug which I've documented here: #709 (i'll put a PR in for it)
Valet doesn't want to work on my machine when I've linked php@5.6, I'm running high sierra and I think this is just a compatibility bug rather than with this feature, I'd value someone else having a test though.

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -137,6 +137,7 @@ class PhpFpm
      * Use a specific version of php
      *
      * @param $version
+     * @return string
      */
     function useVersion($version)
     {
@@ -147,17 +148,17 @@ class PhpFpm
 
         info(sprintf('Finding brew formula for: %s', $version));
         $foundVersion = $this->brew->search($version)
-            ->filter(function ($service) {
+            ->first(function ($service) {
                 return $this->brew->supportedPhpVersions()->contains($service);
-            })
-            ->first();
-        info(sprintf('Found brew formula: %s', $foundVersion));
+            });
 
-        if (!$this->brew->supportedPhpVersions()->contains($foundVersion)) {
+        if (is_null($foundVersion)) {
             throw new DomainException(
-                sprintf('Valet doesn\'t support PHP version: %s', $foundVersion)
+                sprintf('Valet can\'t find a supported version of PHP for: %s', $version)
             );
         }
+
+        info(sprintf('Found brew formula: %s', $foundVersion));
 
         if ($this->brew->hasLinkedPhp()) {
             $currentVersion = $this->brew->linkedPhp();
@@ -171,5 +172,7 @@ class PhpFpm
         $this->brew->link($foundVersion, true);
 
         $this->install();
+
+        return $foundVersion;
     }
 }
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -121,6 +121,16 @@ class PhpFpm
             : "/usr/local/etc/php/${versionNormalized}/php-fpm.d/www.conf";
     }
 
+    /**
+     * Only stop running php services
+     */
+    function stopRunning()
+    {
+        $this->brew->stopService(
+            $this->brew->getRunningServices('php')->all()
+        );
+    }
+
     function useVersion($version)
     {
         // Ensure we have php{version}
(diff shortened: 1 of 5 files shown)


mattstauffer:
I did a quick skim and it looks good, but I look forward to review from the folks who know more than me. For now I'll download and test it out--thanks for your work on this!

mattstauffer:
OK, I was running php 7.2 and this is what I saw:

image

I'll look into it more, gotta run take care of kids.

Here's my current installed list:

image

My guess is it hsa something to do with it being `php` instead of `php@7.2` but this whole new Homebrew versioning system is very confusing to me.

jrbarnard:
@mattstauffer That is confusing, I presume your local has had php@7.2 linked to php for a while, I'll have a look at a workaround that works for when php is linked but is not the latest (e.g php@7.3).

mattstauffer:
That was it. Since I was using `php`, not `php@7.2`, everything broke. So we need to allow for that. I installed `php@7.2` and then swapping back and forth worked fine:

image

mattstauffer:
@jrbarnard Yep, my guess is if someone runs `brew install php` while it's at 7.1, that still stays linked at 7.1 even when the latest becomes 7.2 or 7.3. So we need to figure out to run our uninstall based on which one is linked, whether that's `php` or `php@7.1` or `php@7.2` or whatever, rather than assuming we can construct `php@currentlyInstalledVersion`.

That's my guess.

jrbarnard:
@mattstauffer I've updated it to take from the linked path the php directory, e.g if latest (at any time) it should be under php, so we brew unlink php as opposed to brew unlink php@7.2.

It's using the regex from Brew::linkedPhp() so should handle the older formats too, I've updated the tests and added new ones to cover the refactor.

cli/Valet/PhpFpm.php
```diff
@@ -120,4 +118,53 @@ function fpmConfigPath()
...
+     * @return string
+     */
+    function useVersion($version)
+    {
+        // If passed php7.2 or php72 formats, convert to php@7.2 format:
+        $version = preg_replace('/(php)([0-9+])(?:.)?([0-9+])/i', '$1@$2.$3', $version);
+
+        if (!$this->brew->supportedPhpVersions()->contains($version)) {
+            throw new DomainException(
+                sprintf('Valet doesnn\'t support PHP version: %s', $version)
```
mattstauffer: Typo in "doesn't"
but also maybe we should suggest one?

```php
sprintf("Valet doesn't support PHP versions: %s (try something like 'php71' instead)"
```

And that does also make me curious whether we should consider using "7.1" instead of "php7.1" or whatever as our syntax? `valet use 7.1`

mattstauffer:
image

I'm surprised that Homebrew makes this difficult, but I think that is a Homebrew thing, not a Valet thing. However, I still want to figure this out.

mattstauffer:
https://www.reddit.com/r/PHP/comments/a4vggg/php_70_removed_from_homebrew/

Hrm. I bet we're going to get some requests for that one. But I think that can be a matter of another day.

jrbarnard:
Hi @mattstauffer Good spot with the typo I'll update.

Syntax wise originally I set it up so you could leave off the php prefix however put it back on as 'valet use' is quite generic so wanted to be more explicit in what you're asking valet to use.

I can add a suggested PHP version, maybe if we stick with latest stable and stick to the newer 7.2 syntax? Technically use will convert the old to new anyway but probably still worth using the new.

jrbarnard:
Hi @mattstauffer Good spot with the typo I'll update.

Syntax wise originally I set it up so you could leave off the php prefix however put it back on as 'valet use' is quite generic so wanted to be more explicit in what you're asking valet to use.

I can add a suggested PHP version, maybe if we stick with latest stable and stick to the newer 7.2 syntax? Technically use will convert the old to new anyway but probably still worth using the new.

mattstauffer:
image

Looks better! Maybe if we drop the colon it won't line wrap?

Also... the fact that we're building this with a syntax that could potentially be used with not-PHP makes some inconsistencies clear.

```
valet use mysql
```

```
valet use php71
```

That makes sense, for sure. Although in that context, our error would be "`7.1` isn't on our list of installable packages. Here's the full list: []"

The bigger question is, are we *going to* allow for supporting other packages? My guess is "no" but I can't say for sure. Let's say we weren't. Is there a better syntax? I agree that `valet use 7.1` is too vague. Trying to think of a verb other than `use` that communicates "install PHP version" but I have nothing yet.

So.. I don't know exactly. Gonna be thinking on this a bit and I welcome feedback from others

jrbarnard:
The only thing I can think is we set it up like:
```
valet use php 7.1
```

It's relatively clear then that you want to use php, version 7.1, it may be a little verbose though?
```
valet use invalid
// Valet doesn't currently support that package, we currently support ['php']

valet use php 9.1
// Valet doesn't current support that version of php, why don't you try 'valet use php 7.2'
```

mattstauffer:
@jrbarnard good idea. let's see if we can get folks to weigh in

jalendport:
Meant to comment on this back when Matt asked for some folks to weigh-in. Love the proposed changes and I agree that `use` is the best option here 👌 I've actually got a little bash script that handles the php switching for you ([`switch-php`](https://github.com/jalendport/switch-php)), but happy that this is finally making it into core. Thank you for the work on this @jrbarnard! 🙌

PR #726, opened 2019-02-03
Remove outdated option in brew install options

### LINKED ISSUE #725: Homebrew fails to install nginx because of removal of with http2 option
Homebrew has recently been working to strip options from formulas. Thus Valet fails to install nginx because it specifies `--with-http2`, which is no longer an option.

See homebrew/homebrew-core#31510 😢

zchrykng:
Resolves #725 

Homebrew has been removing options on all formulas, therefore specifying an install option breaks the install of Valet.


drbyte:
Looks good to me. 👍

PR #751, opened 2019-04-01
Change PHP keywords to comply with PSR2

friendlydan:
Hi, I've changed a PHP keyword to comply with [this standard](https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull) in PSR2. It's admittedly a small fix but I hope it helps!


mattstauffer:
Thanks @friendlydan! :)

PR #754, opened 2019-04-03
Updated loading of server environment variables

haokaiyang:
**Modify as described in the document**
Document usage : [site-specific-environment-variables](https://laravel.com/docs/5.8/valet#site-specific-environment-variables)
Actual usage : [actual usage](https://github.com/laravel/valet/commit/67546f1fc32bf453a9cd7eb6203c51372a552b26#diff-33bb37225e8574f3fc6429cd60f9e617)

### DIFF
--- a/cli/drivers/ValetDriver.php
+++ b/cli/drivers/ValetDriver.php
@@ -180,13 +180,11 @@ abstract class ValetDriver
 
     /**
      * Load server environment variables if available.
-     * Processes any '*' entries first, and then adds site-specific entries
      *
      * @param  string  $sitePath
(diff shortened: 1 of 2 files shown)


drbyte:
Why not just fix the docs instead of removing the functionality in the code?

mattstauffer:
Yep--this is a chance to update the docs.

haokaiyang:
> Why not just fix the docs instead of removing the functionality in the code?

The reason I did this is because the code import the configuration file in the root directory of the project, Configuring generic is redundant. Of course, can also change the configuration path to the configuration file of valet.

PR #755, opened 2019-04-03
TravisCI - change nightly(8.0) PHP to 7.4snapshot

drbyte:
Fixes the failing tests.

The former approach of using `nightly` to grab the latest in-dev PHP version isn't working well now that the `nightly` build is an alias for 8.0 ... and 8.0 isn't resolving composer packages since most haven't tagged anything compatible with it yet.
So switching to `7.4snapshot` allows testing against the next announced PHP version set to release later this year.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - nightly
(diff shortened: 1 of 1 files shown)


mattstauffer:
<3

PR #757, opened 2019-04-04
#548 Refinements ("valet logs" Command)

furey:
This PR is submitted in response to @mattstauffer's recent tweet regarding @gahlawat's #548:

[tweet
](https://twitter.com/stauffermatt/status/1113538473976324105)

By default, the `logs` command tails Valet's `/Log/nginx-error.log` file:

```
valet logs
```

To tail **custom** log files, add relevant `"logs"` key/path values to `~/.config/valet/config.json`…

```
{
    "tld": "test",
    "paths": [
        "…"
    ],
    "logs": {
        "nginx-access": "/path/to/nginx-access.log",
        "php-error": "/path/to/php-fpm-errors.log",
        "dnsmasq": "/path/to/dnsmasq.log",
        "etc": "…"
    }
}
```

…then pass the relevant `"logs"` key when calling the `logs` command:

```
valet logs nginx-access
valet logs php-error
valet logs dnsmasq
```

To **follow** the tailed log file, pass the `-f|--follow` flag:

```
valet logs -f
valet logs --follow
```

To **restrict** the number of tailed lines, pass the `--lines` flag (`-n` was taken):

```
valet logs --lines 1
valet logs --lines=1
```

Hope these tweaks add the refinement you're looking for.

Cheers!

🤓🤞

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -295,10 +295,8 @@ if (is_dir(VALET_HOME_PATH)) {
      */
     $app->command('log [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
         $defaultLogs = [
-            'php' => VALET_HOME_PATH.'/Log/php.log',
             'php-fpm' => '/usr/local/var/log/php-fpm.log',
             'nginx' => VALET_HOME_PATH.'/Log/nginx-error.log',
-            'mysql' => VALET_HOME_PATH.'/Log/mysql.log',
             'mailhog' => '/usr/local/var/log/mailhog.log',
             'redis' => '/usr/local/var/log/redis.log',
         ];
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -291,7 +291,7 @@ if (is_dir(VALET_HOME_PATH)) {
     ]);
 
     /**
-     * Tail log files.
+     * Tail log file.
      */
     $app->command('logs [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
         $defaultLogs = [
@@ -304,46 +304,60 @@ if (is_dir(VALET_HOME_PATH)) {
         ];
 
         $configLogs = data_get(Configuration::read(), 'logs');
-        if (!is_array($configLogs)) $configLogs = [];
+        if (! is_array($configLogs)) {
+            $configLogs = [];
+        }
 
         $logs = collect(array_merge($defaultLogs, $configLogs))->sortKeys();
 
-        if (!$key) {
+        if (! $key) {
             info(implode(PHP_EOL, [
-                'Here are the logs you might be interested in.',
-                null,
-                'In order to tail a log, pass the relevant log key (e.g. "php")',
+                'In order to tail a log, pass the relevant log key (e.g. "nginx")',
                 'along with any optional tail parameters (e.g. "-f" for follow).',
                 null,
-                'For example: "valet logs php -f --lines=3"',
+                'For example: "valet logs nginx -f --lines=3"',
+                null,
+                'Here are the logs you might be interested in.',
                 null,
             ]));
+
             table(
                 ['Keys', 'Files'],
                 collect($logs)->map(function ($file, $key) {
                     return [$key, $file];
                 })->toArray()
             );
+
             info(implode(PHP_EOL, [
                 null,
                 'Tip: Set custom logs by adding a "logs" key/file object',
                 'to your "'.Configuration::path().'" file.',
             ]));
+
             exit;
         }
 
-        if (!isset($logs[$key])) return warning('No logs found for ['.$key.'].');
+        if (! isset($logs[$key])) {
+            return warning('No logs found for ['.$key.'].');
+        }
 
         $file = $logs[$key];
-        if (!file_exists($file)) return warning('Log path ['.$file.'] does not (yet) exist.');
+        if (! file_exists($file)) {
+            return warning('Log path ['.$file.'] does not (yet) exist.');
+        }
 
         $options = [];
-        if ($follow) $options[] = '-f';
-        if ((int) $lines) $options[] = '-n '.(int) $lines;
+        if ($follow) {
+            $options[] = '-f';
+        }
+        if ((int) $lines) {
+            $options[] = '-n '.(int) $lines;
+        }
 
         $command = implode(' ', array_merge(['tail'], $options, [$file]));
+
         passthru($command);
-    })->descriptions('Tail log files');
+    })->descriptions('Tail log file');
 }
 
 /**
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -289,6 +289,28 @@ if (is_dir(VALET_HOME_PATH)) {
     })->descriptions('Change the version of php used by valet', [
         'phpVersion' => 'The PHP version you want to use, e.g php@7.2',
     ]);
+
+    /**
+     * Tail log file.
+     */
+    $app->command('logs [-f|--follow] [--lines=] [file]', function ($follow, $lines, $file = null) {
+        $defaultLogFile = VALET_HOME_PATH.'/Log/nginx-error.log';
+
+        $logFile = data_get(Configuration::read(), "logs.$file", $defaultLogFile);
+
+        $options = [];
+        
+        if ($follow) $options[] = '-f';
+
+        if ((int) $lines) {
+            $options[] = '-n';
+            $options[] = (int) $lines;
+        }
+
+        $logCommand = implode(' ', array_merge(['tail'], $options, [$logFile]));
+
+        passthru($logCommand);
+    })->descriptions('Tail log file');
 }
 
 /**
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -293,7 +293,7 @@ if (is_dir(VALET_HOME_PATH)) {
     /**
      * Tail log file.
      */
-    $app->command('logs [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
+    $app->command('log [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
         $defaultLogs = [
             'php' => VALET_HOME_PATH.'/Log/php.log',
             'php-fpm' => '/usr/local/var/log/php-fpm.log',
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -291,26 +291,59 @@ if (is_dir(VALET_HOME_PATH)) {
     ]);
 
     /**
-     * Tail log file.
+     * Tail log files.
      */
-    $app->command('logs [-f|--follow] [--lines=] [file]', function ($follow, $lines, $file = null) {
-        $defaultLogFile = VALET_HOME_PATH.'/Log/nginx-error.log';
+    $app->command('logs [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
+        $defaultLogs = [
+            'php' => VALET_HOME_PATH.'/Log/php.log',
+            'php-fpm' => '/usr/local/var/log/php-fpm.log',
+            'nginx' => VALET_HOME_PATH.'/Log/nginx-error.log',
+            'mysql' => VALET_HOME_PATH.'/Log/mysql.log',
+            'mailhog' => '/usr/local/var/log/mailhog.log',
+            'redis' => '/usr/local/var/log/redis.log',
+        ];
 
-        $logFile = data_get(Configuration::read(), "logs.$file", $defaultLogFile);
+        $configLogs = data_get(Configuration::read(), 'logs');
+        if (!is_array($configLogs)) $configLogs = [];
 
-        $options = [];
-        
-        if ($follow) $options[] = '-f';
+        $logs = collect(array_merge($defaultLogs, $configLogs))->sortKeys();
 
-        if ((int) $lines) {
-            $options[] = '-n';
-            $options[] = (int) $lines;
+        if (!$key) {
+            info(implode(PHP_EOL, [
+                'Here are the logs you might be interested in.',
+                null,
+                'In order to tail a log, pass the relevant log key (e.g. "php")',
+                'along with any optional tail parameters (e.g. "-f" for follow).',
+                null,
+                'For example: "valet logs php -f --lines=3"',
+                null,
+            ]));
+            table(
+                ['Keys', 'Files'],
+                collect($logs)->map(function ($file, $key) {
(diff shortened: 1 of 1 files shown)


drbyte:
👍 
This works. Although I wonder if it should be called `valet tail` instead of `valet logs`.

So does the code by @michielgerritsen mentioned on twitter.
Both have varying approaches.
Perhaps eventually combining their features may be useful.

This is fine to merge if you're satisfied with the features and the command name.

(You can easily test by copy/pasting the new code into a new file in your `~/.config/valet/Extensions/` directory.)

mattstauffer:
Thanks for working on this!!

I am unsure about the best name.

Here's what I would think of for the API:

`valet logs` shows you the output of all the logs

`valet logs key` shows you the output of just one log, based on the provided key

and then we pre-define all the logs that are expected.... that's one of the big values, that people don't have to learn where things are stored.

I'll go look in a second at the code by @michielgerritsen that @drbyte mentioned.

mattstauffer:
OK, just looked at the other PR. It works a bit like I wanted, with the simple keys and pre-defined lists.

I think the value of the structure of this PR is that I can either follow it OR view it, which is really nice.

What if we keep this PR's structure, name it `tail` so people are less surprised when they can pass `tail`-style config options, and then add the pre-defined keys?

drbyte:
@mattstauffer wrote:
> Here's what I would think of for the API:
> 
> `valet logs` shows you the output of all the logs

I'd be a bit cautious of this, since that's a ton of output, and would probably be meaningless. Maybe better to just tail the last 5 lines from "all"?
Maybe for naming, let the `logs` command just list the various log files and their locations -- let it be info-only, and then tail could do the actual display.

mattstauffer:
@drbyte mhm, that's good. I was suggesting that back when I was assuming this did "follow" by default.

Hm.
I sort of think there might be a clever way to make this not two commands.

mattstauffer:
OK, just thinking out loud here:

```bash
$ valet logs

Here are the logs you might be interested in. In order to tail any of these logs, pass the key (e.g. "php-error") to this command, along with any tail parameters like -f for follow. For example, "valet logs php-error -f"

[nginx-access]: "/path/to/nginx-access.log"
[php-error]: "/path/to/php-fpm-errors.log"
[etc.]
```

Maybe?

drbyte:
Ya, that's what I was thinking too when I was comparing the two approaches in the 2 PRs.

furey:
Hi @mattstauffer / @drbyte,

Thanks for the feedback!

I've just pushed an update to the PR that attempts to handle the points you raised.

In summary via screenshots…

Without log key:

valet-logs

With non-existent log key:

valet-logs-invalid-key

With non-existent log path:

valet-logs-mailhog

With valid log key:

valet-logs-redis

With valid log key and optional `tail` flags (added `-l` shortcut for `--lines`; again, `-n` was taken):

valet-logs-redis-flags

The guts of the command should probably be refactored into a `Logs` class, but I thought it best to leave everything at the command level until the functionality is locked down (then refactor if required).

> OK, just looked at the other PR […] with the simple keys and pre-defined lists.

Regarding that pre-defined list, as you can see I've simply dropped (and tweaked slightly) the values from the [weprovide/valet-plus](https://github.com/weprovide/valet-plus/pull/261/files) PR. What key/values should `valet logs` have defined by default? I might be missing something, but I can only find reference to `[VALET_HOME_PATH]/Log/nginx-error.log` in the repository source code.

Really appreciate your thoughts on the updated behaviour (whenever you have the time).

Cheers!

drbyte:
> RE: Travis CI, the [`7.4snapshot` failure](https://travis-ci.org/laravel/valet/jobs/517623181) contains multiple occurrences of the following error:
> ```
> php: error while loading shared libraries: libonig.so.2: cannot open shared object file: No such file or directory
> ```

Argh. The joys of moving-targets. Grrrr. That's just a TravisCI thing. Not related to your code.

drbyte:
This PR looks good. 
Tested it here and it works as promised.

A future update may include specifically configuring (creating) more (matching) logs as part of the core valet install.

mattstauffer:
Thanks for the update here @furey and thanks for reviewing and testing @drbyte.

Testing it locally now; if it works well i'lll review the PR for a good old refactor and then we'll launch!

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+        ];
+
+        $configLogs = data_get(Configuration::read(), 'logs');
+        if (!is_array($configLogs)) $configLogs = [];
+
+        $logs = collect(array_merge($defaultLogs, $configLogs))->sortKeys();
+
+        if (!$key) {
+            info(implode(PHP_EOL, [
+                'Here are the logs you might be interested in.',
```
mattstauffer: Can we move the "Here are the logs" line to just above the table? So make it the last line of text instead of the first?

cli/valet.php
```diff
@@ -289,6 +289,61 @@
     })->descriptions('Change the version of php used by valet', [
         'phpVersion' => 'The PHP version you want to use, e.g php@7.2',
     ]);
+
+    /**
+     * Tail log files.
+     */
+    $app->command('logs [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
+        $defaultLogs = [
+            'php' => VALET_HOME_PATH.'/Log/php.log',
```
mattstauffer: `php` and `mysql` don't exist on mine; i'll take a look at figuring out why we're expecting this to exist and then either have us drop this line or add instructions.

cli/valet.php
```diff
@@ -289,6 +289,61 @@
     })->descriptions('Change the version of php used by valet', [
         'phpVersion' => 'The PHP version you want to use, e.g php@7.2',
     ]);
+
+    /**
+     * Tail log files.
+     */
+    $app->command('logs [-f|--follow] [-l|--lines=] [key]', function ($follow, $lines, $key = null) {
```
mattstauffer: I know I originally said "logs", but what would you think if this said "log" instead of "logs"? I'm not 100% sure but definitely considering `log`

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+            'php' => VALET_HOME_PATH.'/Log/php.log',
+            'php-fpm' => '/usr/local/var/log/php-fpm.log',
+            'nginx' => VALET_HOME_PATH.'/Log/nginx-error.log',
+            'mysql' => VALET_HOME_PATH.'/Log/mysql.log',
+            'mailhog' => '/usr/local/var/log/mailhog.log',
+            'redis' => '/usr/local/var/log/redis.log',
+        ];
+
+        $configLogs = data_get(Configuration::read(), 'logs');
+        if (!is_array($configLogs)) $configLogs = [];
```
mattstauffer: I think we're using PSR-2, so this would need to be in braces.

```suggestion
        if (! is_array($configLogs)) {
            $configLogs = [];
        }
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+            info(implode(PHP_EOL, [
+                'Here are the logs you might be interested in.',
+                null,
+                'In order to tail a log, pass the relevant log key (e.g. "php")',
+                'along with any optional tail parameters (e.g. "-f" for follow).',
+                null,
+                'For example: "valet logs php -f --lines=3"',
+                null,
+            ]));
+            table(
```
mattstauffer: New line before `table(` please just for ease of reading; same for before the `info` in line 327

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+                ['Keys', 'Files'],
+                collect($logs)->map(function ($file, $key) {
+                    return [$key, $file];
+                })->toArray()
+            );
+            info(implode(PHP_EOL, [
+                null,
+                'Tip: Set custom logs by adding a "logs" key/file object',
+                'to your "'.Configuration::path().'" file.',
+            ]));
```
mattstauffer: newline before `exit` plz

cli/valet.php
mattstauffer: Looks like the codebase uses spaces after bangs, so would you mind adding them to all of these?

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+            'mailhog' => '/usr/local/var/log/mailhog.log',
+            'redis' => '/usr/local/var/log/redis.log',
+        ];
+
+        $configLogs = data_get(Configuration::read(), 'logs');
+        if (!is_array($configLogs)) $configLogs = [];
+
+        $logs = collect(array_merge($defaultLogs, $configLogs))->sortKeys();
+
+        if (!$key) {
```
mattstauffer:
```suggestion
        if (! $key) {
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+            );
+            info(implode(PHP_EOL, [
+                null,
+                'Tip: Set custom logs by adding a "logs" key/file object',
+                'to your "'.Configuration::path().'" file.',
+            ]));
+            exit;
+        }
+
+        if (!isset($logs[$key])) return warning('No logs found for ['.$key.'].');
```
mattstauffer:
```suggestion
        if (! isset($logs[$key])) {
            return warning('No logs found for ['.$key.'].');
        }
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+
+        $file = $logs[$key];
+        if (!file_exists($file)) return warning('Log path ['.$file.'] does not (yet) exist.');
```
mattstauffer:
```suggestion
        if (! file_exists($file)) {
            return warning('Log path ['.$file.'] does not (yet) exist.');
        }
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+
+        $options = [];
+        if ($follow) $options[] = '-f';
```
mattstauffer:
```suggestion
        if ($follow) {
            $options[] = '-f';
        }
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+        if ((int) $lines) $options[] = '-n '.(int) $lines;
```
mattstauffer:
```suggestion
        if ((int) $lines) {
            $options[] = '-n '.(int) $lines;
        }
```

cli/valet.php
```diff
@@ -289,6 +289,61 @@
...
+
+        $command = implode(' ', array_merge(['tail'], $options, [$file]));
```
mattstauffer: newline before `passthru` plz

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Done! (cb1136e7a616b9c0231e99f95d23beb5686b0a0c)

cli/valet.php
furey: Gotcha – I'll await further instructions.

Anticipating `php` might be dropped, I've changed the example/guide text to `nginx`:

replace-php-with-nginx


Cheers!

cli/valet.php
furey: I love that Laravel/Symfony console apps can partially match command names.

If you decided to leave the definition as `logs`, you'd still be able to call it via `log`:

logs-vs-log

I'll leave the decision up to you!

🤓👍

drbyte: True. I do like that.

drbyte: I also wonder if `tail` is the better command, since that's what it's actually doing. And then we could add `logs` as an alias (like `domain` is an alias for `tld`).

But I'll defer to @mattstauffer for final direction on this.

mattstauffer: I preferred `tail` for a bit, but asked around a bit and got a pretty overwhelming consensus that, while `tail` is more accurate, `log` or `logs` are more broadly useful--folks familiar with `tail` will get it, but folks unfamiliar with `tail` will get it *much better*.

mattstauffer: OK. My preference is `log`. :) Thanks for working with me!

cli/valet.php
mattstauffer: Let's drop mysql and php for now. We can re-add them later if we figure out what their intended target is.

furey: Done! (138df82aa25073301ab962072f25b9c270101a87)

cli/valet.php
furey: Done! (12ca66ade13597c1a8bcc6f485e5a91a6e3fdc7a)

mattstauffer:
This looks good for me. @drbyte any further input?

drbyte:
I just did some testing on another machine, for sanity check. No issues.

Looks good to me. 👍

ahinkle:
@furey care to submit something to the official docs? I'll write something up tonight if you are unable to. :)

furey:
@mattstauffer / @drbyte: Thanks for the merge! 🎉

> @furey care to submit something to the official docs? I'll write something up tonight if you are unable to. :)

Hey @ahinkle! 👋

If you could tackle the docs that'd be awesome (midnight here in Australia and I'm 😴).

Cheers!

PR #760, opened 2019-04-08
Fix issue with CA serial params

ikari7789:
Should always provide CAserial, but should only provide CAcreateserial if file does not exist. Should probably fix permissions issue mentioned in the comments when attempting to create certificate.


drbyte:
👍  Works as described.

mattstauffer:
Thanks @ikari7789!

ikari7789:
Thank you too @mattstauffer!

PR #761, opened 2019-04-08
Fixes deprecated Process argument passed as string

Padrio:
Passing a command as string is deprecated since Symfony 4.2. Since we don't have command available as array, we should use `Process::fromShellCommandline()` instead.

### DIFF
--- a/cli/Valet/CommandLine.php
+++ b/cli/Valet/CommandLine.php
@@ -74,7 +74,7 @@ class CommandLine
     {
         $onError = $onError ?: function () {};
 
-        $process = new Process($command);
+        $process = Process::fromShellCommandline($command);
 
         $processOutput = '';
         $process->setTimeout(null)->run(function ($type, $line) use (&$processOutput) {


mattstauffer:
Thanks!!

drbyte:
@Padrio have you any insights about the problems people are reporting with this? 
Ref: #763 

(For me this change had no negative consequences, but it is causing trouble for others.)

mattstauffer:
My guess is that since we allow for older versions of Process, some folks have versions 2 or 3.

Padrio:
The issue is caused by allowing too old `symfony/process` versions to be accepted by Valet.
To be exact, Valet required the following versions: `~2.7|~3.0|~4.0 ` and will prefer the newest version on it's own. But when someone is installing something else globally what needs an older package version, this will force composer to downgrade `symfony/process` down to 2.7 where this feature haven't been implemented. 

We coul'd fix this by simply checking if the method exists, and use it - otherwise fallback to the legacy constructor initiation. I will submit a fix in an hour, have to leave work now and do it at the train.

PR #765, opened 2019-04-11
Allow TravisCI to fail on PHP7.4snapshot builds

drbyte:
Until it's stable we probably shouldn't be failing builds that run into glitches with the nightly snapshots.


mattstauffer:
@drbyte what would i do without you? 😍

Issue #763, opened 2019-04-11
Valet 2.2.3 install error


alexmck:
Running valet install using version 2.2.3 produces an error:

```valet install
Stopping nginx...

Fatal error: Uncaught Error: Call to undefined method Symfony\Component\Process\Process::fromShellCommandline() in /Users/alex/.composer/vendor/laravel/valet/cli/Valet/CommandLine.php:77
Stack trace:
#0 /Users/alex/.composer/vendor/laravel/valet/cli/Valet/CommandLine.php(63): Valet\CommandLine->runCommand('sudo -u "alex" ...', Object(Closure))
#1 /Users/alex/.composer/vendor/laravel/valet/cli/Valet/Brew.php(48): Valet\CommandLine->runAsUser('brew list | gre...')
#2 /Users/alex/.composer/vendor/laravel/valet/cli/Valet/Brew.php(91): Valet\Brew->installed('nginx-full')
#3 /Users/alex/.composer/vendor/laravel/valet/cli/Valet/Nginx.php(159): Valet\Brew->nginxServiceName()
#4 /Users/alex/.composer/vendor/laravel/valet/cli/includes/facades.php(28): Valet\Nginx->stop()
#5 /Users/alex/.composer/vendor/laravel/valet/cli/valet.php(49): Facade::__callStatic('stop', Array)
#6 [internal function]: Silly\Application->{closure}()
#7 /Users/alex/.composer/vendor/php-di/invoker/src/Invoker.php(82): call_user_func_array(Object(Closure), Arr in /Users/alex/.composer/vendor/laravel/valet/cli/Valet/CommandLine.php on line 77
```

Installing Laravel Valet 2.2.2 with Composer instead fixes this problem.

mattstauffer:
Ah—we just made that change in order to stop relying on a deprecated method and we may have to introduce a stricter requirement on the Symfony dependency. For now, can you run ‘composer global update’ and see if it helps at all? I’ll try to test this out a bit later.

Artnforces:
> Ah—we just made that change in order to stop relying on a deprecated method and we may have to introduce a stricter requirement on the Symfony dependency. For now, can you run ‘composer global update’ and see if it helps at all? I’ll try to test this out a bit later.

I have tried this and still have the same message.

mattstauffer:
My gut is to revert that change, since it's just deprecated for now, and then resolve the issue where we're allowing older versions of Symfony's `process` component and how to update the deprecation there.

[...]

PR #766, opened 2019-04-11
Fixed warning on newer symfony/process versions

Padrio:
And also check whether the method exists, see Issue #763 which was caused by my previous attempt to fix a warning., which can be found here: #761

### DIFF
--- a/cli/Valet/CommandLine.php
+++ b/cli/Valet/CommandLine.php
@@ -74,7 +74,11 @@ class CommandLine
     {
         $onError = $onError ?: function () {};
 
-        $process = new Process($command);
+        if (method_exists(Process::class, 'fromShellCommandline')) {
+            $process = Process::fromShellCommandline($command);
+        } else {
+            $process = new Process($command);
+        }
 
         $processOutput = '';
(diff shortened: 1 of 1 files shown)


cli/Valet/CommandLine.php
```diff
@@ -74,7 +74,11 @@ function runCommand($command, callable $onError = null)
     {
         $onError = $onError ?: function () {};
 
-        $process = new Process($command);
+        if (method_exists(Process::class, 'fromShellCommandline')) {
```
drbyte: Much as I try to avoid comments by letting the code speak for itself, this is probably one of those cases where we should add a code comment saying why we're doing this, particularly since at some point we'll want to drop the deprecated code altogether, and having the right Symfony version number documented will be helpful.

drbyte:
👍 Tests fine for me
Tested with both symfony/process 3.4.24 and 4.2.5 installed.

cli/Valet/CommandLine.php
mattstauffer: Agreed!

Padrio:
Now it should be fine.

mattstauffer:
Thanks @Padrio!

PR #780, opened 2019-05-14
Fix "sudo: sorry, you are not allowed to preserve the environment" msg

drbyte:
If you've previously run `valet trust` to allow valet to run without specifying `sudo` repeatedly or entering your password with various valet commands, recent updates to MacOS may give you a `sudo: sorry, you are not allowed to preserve the environment` response when trying to run those `valet` commands.

The fix is in updating the sudoers entry that `valet trust` creates.

This PR tells valet how to update the sudoers entry so that this message doesn't continue.

YOU WILL NEED TO RE-RUN `valet trust` IF YOU HAD PREVIOUSLY RUN IT, after installing this update, in order for this code change to have any effect.

(Technical explanation: the sudoers protections have become stricter, so we have to be more explicit that we do indeed want environment variables to flow through to the sudo user's environment when using valet via sudoers entries.)

Updates #770 and related to #321

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -272,7 +272,7 @@ class Brew
         $this->files->ensureDirExists('/etc/sudoers.d');
 
         $this->files->put('/etc/sudoers.d/brew', 'Cmnd_Alias BREW = /usr/local/bin/brew *
-%admin ALL=(root) NOPASSWD: BREW'.PHP_EOL);
+%admin ALL=(root) NOPASSWD:SETENV: BREW'.PHP_EOL);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mattstauffer:
<3

xtrasmal:
Thanks!

driesvints:
How do I run `valet trust` when I get the error message above? I can't run valet at all atm. (using the latest release).

drbyte:
> How do I run `valet trust` when I get the error message above? I can't run valet at all atm. (using the latest release).

Good point -- if Valet won't run at all because of this error, then you can manually delete the `/etc/sudoers.d/valet` file, which is the same as running `valet untrust`, and then valet will start requiring your password when you use it. And then if you're using the latest release running `valet trust` will put the file back with the correct contents. 
Or you could edit the two files shown in this PR, manually. You'll have to do it as sudo.

driesvints:
@drbyte hah, I actually just came here again to tell you that it worked by deleting the `/etc/sudoers.d/valet` file :)

Thanks!

PR #787, opened 2019-06-13
remove unnecessary confusion

CaddyDz:
I know this won't be accepted but am putting this out there if you people haven't noticed!!  
When developing Laravel apps with Valet on the localhost, all static files (especially images) are accessible without the `storage` word in the URL due to this function but once you deploy to production the URL isn't accessible anymore even after running the `storage:link` command  
Lots of times am writing code logic based on the accessibility of the files without the `storage` in URL and this breaks in production, especially when storing user uploaded images to the root of the public storage directory!!

### DIFF
--- a/cli/drivers/LaravelValetDriver.php
+++ b/cli/drivers/LaravelValetDriver.php
@@ -33,10 +33,6 @@ class LaravelValetDriver extends ValetDriver
 
         $storageUri = $uri;
 
-        if (strpos($uri, '/storage/') === 0) {
-            $storageUri = substr($uri, 8);
-        }
-
         if ($this->isActualFile($storagePath = $sitePath.'/storage/app/public'.$storageUri)) {
             return $storagePath;
         }


mattstauffer:
Here's the commit. We'll have to ask Taylor what it's for.. I'm a bit surprised to see it in there, tbh.

https://github.com/laravel/valet/commit/7e57810a7f2a04c16c7e16f28a18b6f37fe969c9

mattstauffer:
@CaddyDz Sorry for the delay, I had a very busy summer. I looked at this code again, with some other folks looking at it with me, and can't actually figure out how you could have the situation you're describing.

If you have time, would you be able to walk me through the steps I should take to reproduce it locally? Thanks!

CaddyDz:
Hey @mattstauffer , Hope you had a great summer
I reproduced this myself and pushed a sample to GitHub [here](https://github.com/CaddyDz/valet-issue)
Also pushed that exact app to a DO droplet untouched [here](https://valet.salimdj.me)

But if you have the time here are the....
# Steps to reproduce

1. Create a new Laravel 6.0 application using the Laravel installer
```shell
laravel new valet-issue
```
2. copy an image to `storage/app/public` directory
```shell
cp ~/Desktop/image.jpg ~/Sites/valet-issue/storage/app/public
```
3. Link to that image in the `welcome.blade.php` view
```html
@verbatim
<!-- <img src="{{ asset('image.jpg') }}" alt="Image not found"> -->
@endverbatim
<style>
    * {
        width: 99%;
        height: 99%;
    }
</style>
<img src="{{ asset('image.jpg') }}" alt="Image not found">
```
4. Access the site served by Laravel Valet
`http://valet-issue.test`
We can see the image tho we didn't link the storage directory
5. Deploy the app to a Digital Ocean droplet using Forge for example
6. Access the site served by nginx
We can not see the image! 
7. link to storage directory
```shell
php artisan storage:link
```
Still same issue
Now obviously the solution is to 
```php
{{ asset('storage/image.png') }}
```
But the developer using Valet doesn't have to do that and so they forget to properly link to static assets with consideration of production
### Solution?
Force the developer to properly link to static assets in storage when using Valet

================
Now I understand that this feature in Valet was meant to be a convenience but at least for me and for many developers I know, it only has been the opposite

keithbrink:
I agree that this should be removed, I just spend the last 30 minutes trying to sort out why a file in my storage folder can be accessed directly.

I believe you would also need to remove the follow code:

```php
$storageUri = $uri;

if ($this->isActualFile($storagePath = $sitePath.'/storage/app/public'.$storageUri)) {
    return $storagePath;
}
```

PR #793, opened 2019-07-09
Add link to upgrade instructions in outdated version error message

nick-f:
Previously when checking if the Valet version in use was the latest,
you'd have to go to Valet repo on GitHub, find out that all the docs
were on laravel.com, go there, find the upgrade instructions, and then
upgrade.

This can be simplified by giving the user the option to view the link.

Sample output:
```
$ ./valet on-latest-version
Your version of Valet (2.3.3) is not the latest version available.
Upgrade instructions can be found in the docs: https://laravel.com/docs/valet#upgrading
```

Future potential improvements:
- Add option asking the user if they would like to have the docs opened for them (and open the page in their browser if they do want that)

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -262,9 +262,10 @@ if (is_dir(VALET_HOME_PATH)) {
      */
     $app->command('on-latest-version', function () use ($version) {
         if (Valet::onLatestVersion($version)) {
-            output('YES');
+            output('Yes');
         } else {
-            output('NO');
+            output(sprintf('Your version of Valet (%s) is not the latest version available.', $version));
+            output('Upgrade instructions can be found in the docs: https://laravel.com/docs/valet#upgrading');
         }
     })->descriptions('Determine if this is the latest version of Valet');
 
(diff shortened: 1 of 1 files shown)


drbyte:
While this is a helpful idea, the `on-latest-version` command is actually an internal command for Valet, and specifically expects a literal `YES`/`NO` response. 
Ideally the command would be hidden from the list.

I wonder if the upgrade text you're adding could be added to the top of the list where Valet displays its version number. 
Or maybe the `Application` `description` could be updated to include a link to the docs and to the upgrade part of the docs, independent of how the version number is currently displayed.

nick-f:
@drbyte Ah right, I didn't see any results for `on-latest-version` when doing a search, so I assumed it was just a check to let the user know if they should upgrade without checking GitHub, the docs, or just trying to upgrade periodically.

I like the idea of having it displayed near the version number. I'll have a think about how to cache the results so it's not checking every any Valet command is run.

drbyte:
In hindsight, maybe it doesn't matter. 
Digging thru the code history I found that the code that relied on the `on-latest-version` command was removed 3 years ago, [in v1.1.15](https://github.com/laravel/valet/commit/13cf1a85e18e4ba44bbec724c1163a05e30c922c#diff-f86c8d410e6e8e236e0370931e28e73e)

Based on that, unless Taylor and Adam are depending on those removed-more-than-three-years-ago scripts (which wouldn't work today anyway unless updated), I don't see why this couldn't be merged.

Issue #729, opened 2019-02-13
Returning a remote url with isStaticFile creates an internal redirect loop


Mike-Hermans:
When working with projects locally, I like to fetch my images from the live version of the website, so that I do not have to download a copy of all images whenever I update my database. I tried doing this by extending the existing WordPressValetDriver and check if a request is for an image, if it is, it tries to find the file on the remote website and if it exists there, it will return its url. Here's the code, where I create a .valetconf.json in my project root containing the following:
```
{
    remoteAssetsUrl: "https://<sitehere>.com"
}
```

And this is the code I use for my driver:
```
class WordPressExtendedValetDriver extends WordPressValetDriver
{
    /**
     * Determine if the incoming request is for a static file.
     *
     * @param  string  $sitePath
     * @param  string  $siteName
     * @param  string  $uri
     * @return string|false
     */
    public function isStaticFile($sitePath, $siteName, $uri)
    {
        $fileExtension = strtolower(pathinfo($uri, PATHINFO_EXTENSION));

        if (preg_match('/(jpe?g|png|svg|ico)\z/', $fileExtension)) {
            $valetconf = $sitePath . '/.valetconf.json';

            if (file_exists($valetconf)) {
                $conf = json_decode(file_get_contents($valetconf), true);
    
                if (isset($conf['remoteAssetsUrl'])) {
                    $remote_url = $conf['remoteAssetsUrl'] . $uri;
                    $headers = get_headers( $remote_url );
                    if (stripos($headers[0], '200 OK')) {
                        return $remote_url;
                    }
                }
            }
        }

        return parent::isStaticFile($sitePath, $siteName, $uri);
    }
}
```

All images return a 500 server error, where the following error is logged:
```
2019/02/12 14:39:24 [error] 165#0: *562 rewrite or internal redirection cycle while processing "/Users/mike/.composer/vendor/laravel/valet/server.php", client: 127.0.0.1, server: testsite.valet, request: "GET /wp-content/uploads/2018/06/image-upload.png HTTP/2.0", upstream: "fastcgi://unix:/Users/mike/.config/valet/valet.sock", host: "testsite.valet", referrer: "https://testsite.valet/hello-world/"
```

I did something similar on Valet+ a while ago, but I can't seem to figure this out.

drbyte:
You're getting the 500 error because Valet's serveStaticFile() attempts to serve the returned URL "relative to" the local site document-root.

The following will work instead:
```diff
[...]

Issue #247, opened 2016-12-10
Proxy missing media from production in nginx configuration


EHLOVader:
With the switch to nginx this became more possible and I would like to integrate this for WordPress uploads and perhaps other drivers. It may be best as a plugin but do you think this could serve others well if added to the core commands like secure. 

When a file isn't found locally it tries the production location for the uri. 

The command setting this up would be something like `valet proxy [productiondomain.com]` if the domain is omitted it would assume folder pathname with .com 

and the config would be similar to this from the [VVV site wizard](https://gist.github.com/brichards/68c9686066e7dc9df5b9#file-site-conf)

```
# Directives to send expires headers and turn off 404 error logging.
location ~* \.(js|css|png|jpe?g|gif|ico)$ {
    expires 24h;
    log_not_found off;
    try_files $uri $uri/ @production;
}

location @production {
    resolver 8.8.8.8;
    proxy_pass http://productiondomain.com/$uri;
}
```

hybridvision:
I have been trying to get this to work and if I add the above code to the nginx configuration for one of my sites manually and then restart Valet, I get a `502 Bad Gateway` response. I'm using valet 2.0.3, freshly updated today.

I tried some other configurations like replacing `proxy_pass` with a redirection by doing:

```
location @production {
    return 302 https://www.example.com$uri;
}
```
This works but it redirects _everything_, not just missing files. It's almost as if the `try_files` line is not recognising that any of the files exist locally, although they clearly do. I've spent far too long on this and it's a mystery... I guess it must be related to some other part of the configuration for nginx in Valet but I'm still a beginner with nginx.

Has anyone managed to make this work properly?

chood531:
@hybridvision 

Just in case you still need it, here is a working version:

Be sure to:

1. replace `your-username-here` with the username in your local env
2. replace `site-name` with the site name
3. replace `example.com` with the remote URL

```
location ~* .(png|jpe?g|gif|ico|svg)$ {
        expires 24h;
        log_not_found off;
        root '/Users/your-username-here/.valet/Sites/site-name/';
        if (-f $request_filename) {
            break;
        }
        try_files $uri $uri/ @production;
    }
    location @production {
        resolver 8.8.8.8;
        proxy_pass https://example.com/$uri;
    }
```

hybridvision:
Hi @chood531, after a long gap, I needed this again today so I came searching and I found your reply that I'd missed earlier (thanks!).

[...]

PR #811, opened 2019-09-07
Fire the .valet-env environment var detection earlier

drbyte:
To support smarter detection of static-file detection and allow Valet to mutate the URI for externally-hosted assets (save bandwidth and local storage etc), detecting any custom-inserted environment variables needs to happen slightly sooner in the execution cycle.

Discovered this issue when troubleshooting #729 and #247. While they have other workarounds, this PR simplifies things, and potentially makes custom drivers slightly easier too ... when the use-case applies.


mattstauffer:
Thanks @drbyte!

PR #814, opened 2019-09-07
Improve security by listening on 127.0.0.1 only

### LINKED ISSUE #772: Add support for listening on 127.0.0.1 only
It would be great if there was a way to generate sites using a secure configuration that sets nginx to only listen on 127.0.0.1 instead of all interfaces. There may be a way to do this, but I couldn't find any documentation on it.

I recently setup Valet and once I had it up and running, I noticed that nginx was listening on port 60, 80, and 443 on all interfaces.
```
tcp4       0      0  *.60                   *.*                    LISTEN     
tcp4       0      0  *.443                  *.*                    LISTEN     
tcp4       0      0  *.80                   *.*                    LISTEN 
```
For security purposes, I'd prefer to only have those ports accessible from my computer and not from remote hosts. So I'd like it to be this:
```
tcp4       0      0  127.0.0.1.60                   *.*                    LISTEN     
tcp4       0      0  127.0.0.1.443                  *.*                    LISTEN     
tcp4       0      0  127.0.0.1.80                   *.*                    LISTEN 
```
It seems the configuration templates are stored in these two locations:
```
~/.composer/vendor/laravel/valet/cli/stubs/valet.conf
~/.composer/vendor/laravel/valet/cli/stubs/secure.valet.conf
```
Once I setup a new site, it stores an nginx configuration file for each website here:
```
~/.config/valet/Nginx
```
So to accomplish what aI was trying to do, I had to edit all the files in this folder to change them to listen on 127.0.0.1.

Having a command line switch to the "link" command that enables you to tell nginx to listen just on 127.0.0.1 would be great.

drbyte:
Fixes #772

I've been using this config change since Aug 17, 2019, without any negative side-effects.
All Valet services continue to work properly, and Valet Share still works just as expected.

If someone were to have a challenge with it, there's an easy downgrade: just remove the `127.0.0.1:` from these files, and run `valet tld test` to rebuild the individual site configs. Or just manually edit the `~/.config/valet/Nginx` site file manually.

### DIFF
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -1,11 +1,11 @@
 server {
-    listen 80;
+    listen 127.0.0.1:80;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
 }
 
 server {
-    listen 443 ssl http2;
+    listen 127.0.0.1:443 ssl http2;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
@@ -43,7 +43,7 @@ server {
 }
 
 server {
-    listen 60;
+    listen 127.0.0.1:60;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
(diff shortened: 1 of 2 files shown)


mattstauffer:
Great--thanks for testing it this whole time!

Issue #815, opened 2019-09-08
Installation fails


grantholle:
I had it installed successfully, but was having trouble using `valet share`. I found others who said to uninstall valet, update composer and reinstall valet. Which I did.

I've used homebrew to install composer:

```
➜ composer --version                   
Composer version 1.9.0 2019-08-02 20:55:32
➜ php -v             
PHP 7.3.9 (cli) (built: Aug 30 2019 01:42:25) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.9, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.9, Copyright (c) 1999-2018, by Zend Technologies
```

But when trying to `composer global require laravel/valet` I get the [following failure](https://gist.github.com/grantholle/bd6749b46aea859a12ced1c1cb45979e).

I've removed the global lock file and the global vendor directory, reran `composer global install`, and still receive the same installation error.

drbyte:
Looks like you've got conflicts in your global composer config.
What's the content of your `~/.composer/composer.json` file?

grantholle:
I've done some tinkering, but here's the last attempt:

```json
{
    "require": {
        "laravel/installer": "~1.1",
        "laravel/envoy": "dev-master",
        "laravel/lumen-installer": "dev-master",
        "bamarni/symfony-console-autocomplete": "dev-master",
        "hirak/prestissimo": "dev-master"
    }
}
```

Could it be the Laravel installer?

drbyte:
What's `composer global why illuminate/contracts` report?

grantholle:
```
➜ composer global why illuminate/contracts
Changed current directory to /Users/grant/.composer
illuminate/support  v6.0.0  requires  illuminate/contracts (^6.0) 
```

drbyte:
now the same with illuminate/support

grantholle:
```

drbyte:
Looks like laravel/envoy is your issue.  Why dev-master ?

grantholle:
Not sure? I'll remove it and see what changes. Thanks @drbyte

drbyte:
Overall explanation: One of your global dependencies is accepting 6.0, but others are not yet.

The simplest way to downgrade is to either add a dependency on an older version of one of those dep's, or re-install all of them "at one time" so Composer calculates the best match.

Here's a quick way to do it (deletes all previous matchings, and then builds fresh with the new addition).
```sh
[...]

PR #816, opened 2019-09-08
Add compatibility with global illuminate 6.0 pkgs

drbyte:
While the "technical" underlying code hasn't changed much, the version number dependencies do create some challenges (ref #815). Hence, I think it's sensible to update the versions.

Tests appear to pass properly when updating to ^6.0 and thus far it runs happily in my environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,11 +25,11 @@
     },
     "require": {
         "php": ">=5.6",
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks @drbyte, both for handling #815 and for fixing this here!

Issue #808, opened 2019-08-14
subdomain with hyphen crashes nginx


sgotre:
Laravel Valet 2.3.3
Added a subdomain containing more than one hyphen.
This crashes nginx. 

Works:
```
$  valet link --secure subdomain-minus.mydomain
```

nginx can not start anymore
```
$  valet link --secure subdomain-minus-test-test.mydomain
```

You can get valet back working by deleting this file: 
```
 /Users/{MY_USER_ACCOUNT}/.config/valet/Nginx/subdomain-minus-test-test.mydomain.{YOURTLD}
```

drbyte:
Using `dev-master` (which isn't significantly different presently from v`2.3.3`), I can't replicate your symptoms, at least not using your exact example.
If your "actual" subdomain+domain is very very long, there's a known issue with that being problematic (I don't recall it being related to hyphens at all though). 
But the length in your example doesn't cause any problem for me:

<img width="1155" alt="Screen Shot 2019-08-15 at 9 12 10 AM" src="https://user-images.githubusercontent.com/404472/63096806-d987a800-bf3c-11e9-84df-71161ba967ab.png">
<img width="1174" alt="Screen Shot 2019-08-15 at 9 08 03 AM" src="https://user-images.githubusercontent.com/404472/63096814-dc829880-bf3c-11e9-84ca-39a1acaeb1e3.png">

sgotre:
@drbyte Thanks for you fast reply
yes, i had issues with longer domains. 
Just retried with long domain names. You are right. It is not related to hyphens, but to domain length. 

**TLDR: Total domain length has a limit of 43 characters. Valet is not throwing an error**

You can try it yourself (i tested on v2.3.3)

valet link --secure 20characters12345678  
 => working

valet link --secure 30characters123456789123456789
 => working

valet link --secure  36characters123456789123456789123456
=> working

valet link --secure  37characters1234567891234567891234567
=> working

valet link --secure  38characters12345678912345678912345678
=> failed

All longer domain names fail: 
valet link --secure  40characters1234567891234567891234567890
=> failing

**The domain length has a limit of 37 characters when using TLD .test**

When switching the TLD to .a:

valet link --secure 40characters1234567891234567891234567890
=> works

valet link --secure 41characters12345678912345678912345678901
 => fails

So the total domain length has a limit of 43 characters (including .tld)
Maybe valet should throw an error if a domain is longer than that.

drbyte:
[...]

Issue #543, opened 2018-03-28
Securing long domain names crash nginx


0x6368:
When I create a long domain name (e.g. `valet link 12345678901234567890123456789012345678901234567890`) and then `valet secure 12345678901234567890123456789012345678901234567890`, I cannot reach any of my valet links any more, because nginx will not start any more.

To fix this, I have to `valet unsecure 12345678901234567890123456789012345678901234567890`.

I think valet should prevent securing such links.

drbyte:
Confirmed. 
37 characters is the max that appears to work when running `valet secure`. Longer than that stops nginx from reloading properly. (Not a problem with `valet link` though.)

drbyte:
Update: Workaround posted at: https://github.com/laravel/valet/issues/808#issuecomment-521651937

drbyte:
This can be closed.

PR #818, opened 2019-09-09
Allow for longer domain names

drbyte:
Increases domain name length support in Valet per #543 and #808

These are the same settings used in Forge.


drbyte:
@mattstauffer remember to version-bump ;P

mattstauffer:
❤️  lol

drbyte:
Notes: to enable on existing installs, run `valet install`. 
Be aware that `valet install` will replace various custom nginx fastcgi and php-fpm configs.

PR #819, opened 2019-09-09
Turn off logging of hits to robots.txt and favicon.ico

drbyte:
A long-requested adjustment to Valet is to bypass the logging of robots.txt and favicon.ico hits. Particularly when keeping an Ngrok session alive.

This update is consistent with default logging settings in Forge.


drbyte:
Notes: to enable on existing installs, run `valet install`. 
Be aware that `valet install` will replace various custom nginx fastcgi and php-fpm configs.

PR #821, opened 2019-09-27
Allow custom .php files to be executed for Magento2

igormatkovic:
By Default Valet pushes all non-static file requests to index.php without checking if the requested file even exists. 

In Production, this would be whitelisted inside of .htaccess or nginx. 

Currently, in Valet there is no option to allow custom php files in this driver

### DIFF
--- a/cli/drivers/Magento2ValetDriver.php
+++ b/cli/drivers/Magento2ValetDriver.php
@@ -142,6 +142,11 @@ class Magento2ValetDriver extends ValetDriver
     public function frontControllerPath($sitePath, $siteName, $uri)
     {
         $this->checkMageMode($sitePath);
+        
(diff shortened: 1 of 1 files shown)


mattstauffer:
@igormatkovic Since I don't use Magento it's hard for me to review this. Could you ask some other Magento-and-Valet-using folks to check it out?

cli/drivers/Magento2ValetDriver.php
```diff
@@ -142,6 +142,11 @@ private function isPubDirectory($sitePath, $route, $pub = '')
     public function frontControllerPath($sitePath, $siteName, $uri)
     {
         $this->checkMageMode($sitePath);
+        
+        if (file_exists($sitePath . $uri) {
```
hulkur: missing parenthesis

Issue #823, opened 2019-10-02
Issue with certificates and tld's with more than one '.'


jrbarnard:
I use a tld with multiple segments usually, this works fine for accessing the sites and if they are secured using `valet secure`.

However when outputting via `valet links` it is not marked as secure.

To recreate:
```
valet tld test
valet secure {a site}
valet links # To show is shown as secured
valet tld test.more.segments.com
valet links # Shows unsecured
```

Digging into the code it looks like the issue is in Site::getCertificates() where it strips off the last segment.

It may be better to update this to just strip the tld we have in config off the end to be more robust. I've checked and when changing the tld, the certificate paths are updated so this should work even if tld's are changed after the certificate generation.

I'll submit a PR to fix

jrbarnard:
Merged #825

PR #825, opened 2019-10-02
Check for tld on end of certificate path to support multi segment tld's

jrbarnard:
Fixes bug #823

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -120,11 +120,21 @@ class Site
      */
     function getCertificates($path)
     {
+        $config = $this->config->read();
+
         return collect($this->files->scandir($path))->filter(function ($value, $key) {
             return ends_with($value, '.crt');
-        })->map(function ($cert) {
+        })->map(function ($cert) use ($config) {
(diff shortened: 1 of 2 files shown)


cli/Valet/Site.php
```diff
@@ -120,11 +120,21 @@ function links()
...
+
         return collect($this->files->scandir($path))->filter(function ($value, $key) {
             return ends_with($value, '.crt');
-        })->map(function ($cert) {
+        })->map(function ($cert) use ($config) {
             $certWithoutSuffix = substr($cert, 0, -4);
-            return substr($certWithoutSuffix, 0, strrpos($certWithoutSuffix, '.'));
+            $trimToString = '.';
+
+            // If we have the cert ending in our tld stripe that tld specifically
```
drbyte: `stripe` vs `strip`?

jrbarnard: @drbyte Good catch, I'll update

drbyte: 👍

drbyte:
Looks good to me.

PR #826, opened 2019-10-02
Fix #708. Use stopRunning instead of stop to only stop running php

### LINKED ISSUE #708: When running valet stop it runs stop on all installed php versions, this can be slow

jrbarnard:


drbyte:
👍

Issue #829, opened 2019-10-09
Paths including a plus symbol ("+") are not resolved


jamieburchell:
While I do not personally use plus symbols in filenames, I inherited a project that has them in directory names today and discovered that valet does not resolve them like a regular web server does.

Plus symbols do not need to be escaped in file paths according to the RFC and can be used literally:

`some+file.jpg`

Due to the use of `urldecode` [here](https://github.com/laravel/valet/blob/master/server.php#L66) the plus symbol is incorrectly decoded to a space character and the file is never resolved. It's the correct behaviour to decode the plus symbol to a space when used in a query string, but not for this use case.

Changing the method to `rawurldecode` fixes the issue.

Issue #320, opened 2017-02-16
Static paths with "+" signs in them won't load.


miladmostavi:
My driver's isStaticFile is the standard one:

```
public function isStaticFile($sitePath, $siteName, $uri)
{
    if (file_exists($staticFilePath = $sitePath.'/app/public_html/'.$uri)) {
        return $staticFilePath;
    }

    return false;
}
```

All files works fine, except: `/js/vendor/history.js/scripts/bundled/html4+html5/jquery.history.js`

miladmostavi:
I temporarily solved my issue by replacing file names which have spaces in them with `+`.

```
public function isStaticFile($sitePath, $siteName, $uri)
{
    $staticFilePath = $sitePath.'/app/public_html'.$uri;
    if(preg_match('/\s/', $staticFilePath) && ! file_exists($staticFilePath)) {
        $staticFilePath = str_replace(' ', '+', $staticFilePath);
    }

    if (file_exists($staticFilePath)) {
        return $staticFilePath;
    }

    return false;
}
```

The `$uri` that is received by the function comes with a space instead of the original `+` sign.

mattstauffer:
Closed via #830

PR #830, opened 2019-10-09
Fix incorrectly decoding plus symbol to space in paths

jamieburchell:
Fixes https://github.com/laravel/valet/issues/829


drbyte:
Does this also fix #320 ?

jamieburchell:
> Does this also fix #320 ?

Yes.

PR #834, opened 2019-10-21
Ignore parked paths that are not directories when getting sites

jrbarnard:
When I added the parked command I forgot about non directory paths, I only noticed this today when a .DS_Store file was created in the directory and showed up in the parked output.

This fix filters out the sites by ensuring the path is a directory.


mattstauffer:
Thanks @jrbarnard!

PR #835, opened 2019-10-21
Do not show sites in parked if they would not be served

jrbarnard:
Another issue I saw, this with duplicate site names.
Currently a linked site takes precedence over a parked site so it shouldn't be shown under 'parked' as it will never be served.
This also fixes an issue if you have multiple parked directories, the later parked ones would overwrite the previous if they share a site name, I've reversed this.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -125,14 +125,21 @@ class Site
 
         $certs = $this->getCertificates($certsPath);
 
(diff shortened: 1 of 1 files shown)


drbyte:
Agreed. This makes sense.

mattstauffer:
Thanks @jrbarnard and thanks as always @drbyte for reviewing!

PR #845, opened 2019-11-28
Allow support for symfony/process 5.0

drbyte:
For compatibility with current symfony release updates

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "php": ">=5.6",
         "illuminate/container": "~5.1 | ^6.0",
         "mnapoli/silly": "~1.0",
-        "symfony/process": "~2.7|~3.0|~4.0",
+        "symfony/process": "~2.7|~3.0|~4.0|~5.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
         "php": ">=5.6",
         "illuminate/container": "~5.1 | ^6.0",
         "mnapoli/silly": "~1.0",
-        "symfony/process": "~2.7|~3.0|~4.0",
+        "symfony/process": "~2.7|~3.0|~4.0|~5.0",
```
ravenberg: Hi, would it hurt to remove the `~2.7` and `~3.0` version requirements?

drbyte: Potentially. 
I haven't gone back and checked what would break if we forced those out.  Ideally forcing a newer PHP version would be easier to maintain
... but there are still people who do 5.6 (via unofficial taps of course), so if it's not necessary to break things, then I've just been leaving those around.

PR #851, opened 2019-11-30
Add detailed valet uninstall instructions

drbyte:
Previously `valet uninstall` would just shutdown the `nginx` service.

This PR also stops `PHP` and gives a bunch of instructions on how to do a (nearly) clean uninstall of most Valet dependencies.
And also some troubleshooting tips.

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -261,9 +261,46 @@ if (is_dir(VALET_HOME_PATH)) {
      * Uninstall Valet entirely.
      */
     $app->command('uninstall', function () {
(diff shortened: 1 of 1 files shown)


drbyte:
@mattstauffer note that these uninstall instructions require #854 to be merged. ... or we can change the wording to reflect an alternate approach to cleaning up old certs. (eg: run `valet links` and `valet parked` to get a list, and unsecure them all individually by hand)

mattstauffer:
@drbye 👌

PR #852, opened 2019-11-30
Display nginx config errors on start

drbyte:
While since #268 valet has been checking nginx configs for errors when starting/restarting, the captured errors are never displayed if there is a failure because it's being run using `quietly()`. 
This PR causes the errors to pass through to the console so we can more readily understand why nginx may not be starting or not serving properly.


mattstauffer:
You are the best

PR #855, opened 2019-11-30
Set robots nofollow for all valet share traffic

drbyte:
Replaces and closes #575 ... for reasons described there.

I've been using this for a couple months without issue. (Granted, I don't leave valet share running for extended periods.)


mattstauffer:
OOh yah i like this way much better

PR #856, opened 2019-11-30
Use custom dnsmasq.d folder for easier config and support

drbyte:
Dnsmasq allows use of a common config folder, much like nginx and php do.

This PR changes Valet's default config to empower this feature.

This will make for easier troubleshooting
... and easier customizing (such as dropping in a custom logging config, additional TLDs, alternate DNS resolvers, etc)

### DIFF
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -9,9 +9,9 @@ class DnsMasq
 {
     var $brew, $cli, $files;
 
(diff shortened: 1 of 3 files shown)


drbyte:
This probably should also include doing a cleanup of the old config file from previous Valet installs.

drbyte:
Updated to include doing a cleanup of the old config file from previous Valet installs.

PR #857, opened 2019-12-01
Update .travis.yml

drbyte:
While `7.4snapshot` is still functional, it's no longer needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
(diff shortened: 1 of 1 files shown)


drbyte:
Argh - looks like I spoke too soon. 
I'm sure someday soon the "snapshot" will be fully renamed.

drbyte:
@mattstauffer if you wanna tell Travis to re-build the 7.4 run, I think it may be fine now.

mattstauffer:
heyyyyy good!

PR #858, opened 2019-12-01
Move PHP-FPM pool config to separate valet-specific file

drbyte:
This allows the valet FPM pool configuration to stand separately from the default config.

This benefits troubleshooting, makes customization of FPM workers and other settings easer
and allows for easier uninstallation.

(Note: existing Valet installs will have had their PHP `/usr/local/etc/php/7.4/php-fpm.d/www.conf` files already altered by Valet. It will be best to restore this file to the default version installed by PHP to prevent duplicate FPM pool instances that conflict with each other. 
Or, simply delete the `www.conf` file.)

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -54,21 +54,35 @@ class PhpFpm
     {
         info('Updating PHP configuration...');
 
-        $contents = $this->files->get($this->fpmConfigPath());
+        $fpmConfigFile = $this->fpmConfigPath();
 
-        $contents = preg_replace('/^user = .+$/m', 'user = '.user(), $contents);
-        $contents = preg_replace('/^group = .+$/m', 'group = staff', $contents);
-        $contents = preg_replace('/^listen = .+$/m', 'listen = '.VALET_HOME_PATH.'/valet.sock', $contents);
-        $contents = preg_replace('/^;?listen\.owner = .+$/m', 'listen.owner = '.user(), $contents);
-        $contents = preg_replace('/^;?listen\.group = .+$/m', 'listen.group = staff', $contents);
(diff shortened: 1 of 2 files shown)


mattstauffer:
@drbyte I love this. Only question is, how should we handle restoring existing PHP 7+ installs' www.conf files?

drbyte:
@mattstauffer ya, I've been debating how best to handle that. 

While I'd prefer not to clutter Valet with a bunch of "upgrade" code, it's probably unavoidable.

Two options:
1. Delete the "old" `www.conf` pool file altogether. If Valet has already customized it then "default PHP" isn't going to be offended if we just remove it.  **I think this is the best option.**  (And for fresh installs I'm inclined to also rename the `www.conf` to `www.conf.default` so that PHP doesn't actually spin up workers that are just wasting memory.)
2. Restore the old contents. Of primary concern is the `listen=` line so that it doesn't point to `valet.sock`. While the other things deserve cleanup too, they're not what's going to "conflict" with Valet.

mattstauffer:
@drbyte I 100% agree with you on all points. I don't *want* the junk upgrade code but not sure what else to do. I prefer the nuclear option, and I love the idea of renaming it on new ones.

drbyte:
I'll work on an update later this afternoon.

drbyte:
@mattstauffer I'm toying with just renaming the `www.conf` to `www.conf.old` ... regardless of whether Valet has touched it before.

mattstauffer:
@drbyte YES. Why not? If someone needs it, they're already in an advanced-enough situation that they can go look it up.

drbyte:
@mattstauffer Update pushed.

drbyte:
Updated to also make sure the `php-fpm.d` directory exists, since it was reported today in #865 that a `php@7.2` install failed to create it properly for some reason.

mattstauffer:
@drbyte <3

PR #864, opened 2019-12-03
Fix auto-detection amongst multiple ngrok processes

### LINKED ISSUE #145: Valet Share does not copy or show the latest ngrok url, but only the first one
If you use it more than once, it'll open up multiple ngrok instances (which is fine), but it will keep displaying and copying to the clipboard only the first one.

drbyte:
Ngrok Pro plans allow multiple processes, which means you could be serving several sites simultaneously. This PR allows Valet to correctly identify amongst the active ngrok process URLs.

Fixes #145

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -226,8 +226,8 @@ if (is_dir(VALET_HOME_PATH)) {
     /**
      * Echo the currently tunneled URL.
      */
-    $app->command('fetch-share-url', function () {
-        output(Ngrok::currentTunnelUrl());
+    $app->command('fetch-share-url [domain]', function ($domain = null) {
+        output(Ngrok::currentTunnelUrl($domain ?: Site::host(getcwd()).'.'.Configuration::read()['tld']));
     })->descriptions('Get the URL to the current Ngrok tunnel');
 
     /**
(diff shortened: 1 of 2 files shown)


cli/valet.php
```diff
@@ -226,8 +226,8 @@
     /**
      * Echo the currently tunneled URL.
      */
-    $app->command('fetch-share-url', function () {
-        output(Ngrok::currentTunnelUrl());
+    $app->command('fetch-share-url [domain]', function ($domain = null) {
+        output(Ngrok::currentTunnelUrl($domain ?: Site::host(getcwd()).'.'.Configuration::read()['tld']));
```
mattstauffer: Would this struggle if someone's parked it as something else?

drbyte: If they've linked it to something other than the same as the foldername, then the `Site::host()` function resolves the link and uses that instead.

https://github.com/laravel/valet/blob/52b3b47db8ea0f77dee00af3a6c78a7c810cc597/cli/Valet/Site.php#L62-L77

mattstauffer: brilliant

mattstauffer:
You are the best

PR #867, opened 2019-12-06
Improve verbosity

### LINKED ISSUE #317: Verbose option isn't that verbose.
The verbose option isn't really that verbose. Debugging a valet start which says everything works fine is quite hard when it doesn't log anything at all and requires edits in the script it self. See for example https://github.com/laravel/valet/issues/305.

Maybe don't show the option if it doesn't change anything.

drbyte:
Detect whether `-v` is passed to certain Valet commands, and cause that to display more detailed Homebrew output in case there's a need to troubleshoot.
(I don't think every command needs a bunch of extra verbosity. But restarting of services, and initial install seem to be logical times to want more verbosity.)

Ideally there'd be a better way to persist the value without having to pass it around all over. Insights appreciated.

Fixes #317

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -152,7 +152,7 @@ class Brew
      *
      * @param
      */
-    function restartService($services)
+    function restartService($services, $verbosity)
     {
         $services = is_array($services) ? $services : func_get_args();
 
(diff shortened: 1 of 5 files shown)


drbyte:
> Ideally there'd be a better way to persist the value without having to pass it around all over. Insights appreciated.

The failing tests just drive this point even harder. The changes feel too invasive. There's gotta be a better way ...

PR #869, opened 2019-12-06
Delay pbcopy of ngrok URL to wait for service startup

### LINKED ISSUE #868: After running valet share the share url is not automatically copied to the clipboard.
**Valet**: 2.5.4 **Ngrok**: 2.3.3 5 **PHP**: 7.4 **Mac OS**: 10.15.1

Now share works fine, but the share url is not automatically copied to the clipboard. Something is being copied, but it appears as blank on the clipboard. After a while I am able to use **valet fetch-share-url | pbcopy** to copy the url, but I would like if it is copied to the clipboard automatically.

drbyte:
Fixes #868

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -16,6 +16,9 @@ class Ngrok
      */
     function currentTunnelUrl($domain = null)
     {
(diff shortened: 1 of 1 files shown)


drbyte:
Note: #872 includes this same fix, so merging it will close this one.

PR #112, opened 2016-08-14
Pass parameters to ngrok

renedekat:
This patch enables passing custom parameters like --region or -subdomain.

valet share --region=eu -subdomain=my_custom_domain

If you have a paid version of Ngrok the subdomain param is available, which is very handy when working with customers. Share the link once, instead of sharing a new random link every time you want your client to look at your progress.

PR #870, opened 2019-12-06
Allow valet share to support multiple domains from same app

### LINKED ISSUE #537: valet share + multiple domains served from same Laravel app
Currently I have multiple domains being served by Valet, from the same Laravel app. I need a way to specific which domain to redirect the ngrok URL to. It looks like Valet is automatically determining this, and I'm not sure how to override it.

Ideally I'd like to be able to do something like:

`valet share domain2.test`

drbyte:
This PR allows passing a specific domain name to `valet share` in order to have the app be served for that domain.
Now how it works is this:
- **(NEW)** if a domain name is passed, it checks whether it matches a link (which would be required for multiple domains served by same app project), and uses that
- if a domain name is NOT passed, it looks up any links for current directory, and uses the first found link
- else falls back to current project foldername

It also still allows passing through custom ngrok parameters if desired (ref: #112), as either the 2nd parameter (no domain name passed) or **(NEW)** 3rd parameter

Fixes #537

### DIFF
--- a/valet
+++ b/valet
@@ -31,7 +31,7 @@ fi
 # process to retrieve the live Ngrok tunnel URL in the background.
 if [[ "$1" = "share" ]]
 then
-    # check for parameters to passthru to ngrok (they start with '-' or '--')
+    # Check for parameters to pass through to ngrok (these will start with '-' or '--')
     PARAMS=${3:-$2}
     if [[ ${PARAMS:0:1} != '-' ]]; then
         PARAMS=''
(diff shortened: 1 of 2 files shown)


valet
```diff
@@ -31,18 +31,38 @@ fi
 # process to retrieve the live Ngrok tunnel URL in the background.
 if [[ "$1" = "share" ]]
 then
+    # check for parameters to passthru to ngrok (they start with '-' or '--')
```
mattstauffer:
```suggestion
    # Check for parameters to pass through to ngrok (these will start with '-' or '--')
```

valet
```diff
@@ -31,18 +31,38 @@ fi
...
+    fi
+
     HOST="${PWD##*/}"
-    TLD=$(php "$DIR/cli/valet.php" tld)
 
-    for linkname in ~/.config/valet/Sites/*; do
-        if [[ "$(readlink $linkname)" = "$PWD" ]]
-        then
-            HOST="${linkname##*/}"
+    # check for custom domain passed through to the share command ($2 w/o '-' prefix)
```
mattstauffer:
```suggestion
    # Check for custom domain passed through to the share command ($2 w/o '-' prefix)
```

valet
```diff
@@ -31,18 +31,38 @@ fi
...
+    if [[ ${2:0:1} != '-' ]]; then
+        # if not blank and is a link or is the cwd use it
```
mattstauffer:
```suggestion
        # If not blank and is a link, or is the cwd, use it
```

valet
```diff
@@ -31,18 +31,38 @@ fi
...
+        if [[ ! -z $2 && (-L ~/.config/valet/Sites/$2 || $2 == $HOST) ]]; then
+            HOST=$2
+            CLIHOST=$2
         fi
-    done
+    fi
+
+    # if no custom domain passed then check if there's a linked site for cwd
```
mattstauffer:
```suggestion
    # If no custom domain passed, then check if there's a linked site for cwd
```

valet
```diff
@@ -31,18 +31,38 @@ fi
...
+    if [[ -z $CLIHOST ]]; then
+        # find the first linked site for the current dir, if one exists
```
mattstauffer:
```suggestion
        # Find the first linked site for the current dir, if one exists
```

mattstauffer:
🙌

fgilio:
Thanks a lot for this!

jordanade:
Can't get this to work. Valet share always picks the first linked domain.

lafs-cloud:
> Can't get this to work. Valet share always picks the first linked domain.

Same

PR #872, opened 2019-12-07
Part 2: Fix auto-detection amongst multiple ngrok processes

### LINKED ISSUE #868: After running valet share the share url is not automatically copied to the clipboard.

drbyte:
Thanks to the generosity of Alan at Ngrok, I've been able to do some additional testing with temporary added features.
Turns out #864 was incomplete. This PR now also loops through the default endpoints in order to find a process matching the passed domain.

Ref: #864
Ref: #145

Also supercedes and Closes #869 and Fixes #868 the same way #869 proposed.

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -7,7 +7,10 @@ use DomainException;
 
 class Ngrok
 {
-    var $tunnelsEndpoint = 'http://127.0.0.1:4040/api/tunnels';
+    var $tunnelsEndpoints = [
+        'http://127.0.0.1:4040/api/tunnels',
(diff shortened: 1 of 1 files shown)


mattstauffer:
I trust you 🤷‍♂  :) Thanks as always for your hard work!

PR #874, opened 2019-12-09
Allow start/restart/stop single service

ahmedash95:
Sometimes all we just need is to restart nginx or php-fpm during development. so having a way to restart a specific service saves some time instead of waiting for dns-masq, php and nginx to restart

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -235,81 +235,79 @@ if (is_dir(VALET_HOME_PATH)) {
      * Start the daemon services.
      */
     $app->command('start [service]', function ($service) {
-        switch($service) {
+        switch ($service) {
             case '':
                 DnsMasq::restart();
                 PhpFpm::restart();
                 Nginx::restart();
 
-                info('Valet services have been started.');
-                return;
-            case 'dns-masq':
+                return info('Valet services have been started.');
+            case 'dnsmasq':
                 DnsMasq::restart();
-                info('dnsmasq has been started.');
-                return;
+
+                return info('dnsmasq has been started.');
             case 'nginx':
                 Nginx::restart();
-                info('Nginx has been started.');
-                return;
+
+                return info('Nginx has been started.');
             case 'php':
                 PhpFpm::restart();
-                info('PHP has been started.');
-                return;
+
+                return info('PHP has been started.');
         }
 
-        return warning(sprintf('Invalid valet service name [%s]',$service));
+        return warning(sprintf('Invalid valet service name [%s]', $service));
     })->descriptions('Start the Valet services');
 
     /**
      * Restart the daemon services.
      */
     $app->command('restart [service]', function ($service) {
-        switch($service) {
+        switch ($service) {
             case '':
                 DnsMasq::restart();
                 PhpFpm::restart();
                 Nginx::restart();
 
-                info('Valet services have been restarted.');
-                return;
-            case 'dns-masq':
+                return info('Valet services have been restarted.');
+            case 'dnsmasq':
                 DnsMasq::restart();
-                info('dnsmasq has been restarted.');
-                return;
+
+                return info('dnsmasq has been restarted.');
             case 'nginx':
                 Nginx::restart();
-                info('Nginx has been restarted.');
-                return;
+
+                return info('Nginx has been restarted.');
             case 'php':
                 PhpFpm::restart();
-                info('PHP has been restarted.');
-                return;
+
+                return info('PHP has been restarted.');
         }
         
-        return warning(sprintf('Invalid valet service name [%s]',$service));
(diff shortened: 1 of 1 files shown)


drbyte:
While my initial reaction to this was that it's not needed, I have found myself reaching for it several times in the past couple weeks.

While I'm not a fan of the duplicate code it adds, it's not unbearable (and Valet already does some duplicate-code anyway, and this PR is correctly following the existing patterns in Valet).

So, I'm okay with this one.

👍

ahmedash95:
> While my initial reaction to this was that it's not needed, I have found myself reaching for it several times in the past couple weeks.
> 
> While I'm not a fan of the duplicate code it adds, it's not unbearable (and Valet already does some duplicate-code anyway, and this PR is correctly following the existing patterns in Valet).
> 
> So, I'm okay with this one.
> 
> 👍

Thanks for your comment. I also hate duplicate code too. but as the file is too small and it's not too complicated with repeated lines I think it's okay to have it like the PR now as it follows the already used pattern in Valet.

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
      * Start the daemon services.
      */
-    $app->command('start', function () {
-        DnsMasq::restart();
-
-        PhpFpm::restart();
-
-        Nginx::restart();
+    $app->command('start [service]', function ($service) {
+        switch($service) {
```
mattstauffer:
```suggestion
        switch ($service) {
```

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
+            case '':
+                DnsMasq::restart();
+                PhpFpm::restart();
+                Nginx::restart();
+
+                info('Valet services have been started.');
+                return;
+            case 'dns-masq':
```
mattstauffer:
```suggestion
            case 'dnsmasq':
```

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
+                info('Nginx has been started.');
+                return;
+            case 'php':
+                PhpFpm::restart();
+                info('PHP has been started.');
+                return;
+        }
 
-        info('Valet services have been started.');
+        return warning(sprintf('Invalid valet service name [%s]',$service));
```
mattstauffer:
```suggestion
        return warning(sprintf('Invalid valet service name [%s]', $service));
```

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
+                Nginx::restart();
+                info('Nginx has been restarted.');
+                return;
+            case 'php':
+                PhpFpm::restart();
+                info('PHP has been restarted.');
+                return;
+        }
+        
+        return warning(sprintf('Invalid valet service name [%s]',$service));
```
mattstauffer:
```suggestion
        return warning(sprintf('Invalid valet service name [%s]', $service));
```

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
-
-        Nginx::restart();
+    $app->command('start [service]', function ($service) {
+        switch($service) {
+            case '':
+                DnsMasq::restart();
+                PhpFpm::restart();
+                Nginx::restart();
+
+                info('Valet services have been started.');
```
mattstauffer: Can re modify all of  these to `return` the message? 

`return info('Valet services have been started.');`

Just helpful because it drops an extra line.

mattstauffer: Can you also make sure that you have a newline before each of these switch return statements?

cli/valet.php
```diff
@@ -234,38 +234,82 @@
...
+                info('Nginx has been stopped.');
+                return;
+            case 'php':
+                PhpFpm::stop();
+                info('PHP has been stopped.');
+                return;
+        }
 
-        info('Valet services have been stopped.');
+        return warning(sprintf('Invalid valet service name [%s]',$service));
```
mattstauffer:
```suggestion
        return warning(sprintf('Invalid valet service name [%s]', $service));
```

ahmedash95:
Hey @mattstauffer , thanks for the review. I've fixed all style issues. 
Thanks

cli/valet.php
```diff
@@ -234,38 +234,80 @@
...
+                PhpFpm::stopRunning();
+                Nginx::stop();
+
+                return info('Valet services have been stopped.');
+            case 'nginx':
+                Nginx::stop();
+
+                return info('Nginx has been stopped.');
+            case 'php':
+                PhpFpm::stop();
```
mattstauffer:
```suggestion
                PhpFpm::stopRunning();
```

Sorry, didn't catch this last time around.

ahmedash95: Done :+1:

ahmedash95: Nice catch @mattstauffer . Done 👍

PR #881, opened 2019-12-21
Enable multiple ngrok arguments from command line

comes:
Enable the possibility to pass more than one argument to ngrok.

before this patch, it was not possible to pass more than one argument beginning with --.

now you can add something like this:
valet share --region=eu --authtoken=tokenstring

before this patch, only the first arg --region was used


drbyte:
Good catch. Nicely done fix! 👍

PR #885, opened 2019-12-23
Comment out memory limits so other configs take precedence

### LINKED ISSUE #884: php_admin_value[memory_limit] determines memory limit?
Was trying to change php's memory_limit through the normal methods we've been using for a couple years now (`conf.d/php-memory-limits.ini` / `php.ini`) without success. I couldn't figure out why this was! phpinfo() was always reporting back 128M, regardless of what `php -i | grep memory_limit` returned. 

I had uninstalled php/valet completely, reinstalled, used different versions of php, etc.

Found this recent Stack Overflow question from someone who also got stumped by this (https://stackoverflow.com/questions/59336092/macos-laravel-valet-increase-memory-limit) and it turns out the `php_admin_value[memory_limit] = 128M` within `valet-fpm.conf` is determining the memory limit used everywhere by php?

I can see that adding the valet-fpm.conf file and those values are an intentional change introduced with 2.6.0, but unsure whether that value is supposed to be intentionally restricting the memory_limit. If so, since this is an additional config specific to valet, and not brew/php, should it be documented somewhere?

drbyte:
The default FPM stub was overriding related settings, making customization more complex.
Commenting out the overrides so they can be used only if specific FPM settings need to supercede CLI settings.

Fixes #884

### DIFF
--- a/cli/stubs/etc-phpfpm-valet.conf
+++ b/cli/stubs/etc-phpfpm-valet.conf
@@ -8,10 +8,10 @@ listen.owner = VALET_USER
 listen.group = staff
 listen.mode = 0777
 
-
-php_admin_value[memory_limit] = 128M
-php_admin_value[upload_max_filesize] = 128M
-php_admin_value[post_max_size] = 128M
+# when uncommented, these values will take precedence over settings declared elsewhere
+;php_admin_value[memory_limit] = 128M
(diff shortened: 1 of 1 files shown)


drbyte:
This should fix the too-aggressive settings (accidentally) added in v2.6.0

PR #888, opened 2019-12-27
Adds a directory-listing setting for dev convenience

drbyte:
After running `valet directory-listing on` if a URI points to a directory that exists, but the requested file cannot be found in that directory, a directory listing will be displayed.

A 404 will still happen if the project directory cannot be found.

Running `valet directory-listing off` makes a 404 display (without a listing) if the requested file cannot be found.

Supersedes and Closes #349 and Closes #587

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -462,6 +462,25 @@ You might also want to investigate your global Composer configs. Helpful command
 
         passthru($command);
     })->descriptions('Tail log file');
+
+    /**
+      * Configure or display the directory-listing setting.
+      */
+    $app->command('directory-listing [status]', function ($status = null) {
+        $key = 'directory-listing';
+        $config = Configuration::read();
+
+        if (in_array($status, ['on', 'off'])) {
+            $config[$key] = $status;
+            Configuration::write($config);
+            return output('Directory listing setting is now: '.$status);
+        }
+
+        $current = isset($config[$key]) ? $config[$key] : 'off';
+        output('Directory listing is '.$current);
+    })->descriptions('Determine directory-listing behavior. Default is off, which means a 404 will display.', [
+        'status' => 'on or off. (default=off) will show a 404 page; [on] will display a listing if project folder exists but requested URI not found'
+    ]);
 }
 
 /**
--- a/server.php
+++ b/server.php
@@ -18,6 +18,35 @@ function show_valet_404()
 }
 
 /**
+ * Show directory listing or 404 if directory doesn't exist.
+ */
+function show_directory_listing($valetSitePath, $uri)
+{ 
+    $is_root = ($uri == '/');
+    $directory = ($is_root) ? $valetSitePath : $valetSitePath.$uri;
+
+    if (!file_exists($directory)) { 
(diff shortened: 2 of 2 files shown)


server.php
```diff
@@ -17,6 +17,35 @@ function show_valet_404()
...
+function show_directory_listing($valetSitePath, $uri)
+{ 
+    $is_root = ($uri == '/');
+    $directory = ($is_root) ? $valetSitePath : $valetSitePath.$uri;
+
+    if(!file_exists($directory)) { 
+        show_valet_404(); 
+    }
+
+    // sort directories at the top
```
mattstauffer:
```suggestion
    // Sort directories at the top
```

server.php
```diff
@@ -150,7 +179,11 @@ function valet_default_site_path($config)
 );
 
 if (! $frontControllerPath) {
-    show_valet_404();
+    if (isset($valetConfig['directory-listing']) && $valetConfig['directory-listing'] == 'on') {
```
mattstauffer: I think you can skip the `else` since `show_directory_listing` exits.

cli/valet.php
```diff
@@ -462,6 +462,28 @@
...
     })->descriptions('Tail log file');
+
+    /**
+      * Configure or display the directory-listing setting.
+      */
+    $app->command('directory-listing [status]', function ($status = null) {
+        $key = 'directory-listing';
+        $config = Configuration::read();
+
+        if (is_null($status)) {
```
mattstauffer: Can we re-write this this way?

```php
        if (in_array($status, ['on', 'off'])) {
            $config[$key] = $status;
            Configuration::write($config);
            return output('Directory listing configuration is saved as '.$status);
        }

        $current = isset($config[$key]) ? $config[$key] : 'off';
        output('Directory listing is '.$current);
```

I know it doesn't give directions but I feel like that's ok. What do you think?

server.php
drbyte: But, then the directory-listing switch would have no purpose, right?

cli/valet.php
drbyte: Agreed - I like this brevity. Funny how you can stare at something so long and not see the obvious way to simplify! :)

server.php
mattstauffer: No, I mean do it like this:

```php
    if (isset($valetConfig['directory-listing']) && $valetConfig['directory-listing'] == 'on') {
        show_directory_listing($valetSitePath, $uri);
    }

    show_valet_404();
```

drbyte: Oh. Yes, of course. 👍

mattstauffer:
Gonna save this for the next release so I can do another read-over and test it out.

mattstauffer:
Merging but pausing on a release for a sec. Want to test it a bit through actual Composer.

mattstauffer:
Works great so far. Just trying to make sure we're not accidentally opening up the ability for someone to pass some kind of wonky folder in and then get access to your entire personal documents folder or whatever. Obviously this isn't meant to be open to the public but still.. i'd love to get ahead of that if it were possible.

PR #891, opened 2020-01-04
Cleanup old tld refs when switching tld or uninstalling

drbyte:
Previously we would leave old tld stubs around; while it's harmless, it's less tidy.

### DIFF
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -7,7 +7,7 @@ use Symfony\Component\Process\Process;
 
 class DnsMasq
 {
-    var $brew, $cli, $files;
+    var $brew, $cli, $files, $configuration;
 
     var $dnsmasqMasterConfigFile = '/usr/local/etc/dnsmasq.conf';
     var $dnsmasqSystemConfDir = '/usr/local/etc/dnsmasq.d';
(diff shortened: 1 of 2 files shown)


cli/Valet/DnsMasq.php
```diff
@@ -145,6 +148,7 @@ function createTldResolver($tld)
     function updateTld($oldTld, $newTld)
     {
         $this->files->unlink($this->resolverPath.'/'.$oldTld);
+        $this->files->unlink(VALET_HOME_PATH . '/dnsmasq.d/tld-' . $oldTld);
```
mattstauffer: Should this instead be the same as what we're using to create it?

```php
        $tldConfigFile = $this->dnsmasqUserConfigDir() . 'tld-' . $tld . '.conf';
```

https://github.com/laravel/valet/blob/master/cli/Valet/DnsMasq.php#L120

I'll be honest, though, this is all a bit magical for me.

drbyte:
Tech background:
Dnsmasq requires two configurations for each TLD:
- a dnsmasq conf file (previously we just dropped this in a generic file; now we do it in separate tld-xxxx.conf files), which the dnsmasq app uses to offer DNS
- a resolver file for the OS to proxy DNS lookups into dnsmasq.

We've been properly cleaning up old resolver files when changing TLD, but not the dnsmasq conf file. This PR brings back that consistency.

cli/Valet/DnsMasq.php
drbyte: Agreed - consistency here will make maintenance less complicated.

mattstauffer:
Thanks as always @drbyte!

PR #894, opened 2020-01-04
php config comments must only use semicolons, not hash symbols

drbyte:
PHP configs no longer support `#` for comments. Must use semicolons `;`.

Otherwise PHP may not start, and may show a confusing error in logs: `value is NULL for a ZEND_INI_PARSER_ENTRY`

Updates #885


mattstauffer:
Wow, TIL.

PR #895, opened 2020-01-04
Detect running php when no version constraint specified with use command

### LINKED ISSUE #756: Brew::installed() fails to correctly parse installed version of PHP if "php" is installed with no version constraint
See notes here:

https://github.com/laravel/valet/issues/728#issuecomment-479523782

To reproduce:

Install `php` instead of `php@7.3` with Homebrew, then run `valet use php73`. Valet will suggest it's not installed, even though it is, because it looks at `php` which doesn't have a version number and therefore doesn't match `php73`.

drbyte:
Fixes #756

Previously if `php` was installed as just `php` (the default Homebrew alias), Valet would not detect **_which_** PHP version was actually installed, and therefore `valet use` might do incorrect or unnecessary installations/links/etc.

NOTE: This does NOT "convert" existing `php` alias to a numbered version. It merely accepts it as-is, but notes its version in an attempt to avoid extra installations.

**NOTE: Specifally tested with PHP 7.4 and 7.3. No promises about old 5.6, etc aliases.**

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -72,6 +72,15 @@ class Brew
     }
 
     /**
+     * Get the aliased formula version from Homebrew
+     */
+    function determineAliasedVersion($formula)
+    {
+        $details = json_decode($this->cli->runAsUser("brew info $formula --json"));
+        return $details[0]->aliases[0] ?: 'ERROR - NO BREW ALIAS FOUND';
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


drbyte:
Note: Best to merge bugfix #894 before testing this.

mattstauffer:
I love it. The whole concept of `php` existing separately from `php@whatever` kills me.

PR #903, opened 2020-01-31
Enabling IPv6 listen

kroko:
Enabling IPv6 listen.  
Nginx on macOS via brew comes built with "--with-ipv6".  
[Dnsmasq options](https://github.com/laravel/valet/blob/master/cli/Valet/DnsMasq.php#L117) do not need to be changed.

### DIFF
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -1,13 +1,13 @@
 server {
     listen 127.0.0.1:80;
-    listen [::1]:80;
+    listen [::]:80;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
 }
 
 server {
     listen 127.0.0.1:443 ssl http2;
-    listen [::1]:443 ssl http2;
+    listen [::]:443 ssl http2;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
@@ -51,7 +51,7 @@ server {
 
 server {
     listen 127.0.0.1:60;
-    listen [::1]:60;
+    listen [::]:60;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
--- a/cli/stubs/secure.valet.conf
+++ b/cli/stubs/secure.valet.conf
@@ -1,11 +1,13 @@
 server {
     listen 127.0.0.1:80;
+    listen [::1]:80;
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
 }
 
 server {
     listen 127.0.0.1:443 ssl http2;
+    listen [::1]:443 ssl http2;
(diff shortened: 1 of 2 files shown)


kroko:
On second thought.

It is unclear though why Nginx listens to localhost explicitly. Obscurity?

Leaving IPv4 to listen explicitly to 127.0.0.1, but setting IPv6 to accept all, as that is the way to make Valet project accessible from WAN via AAAA.

```
    listen 127.0.0.1:<PORT> ...
    listen [::]:<PORT> ...
```

drbyte:
> It is unclear though why Nginx listens to localhost explicitly. Obscurity?

In v2.4.0 via #814 to address the good concerns shared in #772 it was pointed to localhost specifically.

Opening IPv6 to all IPs re-introduces those concerns.

Where's the middle ground?

kroko:
The two reasons mentioned https://github.com/laravel/valet/issues/772#issuecomment-522283653

* Giving localhost only because of possible Nginx vulnerabilities to me sounds unreasonable in praxis. Homebrew supplies mainline Nginx version. Sure, it is technically valid reason that one day Nginx gets exploited and servers all around the world including the one running Valet are exploited.
* What does Nginx listen port has to do with insecure local database engine listening on some other port?

Loss and gain perspective for valet-as-tool usage when choosing localhost only vs anywhere? Does Valet has to babysit DevOps? Does tool such as this has to keep out features so that the uneducated ones (knowledge how to set up firewall should be in primary education curriculum @ 21st century :)) do not get hurt (and maybe learn something along the way), you know what I mean?

As (apparently) I do not consider reasons given in the referenced ticket to be practical/strong I put them into the category of - opinionated decision of *let's use localhost so that everything is in DMZ*. I'm fine with opinionated decisions made by forkable FOSS projects, I'm trully am.

Note that it introduces issues for LAN. Which may be impractical for devteams where multiple people are working on microservices that all have to talk to each other.

IPv6. Same security rules apply. One should use firewall to block IPv6 or ports from outside if needed. However, when needed, it gives the ability to have a service available by a client/remote teammate without NATing and/or pushing to testing/staging env. IMHO (again) it is DevOp responsibility to make his/her computer secure, not Valet's. It is just another discussion on *Security for Internet of Things*, hacking insecured gadgets, cameras and now computers running Valet. Is really having Nginx listening to 80/433 the biggest issue when computer has worldwide accessible unique IPv6 address without any firewall - is it really the entrance point for wrongdoer. SSH/SMB/AFP/VNC...

Now. Assuming that Valet tries to babysit. What about templating and extra valet commands.

```sh
valet global
valet local
```

or

```sh
valet wan
valet local
```

that toggle placeholders in `cli/stubs/secure.valet.conf` and `cli/stubs/valet.conf`

```
listen 127.0.0.1:80; <-> listen 80;
listen 127.0.0.1:443 <-> listen 443;
listen [::1]:80; <-> listen [::]:80;
listen [::1]:443; <-> listen [::]:443;
```

and regenerate files (or edits in place) at `.config/valet/Nginx`.

Valet default install comes with `local enabled`.

PR #913, opened 2020-02-26
Add proxy site-handling commands

drbyte:
Provides helpers to create an Nginx proxy site for the specified host. 
Useful for `docker`, `vessel`, `mailhog` etc, so that both Valet and those other services can co-exist on the same machine. 
(In the case of docker, proxying avoids the problem where Valet and Docker want to share port 80, which is not allowed.)
(In the case of a service like mailhog, it already intends to run on a different port; but proxying allows this to be given a name that Valet recognizes and can serve quickly and easily.)

Adds 3 new commands:
`valet proxy [domain] [host]`  -- generates a proxy config file (see example below)
`valet unproxy [domain]` -- removes a proxy config file
`valet proxies` -- gives a list of proxies Valet is aware of

eg: 
`valet proxy docker http://127.0.0.1:8080` creates a config file for `docker.test`, which proxies all traffic for `https://docker.test` through to `http://127.0.0.1:8080`

eg:
`valet proxy elasticsearch http://127.0.0.1:9200` creates a config file for `elasticsearch.test`, which proxies all traffic for `https://elasticsearch.test` accordingly.

eg:
`valet proxy mailhog http://localhost:8025` will proxy `https://mailhog.test` to the mailhog service running on localhost.

--

Credit for this comes from discussions with @simensen 

--

Note: This implementation always "secures" the site, without needing to run `valet secure` on it.

--

Race condition (or "precedence"): A proxy config takes precedence because you can only have one config for a site. So, even if you have linked or secured a project to `domain` already, and run `valet proxy` with that domain, it will replace the link with the proxy config instead. To return to non-proxy for that domain, `unproxy` and then re-link and/or re-secure as you would normally for a non-proxy project. Or just don't re-use names unless truly important. :)

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -146,6 +146,75 @@ class Site
     }
 
     /**
+     * Get all sites which are proxies (not Links, and contain proxy_pass directive)
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    function proxies()
+    {
+        $dir = VALET_HOME_PATH.'/Nginx/';
+        $tld = $this->config->read()['tld'];
+        $links = $this->links();
+        $certsPath = VALET_HOME_PATH.'/Certificates';
+        $certs = $this->getCertificates($certsPath);
+
+        $proxies = collect($this->files->scandir($dir))
+        ->filter(function ($site, $key) use ($tld) {
+            // keep sites that match our TLD
+            return ends_with($site, '.'.$tld);
+        })->map(function ($site, $key) use ($tld) {
+            // remove the TLD suffix for consistency
+            return str_replace('.'.$tld, '', $site);
+        })->reject(function ($site, $key) use ($links) {
+            return $links->has($site);
+        })->mapWithKeys(function ($site) {
+            $host = $this->getProxyHostForSite($site) ?: '(other)';
+            return [$site => $host];
+        })->reject(function ($host, $site) {
+            // If proxy host is null, it may be just a normal SSL stub, or something else; either way we exclude it from the list
+            return $host === '(other)';
+        })->map(function ($host, $site) use ($certs) {
+            $secured = $certs->has($site);
+            $url = ($secured ? 'https': 'http').'://'.$site;
+
+            return [
+                'site' => $site,
+                'secured' => $secured ? ' X': '',
+                'url' => $url,
+                'path' => $host,
+            ];
+        });
+
+        return $proxies;
+    }
+
+    /**
+     * Identify whether a site is for a proxy by reading the host name from its config file.
+     *
+     * @param string $site Site name without TLD
+     * @param string $configContents Config file contents
+     * @return string|null
+     */
+    function getProxyHostForSite($site, $configContents = null)
+    {
+        $siteConf = $configContents ?: $this->getSiteConfigFileContents($site);
+        $host = null;
+        if (preg_match('~proxy_pass\s+(?<host>https?://.*)\s*;~', $siteConf, $patterns)) {
+            $host = trim($patterns['host']);
+        }
+        return $host;
+    }
+
+    function getSiteConfigFileContents($site, $dir = null)
+    {
+        $config = $this->config->read();
+        $dir = $dir ?: VALET_HOME_PATH.'/Nginx/';
+        $suffix = '.'.$config['tld'];
+        $file = $dir.str_replace($suffix,'',$site).$suffix;
+        return $this->files->get($file);
+    }
+
+    /**
      * Get all certificates from config folder.
      *
      * @param string $path
@@ -265,12 +334,41 @@ class Site
         $secured = $this->secured();
 
         foreach ($secured as $url) {
-            $this->unsecure($url);
-        }
+            $newUrl = str_replace('.'.$oldTld, '.'.$tld, $url);
+            $siteConf = $this->getSiteConfigFileContents($url);
 
-        foreach ($secured as $url) {
-            $this->secure(str_replace('.'.$oldTld, '.'.$tld, $url));
+            if (strpos($siteConf, '# valet stub: proxy.valet.conf') === 0) {
+                $this->unsecure($url);
+                $this->secure($newUrl, $this->replaceOldDomainWithNew($siteConf, '.'.$url, '.'.$newUrl));
+            } else {
+                $this->unsecure($url);
+                $this->secure($newUrl);
+            }
         }
+    }
+
+    /**
+     * Parse Nginx site config file contents to swap old domain to new.
+     * 
+     * @param  string $siteConf Nginx site config content
+     * @param  string $old  Old domain
+     * @param  string $new  New domain
+     * @return string
+     */
+    function replaceOldDomainWithNew($siteConf, $old, $new)
+    {
+        $lookups = [];
+        $lookups[] = '~server_name .*;~';
+        $lookups[] = '~error_log .*;~';
+
+        foreach ($lookups as $lookup) {
+            preg_match($lookup, $siteConf, $matches);
+            foreach ($matches as $match) {
+                $replaced = str_replace($old, $new, $match);
+                $siteConf = str_replace($match, $replaced, $siteConf);
+            }
+        }
+        return $siteConf;
     }
 
     /**
@@ -290,9 +388,10 @@ class Site
      * Secure the given host with TLS.
      *
      * @param  string  $url
+     * @param  string  $siteConf  pregenerated Nginx config file contents
      * @return void
      */
-    function secure($url)
+    function secure($url, $siteConf = null)
     {
         $this->unsecure($url);
 
@@ -305,7 +404,7 @@ class Site
         $this->createCertificate($url);
 
         $this->files->putAsUser(
-            VALET_HOME_PATH.'/Nginx/'.$url, $this->buildSecureNginxServer($url)
+            VALET_HOME_PATH.'/Nginx/'.$url, $this->buildSecureNginxServer($url, $siteConf)
         );
     }
 
@@ -453,16 +552,21 @@ class Site
      * Build the TLS secured Nginx server for the given URL.
      *
      * @param  string  $url
+     * @param  string  $siteConf  (optional) Nginx site config file content
      * @return string
      */
-    function buildSecureNginxServer($url)
+    function buildSecureNginxServer($url, $siteConf = null)
     {
         $path = $this->certificatesPath();
+
+        if ($siteConf === null) {
+            $siteConf = $this->files->get(__DIR__.'/../stubs/secure.valet.conf');
+        }
 
         return str_replace(
             ['VALET_HOME_PATH', 'VALET_SERVER_PATH', 'VALET_STATIC_PREFIX', 'VALET_SITE', 'VALET_CERT', 'VALET_KEY'],
             [VALET_HOME_PATH, VALET_SERVER_PATH, VALET_STATIC_PREFIX, $url, $path.'/'.$url.'.crt', $path.'/'.$url.'.key'],
-            $this->files->get(__DIR__.'/../stubs/secure.valet.conf')
+            $siteConf
         );
     }
 
@@ -520,6 +624,56 @@ class Site
             table(['Site', 'SSL', 'URL', 'Path'], $remaining->toArray());
         }
         info('unsecure --all was successful.');
+    }
+
+    /**
+     * Build the Nginx proxy config for the specified domain
+     *
+     * @param  string  $url The domain name to serve
+     * @param  string  $host The URL to proxy to, eg: http://127.0.0.1:8080
+     * @return string
+     */
+    function proxyCreate($url, $host)
+    {
+        if (!preg_match('~^https?://.*$~', $host)) {
+            throw new \InvalidArgumentException(sprintf('"%s" is not a valid URL', $host));
+        }
+
+        $tld = $this->config->read()['tld'];
+        if (!ends_with($url, '.'.$tld)) {
+            $url .= '.'.$tld;
+        }
+
+        $siteConf = $this->files->get(__DIR__.'/../stubs/proxy.valet.conf');
+
+        $siteConf = str_replace(
+            ['VALET_HOME_PATH', 'VALET_SERVER_PATH', 'VALET_STATIC_PREFIX', 'VALET_SITE', 'VALET_PROXY_HOST'],
+            [VALET_HOME_PATH, VALET_SERVER_PATH, VALET_STATIC_PREFIX, $url, $host],
+            $siteConf
+        );
+
+        $this->secure($url, $siteConf);
+
+        info('Valet will now proxy [https://'.$url.'] traffic to ['.$host.'].');
+    }
+
+    /**
+     * Unsecure the given URL so that it will use HTTP again.
+     *
+     * @param  string  $url
+     * @return void
+     */
+    function proxyDelete($url)
+    {
+        $tld = $this->config->read()['tld'];
+        if (!ends_with($url, '.'.$tld)) {
+            $url .= '.'.$tld;
+        }
+
+        $this->unsecure($url);
+        $this->files->unlink(VALET_HOME_PATH.'/Nginx/'.$url);
+
+        info('Valet will no longer proxy [https://'.$url.'].');
     }
 
     /**
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -68,7 +68,7 @@ class Site
     function host($path)
     {
         foreach ($this->files->scandir($this->sitesPath()) as $link) {
-            if ($resolved = realpath($this->sitesPath().'/'.$link) === $path) {
+            if ($resolved = realpath($this->sitesPath($link)) === $path) {
                 return $link;
             }
         }
@@ -103,7 +103,7 @@ class Site
      */
     function links()
     {
-        $certsPath = VALET_HOME_PATH.'/Certificates';
+        $certsPath = $this->certificatesPath();
 
         $this->files->ensureDirExists($certsPath, user());
 
@@ -119,11 +119,7 @@ class Site
      */
     function parked()
     {
-        $certsPath = VALET_HOME_PATH.'/Certificates';
-
-        $this->files->ensureDirExists($certsPath, user());
-
-        $certs = $this->getCertificates($certsPath);
+        $certs = $this->getCertificates();
 
         $links = $this->getSites($this->sitesPath(), $certs);
 
@@ -152,11 +148,14 @@ class Site
      */
     function proxies()
     {
-        $dir = VALET_HOME_PATH.'/Nginx/';
+        $dir = $this->nginxPath();
         $tld = $this->config->read()['tld'];
         $links = $this->links();
-        $certsPath = VALET_HOME_PATH.'/Certificates';
-        $certs = $this->getCertificates($certsPath);
+        $certs = $this->getCertificates();
+
+        if (! $this->files->exists($dir)) {
+            return collect();
+        }
 
         $proxies = collect($this->files->scandir($dir))
         ->filter(function ($site, $key) use ($tld) {
@@ -173,9 +172,9 @@ class Site
         })->reject(function ($host, $site) {
             // If proxy host is null, it may be just a normal SSL stub, or something else; either way we exclude it from the list
             return $host === '(other)';
-        })->map(function ($host, $site) use ($certs) {
+        })->map(function ($host, $site) use ($certs, $tld) {
             $secured = $certs->has($site);
-            $url = ($secured ? 'https': 'http').'://'.$site;
+            $url = ($secured ? 'https': 'http').'://'.$site.'.'.$tld;
 
             return [
                 'site' => $site,
@@ -205,13 +204,12 @@ class Site
         return $host;
     }
 
-    function getSiteConfigFileContents($site, $dir = null)
+    function getSiteConfigFileContents($site)
     {
         $config = $this->config->read();
-        $dir = $dir ?: VALET_HOME_PATH.'/Nginx/';
         $suffix = '.'.$config['tld'];
-        $file = $dir.str_replace($suffix,'',$site).$suffix;
-        return $this->files->get($file);
+        $file = str_replace($suffix,'',$site).$suffix;
+        return $this->files->get($this->nginxPath($file));
     }
 
     /**
@@ -220,8 +218,12 @@ class Site
      * @param string $path
      * @return \Illuminate\Support\Collection
      */
-    function getCertificates($path)
+    function getCertificates($path = null)
     {
+        $path = $path ?: $this->certificatesPath();
+
+        $this->files->ensureDirExists($path, user());
+
         $config = $this->config->read();
 
         return collect($this->files->scandir($path))->filter(function ($value, $key) {
@@ -265,6 +267,8 @@ class Site
     {
         $config = $this->config->read();
 
+        $this->files->ensureDirExists($path, user());
+
         return collect($this->files->scandir($path))->mapWithKeys(function ($site) use ($path) {
             $sitePath = $path.'/'.$site;
 
@@ -299,7 +303,7 @@ class Site
     {
         $name = $this->getRealSiteName($name);
 
-        if ($this->files->exists($path = $this->sitesPath().'/'.$name)) {
+        if ($this->files->exists($path = $this->sitesPath($name))) {
             $this->files->unlink($path);
         }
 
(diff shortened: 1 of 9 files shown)


alexitgeek:
@drbyte it awesome, it works 😎 👏 

@taylorotwell Please merge that 🤝

mattstauffer:
@alexitgeek I'm the one holding this up. Will try to get to it ASAP :)

fgilio:
Hey guys! Just wanted to confirm a use case for this feature.
If I'm not mistaken, this could be used to have multiple Vessel environments running at the same time and each one with it's own domain. Right?

drbyte:
@fgilio that's the intention, yes

drbyte:
Note: this implementation is geared around one-off individual named hosts, like `mydocker.test` and `docker1.test` and `baz.test`, all being separate proxy sites to their respective destination hosts.

It does NOT support proxying an entire subdomain such as `*.docker.test` into something like Traefik. That will probably require some custom dnsmasq things and some custom proxy stubs, perhaps also skipping the auto-SSL setup in these proxy stubs or maybe invoking some wildcard SSL in such cases.

fgilio:
Thanks for the clarification!

simensen:
I am only just now finally getting to work with this. My bad on the delay.

I found I was able to run:

```
valet proxy
```

... by itself, and it tried to proxy '' which seemed weird to me:

```
Valet will now proxy [https://.test] traffic to [].
Restarting nginx...
```

Fortunately (😄) unproxy without an argument worked, too!

```
valet unproxy
Valet will no longer proxy the [] domain.
```

Unclear if that actually needs to be fixed or not?

Otherwise, the functionality looks GREAT! Love it. Thanks for your work on this, @drbyte!

----

My use case and how I tested this PR:

I have Traefik setup to port 8080. I created two named proxies that point to my Traefik port.

```
valet proxy cds.docker http://127.0.0.1:8080
valet proxy cds-test.docker http://127.0.0.1:8080
```

... and then in my Docker Compose setup, I have my nginx containers setup to declare the same host names as labels:

```
services:
  nginx:
    image: nginx:latest
    networks:
      - internal
      - proxy
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.cds.rule=Host(`cds.docker.test`)"
      - "traefik.docker.network=proxy"

  nginx_test:
    image: nginx:latest
    networks:
      - internal
      - proxy
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.cds-test.rule=Host(`cds-test.docker.test`)"
      - "traefik.docker.network=proxy"
```

Now, I can just visit `cds.docker.test` or `cds-test.docker.test` and Valet's dnsmsaq handles the hostname lookup, Valet's nginx handle the https and http proxying, and Traefik is able to route the request to the right container based on host header!

THANK YOU!

simensen:
Hm, starting to see some other issues, here. Will try to file a report as soon as I can. :)

cli/Valet/Site.php
```diff
@@ -522,6 +622,53 @@ function unsecureAll()
...
+    {
+        $tld = $this->config->read()['tld'];
+        if (!ends_with($url, '.'.$tld)) {
+            $url .= '.'.$tld;
+        }
+
+        $this->unsecure($url);
+        $this->files->unlink(VALET_HOME_PATH.'/Nginx/'.$url);
+
+        info('Valet will no longer proxy the ['.$domain.'] domain.');
```
simensen, line 669:
```
info('Valet will no longer proxy [https://'.$url.'].');
```

cli/Valet/Site.php
```diff
@@ -145,6 +145,71 @@ function parked()
...
+        ->filter(function ($site, $key) use ($tld) {
+            // keep sites that match our TLD
+            return ends_with($site, '.'.$tld);
+        })->map(function ($site, $key) use ($tld) {
+            // remove the TLD suffix for consistency
+            return str_replace('.'.$tld, '', $site);
+        })->reject(function ($site, $key) use ($links) {
+            return $links->has($site);
+        })->reject(function ($site, $key) use ($parked) {
+            return $parked->has($site);
```
simensen, line 172: If I understand this correctly, this means in theory we are not expecting sites that are otherwise parked to be able to be proxied?

Here is my situation:

 * All projects live in ~/Code
 * I have a Docker-based project in ~/Code/app.example.com
 * I expected I'd be able to `valet proxy app.example.com https://127.0.0.1:8443` to allow me to add app.example.com.test

This **actually works**, currently. However, afterwards, when I run `valet proxies` it does not show up in the list.

I see two potential fixes:

 * Remove this check here so that "parked proxies" would show up correctly in the `valet proxies` list (my favorite solution!)
 * Keep this check but block the addition of a proxy if the site is otherwise parked (not my favorite but would understand if that is the way it needs to be)

cli/valet.php
```diff
@@ -181,6 +181,35 @@
         info('The ['.$url.'] site will now serve traffic over HTTP.');
     })->descriptions('Stop serving the given domain over HTTPS and remove the trusted TLS certificate');
 
+    /**
+     * Create an Nginx proxy config for the specified domain
+     */
+    $app->command('proxy [domain] [host]', function ($domain, $host) {
```
simensen, line 187: I think `domain` and `host` should probably be required. Not an expert on Silly, but it looks like dropping the `[]` should do it:

```
$app->command('proxy domain host', function ($domain, $host) {
```

cli/valet.php
```diff
@@ -181,6 +181,35 @@
...
+
+        Site::proxyCreate($domain, $host);
+        Nginx::restart();
+
+    })->descriptions('Create an Nginx proxy site for the specified host. Useful for docker, mailhog etc.');
+
+    /**
+     * Delete an Nginx proxy config
+     */
+    $app->command('unproxy [domain]', function ($domain) {
```
simensen, line 197: I think `domain` should probably be required. Not an expert on Silly, but it looks like dropping the `[]` should do it:

```
$app->command('unproxy domain', function ($domain) {
```

simensen:
I used a code review here. As I was in the middle of it, realized some people hate code reviews on OSS projects. My bad if this was a mistake or ruffles any feathers. Was not my intention. :)

cli/Valet/Site.php
simensen: Oh! Actually, I just did an experiment. If i removed the `parked` check things got super wonky. LOL. So, I removed it and added another check after the map to remove anything that says `(other)`. Kinda janky, and maybe there are better solutions, but trying these changes locally, this seems to work the best for me right now.

```
...
        $proxies = collect($this->files->scandir($dir))
        ->filter(function ($site, $key) use ($tld) {
            // keep sites that match our TLD
            return ends_with($site, '.'.$tld);
        })->map(function ($site, $key) use ($tld) {
            // remove the TLD suffix for consistency
            return str_replace('.'.$tld, '', $site);
        })->reject(function ($site, $key) use ($links) {
            return $links->has($site);
        })->mapWithKeys(function ($site) {
            $host = $this->getProxyHostForSite($site) ?: '(other)';
            return [$site => $host];
        })->reject(function ($host, $site) {
            return $host === '(other)';
        })->map(function ($host, $site) use ($certs) {
            $secured = $certs->has($site);
            $url = ($secured ? 'https': 'http').'://'.$site;
...
```

Maybe this is too heavy-handed and removes some of the flexibility of the "show proxies"? I don't know. It *should* list all of the proxies that are managed by Valet directly, though, so might be a solid enough solution for now.

drbyte:
Thanks @simensen - these feedback points are very helpful. Thanks for all your testing!

cli/Valet/Site.php
drbyte: I think it makes sense. I've added it with some explanatory comments.

drbyte:
Thanks @simensen - I've updated the PR with your suggestions.

cli/Valet/Site.php
```diff
@@ -145,6 +145,77 @@ function parked()
...
+     *
+     * @param string $site Site name without TLD
+     * @param string $configContents Config file contents
+     * @return string|null
+     */
+    function getProxyHostForSite($site, $configContents = null)
+    {
+        $siteConf = $configContents ?: $this->getSiteConfigFileContents($site);
+        $host = null;
+        if (preg_match('~proxy_pass (?<host>[^/]+/.*)~', $siteConf, $patterns)) {
```
simensen, line 204: I've got two conflicting ideas on this.

Here are three commands I ran as I continued to see weird situations come up. Some of this is broken assumptions, of course, but it help me better understand things as I fine tuned the regular expression for various cases.

```
valet proxy foo bar
valet proxy foo2 /bar
valet proxy foo3 b/a/r
```

**First**, the current regular expression seems too strict/specific, and I'm not sure where it came from. It looks like it is saying, "the proxied host needs to not start with a leading `/`, but that it needs to have exactly one `/` at some point later." I don't know why this is a requirement? Could totally be just me not knowing how this might also hurt things, but it seems like we should trap *any* proxy pass host at this point.

```
+---------------------------+-----+-----------------------------------+--------------------------------------------------------------------------+
| Site                      | SSL | URL                               | Host                                                                     |
+---------------------------+-----+-----------------------------------+--------------------------------------------------------------------------+
| app-test.aaaaaaaaaaaa.com |  X  | https://app-test.aaaaaaaaaaaa.com | https://127.0.0.1:8443;                                                  |
| app.aaaaaaaaaaaa.com      |  X  | https://app.aaaaaaaaaaaa.com      | https://127.0.0.1:8443;                                                  |
| foo                       |  X  | https://foo                       | bar;                                                                     |
|                           |     |                                   |         proxy_set_header   Host              $host;                      |
|                           |     |                                   |         proxy_set_header   X-Real-IP         $remote_addr;               |
|                           |     |                                   |         proxy_set_header   X-Forwarded-For   $proxy_add_x_forwarded_for; |
|                           |     |                                   |         proxy_set_header   X-Forwarded-Proto $scheme;                    |
|                           |     |                                   |         proxy_set_header   X-Client-Verify   SUCCESS;                    |
|                           |     |                                   |         proxy_set_header   X-Client-DN       $ssl_client_s_dn;           |
|                           |     |                                   |         proxy_set_header   X-SSL-Subject     $ssl_client_s_dn;           |
|                           |     |                                   |         proxy_set_header   X-SSL-Issuer      $ssl_client_i_dn;           |
|                           |     |                                   |         proxy_set_header   X-NginX-Proxy true;                           |
|                           |     |                                   |         proxy_set_header   Upgrade $http_upgrade;                        |
|                           |     |                                   |         proxy_set_header   Connection "upgrade";                         |
|                           |     |                                   |         proxy_http_version 1.1;                                          |
|                           |     |                                   |         proxy_read_timeout 1800;                                         |
|                           |     |                                   |         proxy_connect_timeout 1800;                                      |
|                           |     |                                   |         chunked_transfer_encoding on;                                    |
|                           |     |                                   |         proxy_redirect off;                                              |
|                           |     |                                   |         proxy_buffering off;                                             |
|                           |     |                                   |     }                                                                    |
|                           |     |                                   |                                                                          |
|                           |     |                                   |     location ~ /\.ht {                                                   |
| foo3                      |  X  | https://foo3                      | b/a/r;                                                                   |
+---------------------------+-----+-----------------------------------+--------------------------------------------------------------------------+
```

As it stands right now, it shows `foo` and `foo3`, but not `foo2`. The value captured for `foo` is not what is expected.

At the very least, I'd suggest we terminate the regular expression by not capturing any whitespace before the trailing `;` which I believe is required to terminate the Nginx config line anyway.

So, my naive solution would be to use the following regex instead: `~proxy_pass\s+(?<host>.*)\s*;~`

By using this, I was able to see all of the proxies I added using the `valet proxy` command and shows the value we might expect:

```
+---------------------------+-----+-----------------------------------+------------------------+
| Site                      | SSL | URL                               | Host                   |
+---------------------------+-----+-----------------------------------+------------------------+
| app-test.aaaaaaaaaaaa.com |  X  | https://app-test.aaaaaaaaaaaa.com | https://127.0.0.1:8443 |
| app.aaaaaaaaaaaa.com      |  X  | https://app.aaaaaaaaaaaa.com      | https://127.0.0.1:8443 |
| foo                       |  X  | https://foo                       | bar                    |
| foo2                      |  X  | https://foo2                      | /bar                   |
| foo3                      |  X  | https://foo3                      | b/a/r                  |
+---------------------------+-----+-----------------------------------+------------------------+
```

**Second**, none of these `foo`, `foo2`, `foo3` proxies actually work! Or rather, they add the configs, but breaks Nginx in the process:

```
$ valet proxy foo bar
Valet will now proxy [https://foo.test] traffic to [bar].

In Nginx.php line 121:

  Nginx cannot start; please check your nginx.conf [1: nginx: [emerg] invalid URL prefix in /User
  s/simensen/.config/valet/Nginx/foo.test:35
  ].


proxy <domain> <host>
```
```
$ valet proxy foo2 /bar
Valet will now proxy [https://foo2.test] traffic to [/bar].

In Nginx.php line 121:

  Nginx cannot start; please check your nginx.conf [1: nginx: [emerg] invalid URL prefix in /User
  s/simensen/.config/valet/Nginx/foo2.test:35
  ].


proxy <domain> <host>
```
```
$ valet proxy foo3 b/a/r
Valet will now proxy [https://foo3.test] traffic to [b/a/r].

In Nginx.php line 121:

  Nginx cannot start; please check your nginx.conf [1: nginx: [emerg] invalid URL prefix in /User
  s/simensen/.config/valet/Nginx/foo3.test:35
  ].


proxy <domain> <host>
```

So, I think my regex is better (`~proxy_pass\s+(?<host>.*)\s*;~`) if only because it catches all of the possibilities for proxies added with `valet proxy domain host` and because it lists them correctly (_and_ without the trailing `;`). However, I don't know if there are other reasons for the original regex since it was quite specific. My use cases could be too simple/naive.

I think this also means we should actually be checking the input for the proxy host value for Nginx to work, anyway, since running `valet proxy domain url` should not break Nginx. If we do that, we could be more specific about the regular expression for proxies to match http(s):// urls. For that, I came up with this solution: `~proxy_pass\s+(?<host>http(?:s|)://.*)\s*;~`

**TL;DR**:

 * Use this regex if we don't want to sanitize input and be able to list all proxies, even if they are broken: `~proxy_pass\s+(?<host>.*)\s*;~` (at leasts makes it easier/possible to see proxies to try and remove them with `valet unproxy`)
 * Use this regex if we plan to sanitize input to make sure the proxy host used looks like a valid URL: `~proxy_pass\s+(?<host>http(?:s|)://.*)\s*;~`

cli/valet.php
```diff
@@ -181,6 +181,35 @@
         info('The ['.$url.'] site will now serve traffic over HTTP.');
     })->descriptions('Stop serving the given domain over HTTPS and remove the trusted TLS certificate');
 
+    /**
+     * Create an Nginx proxy config for the specified domain
+     */
+    $app->command('proxy domain host', function ($domain, $host) {
+
+        Site::proxyCreate($domain, $host);
```
simensen, line 189: I'm not sure where the host-must-be-a-valid-URL-ish-thing validation should happen. Here, or inside `Site::proxyCreate`?

I used the following inside `Site::proxyCreate` and it prevented me from entering hosts like `bar`, `/bar`, and `b/a/r`:

```
if (! preg_match('~^http(?:s|)://.*$~', $host) ) {
    throw new \InvalidArgumentException(sprintf('"%s" is not a valid URL', $host));
}
```

```
$ valet proxy foo bar

In Site.php line 643:

  "bar" is not a valid URL


proxy <domain> <host>
```

cli/Valet/Site.php
drbyte: I'm unable to recall how I arrived at using that regex. I share your reasoning.

One adjustment I'd likely make to your 2nd proposed regex:

```diff
-~proxy_pass\s+(?<host>http(?:s|)://.*)\s*;~
+~proxy_pass\s+(?<host>https?://.*)\s*;~
```
(the `(?:s|)` can just be `s?`, unless I'm misreading your intention)

cli/valet.php
drbyte: Agreed.

cli/Valet/Site.php
drbyte: @simensen I've updated the PR using your feedback. Thanks!

simensen: Yeah, that looks good. :) Thanks!

simensen:
Alright! Pretty excited about this. :) Seems like it is good to go, now! At the very least, I can't seem to break Nginx by running any of these commands with weird inputs! Plus, IT IS SO COOL. Thank you so much for your work on this, @drbyte! I think it's going to help a ton with people who are using both Valet + some other system behind the scenes. Way to go!

@mattstauffer anything else we can do to help make it easier to get this merged?

simensen:
Anything more we can do on this one to help get it merged?

drbyte:
I have nothing more that I feel needs to be added to it at this time.

mattstauffer:
@simensen sorry! I didn't see your tag 16 days ago because it was buried under my weeks of COVID email. Let me take a look now!

mattstauffer:
Oops, original author was @drbyte, so even more apologies to you!

drbyte:
I agree that tests should be added. I'd planned to, and then time passed ... and ... well ... :)

This week I don't have the time to build those. That said, if someone wants to PR to my branch I'll happily review and merge them, and that will become part of this PR.

simensen:
You asked for it, you got it, @drbyte. Please see drbyte/valet#1 for the tests that should hopefully satisfy @mattstauffer's (correct) preference for having tests to fall back on if there are bugs and such.

This was much more invasive than I'd hoped. Took longer than I wanted and but close to what I feared it would take. :-/ Still, I think it is worth it since I *really* want to see this code get merged so I can blog about it. :) Also, I want to think I helped clean up some parts of the `Site` class in the process. :)

drbyte:
Thanks @simensen This is great. I've merged the tests.

drbyte:
And Travis is failing because it's confused about the config file and home directories ...

Argh - I wish this used github actions. 👀

drbyte:
Turns out it's because the tests are indirectly expecting the presence of `config.json`

simensen:
Alright. I'll try to fix this.

drbyte:
Tests are fixed. Thanks @simensen 
Ready for @mattstauffer

mattstauffer:
Re-reading now!

mattstauffer:
Great work y'all!

Wiejeben:
> Note: This implementation always "secures" the site, without needing to run valet secure on it.

Is it possible to create a none-secure proxy?

drbyte:
> > Note: This implementation always "secures" the site, without needing to run valet secure on it.
> 
> Is it possible to create a none-secure proxy?

No. Why would you need to?

mikaelpopowicz:
> > > Note: This implementation always "secures" the site, without needing to run valet secure on it.
> > 
> > 
> > Is it possible to create a none-secure proxy?
> 
> No. Why would you need to?

Hi @drbyte thank you for this.

In case of using `docker-compose`'d containers, we may want have none secure proxy. If containers are isolated in a bridge network you may access container from another by service name or network alias.

docker-compose.yml
```yml
version: 3
services:
    backend:
        image: someLaravelImage
        networks:
            my_network:
                alias: backend.test
    frontend:
        image: someFrontentImage
        networks:
            my_network:
                alias: app.test

networks:
    my_network:
        driver: bridge
```

From frontend container
```bash
curl backend
# works
```

simensen:
@mikaelpopowicz I'm a bit unclear on how Valet proxy would be used in this scenario. Are you saying that from the frontend container, trying `curl https://backend.test` is actually somehow going through Valet proxy to reach *backend.test*? It might just be that I'm not super familiar with how `bridge` works in these cases.

Also, I didn't think that `bridge` worked on macos. Has that changed recently?

In any event, if you can provide more details about what "doesn't work" means for `curl https://backend.test` that would be helpful.

While maybe not ideal in every situation, I've been able to get my non-HTTP applications being served successfully via Valet proxies because my Docker Compose environments are all using Traefik:

```
version: '3.7'
services:
    nginx:
        networks:
            - internal
            - traefik
        labels:
            - traefik.enable=true
            - 'traefik.http.routers.example-portal_nginx.rule=Host(`${DOCKER_ENV_APP_HOST:-example-portal.test}`)'
            - traefik.http.routers.example-portal_nginx.tls=true
            - traefik.docker.network=traefik
```

So, I know it's possible to connect to non-HTTPS applications inside Docker Compose from the host computer. Visiting `https://example-portal.test` works just fine, even though `nginx` is not configured for SSL/TLS/whatever.

mikaelpopowicz:
Hi @simensen sorry for bad explanation.


```yml
version: 3
services:
    backend:
        image: someLaravelImage
        networks:
            my_network:
                alias: backend.test
        ports:
            - 8080:80
    frontend:
        image: someFrontentImage
        networks:
            my_network:
                alias: app.test
        ports:
            - 4000:3000

networks:
    my_network:
        driver: bridge
```

The bridge network is an isolated network. In a bridge network a container can reach another container by service name or a network alias. In this case frontend container can reach backend container using `backend` or `backend.test`.

Using actual proxy : `https://backend.test` to `http://127.0.0.1:8080` and `https://app.test` to `127.0.0.1:4000`.

This works well except in case of SSR requests. Using SSR, frontend container makes api calls directly to the backend container which doesn't expose the port 443.

I didn't know Treafik, i'll checkout it.

simensen:
I see. Yeah, I had to do some "fun stuff" to get things working with Dusk for similar reasons. The `selenium` container needed to be able to reach the `nginx_testing` container via the actual configured hostname. A lot of fun, that was. :)

I didn't know _any_ version of `bridge` worked on macos. Do you have links to how this is set up and configured/used on macos? Because I'd love to look into it more... my most recent Google search on the topic only seemed to confirm for me that it doesn't really work? Entirely possible I'm misunderstanding *everything*.

From my perspective, container-to-container communication shouldn't involve Valet at all. It seems like you'd only need to configure the `frontend` application to send non-HTTPS requests to backend.test, somehow? Using my Dusk example from earlier, I configured my `php_test` container with the URL of the backend application's URL:

```
version: '3.7'
services:
  php_test:
    build: php
    links:
      - db_test
    volumes:
      - project_root:/var/www
    working_dir:
      /var/www
    environment:
      APP_URL: "${DOCKER_ENV_APP_TEST_URL:-https://test.beausimensen.test}"
      USE_SELENIUM: 'true'
      SELENIUM_HOST: selenium
      SELENIUM_PORT: 4444
```


Assuming your `frontend` container has some way to figure out what the URL is for the backend application, you could do something similar, maybe?

```
version: '3'
services:
    frontend:
        image: someFrontentImage
        networks:
            my_network:
                alias: app.test
        ports:
            - 4000:3000
    environment:
        BACKEND_URL: "${DOCKER_ENV_BACKEND_URL:-http://backend}"
```

... in this case, your frontend container should just be able to connect directly to the backend container for SSR without leaving the container network at all. It would just be accessing standard HTTP via port 80 on the `backend` container.

This all assumes, of course, that your frontend app can have the backend URL configured by environment variable.

If for some reason you need the frontend application to be able to tell the rendered application what the URL the client should use for future backend requests (that WILL require accessing via Valet + proxy), you could split the URL into two pieces.

```
version: '3'
services:
    frontend:
        image: someFrontentImage
        networks:
            my_network:
                alias: app.test
        ports:
            - 4000:3000
    environment:
        SSR_BACKEND_URL: "${DOCKER_ENV_SSR_BACKEND_URL:-http://backend}"
        PUBLIC_BACKEND_URL: "${DOCKER_ENV_PUBLIC_BACKEND_URL:-https://backend.test}"
```

So, if the frontend application renders via SSR, it would access the backend via `http://backend` (internal container communication), but the rendered application would broadcast future backend requests at `https://backend.test` (HTTPS, served via Valet + proxy).

I can see how this might be too much work, though, and it's still possible I'm completely missing the point.

All this to avoid adding support for non-HTTPS proxies. :-/ I mean, partly I don't think anyone who worked on this original PR actually *wants* to revisit the work – or workflow – to make HTTPS optional. I know that's true at least for me. Not nice for a consumer that needs it, though. :-/ Hopefully this can help you get unblocked sooner rather than later, tho.

----

RE: Traefik, it's awesome, but might be overkill for you. :) Took me awhile to figure it out and still not 100% sure I have it down right. I have been trying to add a Traefik layer on top of my container configurations to make them optional.

mikaelpopowicz:
First, thank you for your time and advices.

> I didn't know any version of bridge worked on macos. Do you have links to how this is set up and configured/used on macos? Because I'd love to look into it more... my most recent Google search on the topic only seemed to confirm for me that it doesn't really work? Entirely possible I'm misunderstanding everything.

I'm also misunderstanding what you mean 😄 . I maybe give you a bad explanation on how I consider it should work.

> It seems like you'd only need to configure the frontend application to send non-HTTPS requests to backend.test

This may be a workaround for my use case.

> This all assumes, of course, that your frontend app can have the backend URL configured by environment variable.

This is the case.

> All this to avoid adding support for non-HTTPS proxies. :-/

Fortunately you will not have to, take a look at #1005 

Have a good evening or any other day time.

PR #916, opened 2020-03-02
Increase default PHP memory_limit to 512M

### LINKED ISSUE #914: 413 Request Entity Too Large on Database Import
Hi,

Getting an error on import of a database that exceeds 128mb, where can I change this setting in Valet - also is it possible to override per site?

drbyte:
PHP's default is 128M, but it's becoming common practice to set the value to 512M in dev and production
Plus, currently homestead sets it to 512M as well.

(Note: Only takes effect after running `valet install`, or when `valet use php@X.Y` installs a new X.Y PHP version.)

Also closes #914

### DIFF
--- a/cli/stubs/etc-phpfpm-valet.conf
+++ b/cli/stubs/etc-phpfpm-valet.conf
@@ -9,7 +9,7 @@ listen.group = staff
 listen.mode = 0777
 
 ;; When uncommented, the following values will take precedence over settings declared elsewhere
-;php_admin_value[memory_limit] = 128M
+;php_admin_value[memory_limit] = 512M
 ;php_admin_value[upload_max_filesize] = 128M
 ;php_admin_value[post_max_size] = 128M
 
@@ -17,7 +17,7 @@ listen.mode = 0777
 ;php_admin_flag[log_errors] = on
 
 
-
+;; Note: increasing these values will increase the demand on your CPU and RAM resources
 pm = dynamic
 pm.max_children = 5
 pm.start_servers = 2
--- a/cli/stubs/nginx.conf
+++ b/cli/stubs/nginx.conf
@@ -12,7 +12,8 @@ http {
     sendfile on;
     keepalive_timeout  65;
     types_hash_max_size 2048;
-    client_max_body_size 128M;
+
+    client_max_body_size 512M;
 
     server_names_hash_bucket_size 128;
 
--- a/cli/stubs/php-memory-limits.ini
+++ b/cli/stubs/php-memory-limits.ini
@@ -2,7 +2,9 @@
 memory_limit = 512M
 
 ;The maximum size of an uploaded file.
-upload_max_filesize = 128M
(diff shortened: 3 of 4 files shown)


neilgee:
Would you consider also upping the `upload_max_filesize` and  `post_max_size` to 512M also.

```
; Max memory per instance
memory_limit = 512M

;The maximum size of an uploaded file.
upload_max_filesize = 512M

;Sets max size of post data allowed. This setting also affects file upload. To upload large files, this value must be larger than upload_max_filesize
post_max_size = 512M
```

This will allow for larger uploads, my thinking is more for larger databases which a user might upload via a gui database tool like phpMyAdmin.

drbyte:
My only consideration there was consistency with Homestead which only sets it to 100M by default, and I didn't want to reduce Valet's default.

It doesn't matter to me. I rarely see production values set anywhere close to this high, but I realize sometimes we do things in dev that would never be done in production.

ihorvorotnov:
> Would you consider also upping the `upload_max_filesize` and `post_max_size` to 512M also.

For this to work you might also need to increase `client_max_body_size` in Nginx configuration. Now it's capped at 128M.

drbyte:
> Would you consider also upping the `upload_max_filesize` and `post_max_size` to 512M also.

Done

neilgee:
Good stuff

mattstauffer:
Forge sets the memory limit to 512 so I think this is good.

The other values are much higher than Forge... but they already were so I don't think this change should make a huge difference.

sebszocinski:
Is there any way to override the 512 default with an even higher limit? I often run into memory limits installing composer packages so I just set mine to `-1` now so I never have to deal with issues (I don't see the need to set a limit on local development), but every-time I switch php versions using `valet use php@x.x` it resets the memory_limit back to 512mb.

drbyte:
> but every-time I switch php versions using `valet use php@x.x` it resets the memory_limit back to 512mb.

Valet uses the `~/.composer/vendor/laravel/valet/cli/stubs/php-memory-limits.ini` file for building the `/usr/local/etc/php/X.Y/conf.d/php-memory-limits.ini` file for a new PHP version.
So, you could edit that stub file so that the change affects new PHP version installs that Valet manages.

sebszocinski:
> So, you could edit that stub file

Thanks Chris, thats great however this will reset every-time you upgrade Valet, correct?

drbyte:
>  however this will reset every-time you upgrade Valet, correct?

Not necessarily. But if you want a little extra control over that, then first do `composer global require laravel/valet --prefer-source` so that it keeps a git repo in that directory, and then anytime there's an update it'll ask you if you want to replace or edit or stash. Choose `s` to have it temporarily stash your customizations before upgrading and then re-apply the stash back again. It'll tell you if it has a problem restoring the stash and then you can do that manually. 95% of the time I can just use `s` and it's fine.

sebszocinski:
Thanks Chris that's a great tip!!

PR #925, opened 2020-03-24
migrated to use brew prefix.

### LINKED ISSUE #924: HOMEBREW_PREFIX not being honoured.
there are 9 cases in valet where "/usr/local/" is hardcoded in, rather than honouring HOMEBREW_PREFIX

https://github.com/laravel/valet/search?q=%2Fusr%2Flocal&unscoped_q=%2Fusr%2Flocal

This means that any installation of homebrew outside of /usr/local (not so unusual when dealing with multiple toolchains) requires manual edits to the valet codebase.

This is such an easy thing to fix, I don't really know why it hasn't been done.

MrBenGriffin:
I am sorry - I have not been able to test this properly.
However, it seems to be ok in general.

I failed to modify the Brew tests.

Closes #924

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -47,9 +47,7 @@ class Brew
      */
     function prefix() {
         if (!$this->prefix_loaded) {
-            $binary = $this->cli->runCommand('which brew');
-//          while dirname()) is primitive, homebrew strongly suggests using primitive paths to the install.
-//          we use nested dirname because we want to shift off the bin directory.
+            $binary = $this->cli->runCommand('brew --prefix');
             if (strlen($binary) > 0) {
                 $this->homebrew_prefix = dirname(dirname($binary));
             }
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -23,7 +23,7 @@ class Brew
 
     const LATEST_PHP_VERSION = 'php@7.4';
 
-    var $cli, $files;
+    var $cli, $files, $homebrew_prefix, $prefix_loaded;
 
     /**
      * Create a new Brew instance.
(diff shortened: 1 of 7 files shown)


cli/Valet/Brew.php
```diff
@@ -36,6 +36,26 @@ function __construct(CommandLine $cli, Filesystem $files)
     {
         $this->cli = $cli;
         $this->files = $files;
+        $this->homebrew_prefix = '/usr/local';
+        $this->prefix_loaded = false;
+    }
+
+    /**
+     * Determine HOMEBREW_PREFIX
```
drbyte: In your modified environment, does calling `brew --prefix` work properly as a replacement for this custom function?

drbyte:
@MrBenGriffin Thanks for this!

I hope to do some testing on it soon, and take a stab at what needs doing with the Brew tests.

When you have a minute, please edit your initial post above to add `Closes #924` so that the issue automatically closes when this gets merged. It just helps with housekeeping ;)

cli/Valet/Brew.php
MrBenGriffin: OMG. yep.  just use that!! "brew --prefix"

MrBenGriffin: I modified the code to use "brew --prefix" internally. It was a better choice.
Internally brew uses cwd, so getting the path from which brew was more like what it does - but as it has a switch, we should use it.

MrBenGriffin:
Bit confused - not hurrying - but just checking in - am I done on this?  If there's more, let me know.

drbyte:
@MrBenGriffin I think so. I thought I'd have time yesterday to dig, but a work project came up that needed my full attention.

I might get a bit of time to explore it this afternoon.

drbyte:
Observations:

- The Nginx.php `lint()` method needs `$prefix` inserted, else `valet secure` and `valet restart` won't work.

- Several tests will need updating: mostly to mock out the prefix so it can be tested against

mattstauffer:
I believe this has been replaced with a recently merged PR, so closing. Thanks!

PR #947, opened 2020-05-05
Allow Valet to open project URLs using different browsers

aryehraber:
This PR adds a `valet browser` command to set a default browser inside Valet to open project URLs in the browser.

Currently the `valet open` command uses the system default browser. Unfortunately, this renders the command useless in my case since my default browser is Safari, but use Chrome for work. I'd love to use the `valet open` command more but this has been a blocking issue so far.

There are no breaking changes as the default behaviour still works as normal if the config option doesn't exist.

Hopefully others find this feature useful too. Let me know if there's any issues/questions!

Side note: within this PR I added a `Configuration::get()` helper function which uses `Tightenco\Collect\Support\Arr::get()` to ensure no errors are thrown if a key doesn't exist in an array yet + adds the ability to enter a fallback default value. I feel it cleans up the code and reads more expressively compared to current setup:

```php
// using isset()
isset(Configuration::read()['browser']) ? Configuration::read()['browser'] : "None (system default)";

// using data_get()
data_get(Configuration::read(), 'browser', "None (system default)");

// using Configuration::get()
Configuration::get('browser', "None (system default)");
```

Happy to revert to `data_get` if you'd prefer not introduce a new method in `Configuration`.

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -2,6 +2,8 @@
 
 namespace Valet;
 
+use Tightenco\Collect\Support\Arr;
+
 class Configuration
 {
     var $files;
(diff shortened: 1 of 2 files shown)


mattstauffer:
I really appreciate your work on this. However, I've left it sitting for so long just because it's a decent amount of code, which will likely change over time, for a pretty edge case need, and I just can't justify taking the cost on. Sorry, but thanks for your PR!

PR #949, opened 2020-05-06
Stops interfering with favicon and robots for proxies

simensen:
The stub we used for normal usage as it relates to favicon.ico and robots.txt may not be appropriate for proxies. I'm seeing some weird behavior. I'm not even sure what the current configuration *does*, exactly, but if I remove them entirely Nginx forwards all requests to the proxy which is what I expected.

If someone has a better idea I'd be up for it. :) Otherwise, I think this probably fits better with the intent of the proxies.


mattstauffer:
Thanks!

PR #950, opened 2020-05-07
Alias dev-master to 2.x-dev

simensen:
You all seem to be pretty good about updating cutting new releases and updating version numbers quickly, but I thought it would be nice to allow people to target `@dev` more easily (like `^2.10@dev` or something).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,5 +37,10 @@
     },
     "bin": [
         "valet"
-    ]
+    ],
+    "extra": {
+        "branch-alias": {
+            "dev-master": "2.x-dev"
+        }
+    }
 }


drbyte:
Ya, that's fair. And it helps target a version-specific `@dev` series too, instead of generic `@dev-master`.

drbyte:
@simensen Double-checking: merging this doesn't require tagging a new release for this alias to have immediate effect, does it?

simensen:
I am pretty sure as soon as it hits master and packaging picks it up it will be active regardless of any tagging. :)

mattstauffer:
Thanks Beau!!

PR #954, opened 2020-05-26
Make tunnel services configurable

irazasyed:
This will let us configure the various tunnel services online using pattern based matching.

Examples (not included in code):

`valet/config.json`

```json
{
    "tunnel_services": [
        ".serveo.net",
        "-*.localhost.run",
        "-*-*-*-*.sslip.io"
    ]
}
```

This will support [serveo.net](https://serveo.net), [localhost.run](http://localhost.run) & [sslip.io](https://sslip.io).

It also supports nip.io's "dash" notation which wasn't possible before.

Don't have to rely on ngrok limitations since we can use one of these services with a custom subdomain and still be able to connet to valet/project.

### DIFF
--- a/server.php
+++ b/server.php
@@ -53,16 +53,31 @@ function show_directory_listing($valetSitePath, $uri)
  */
 function valet_support_wildcard_dns($domain)
 {
-    if (in_array(substr($domain, -7), ['.xip.io', '.nip.io'])) {
-        // support only ip v4 for now
-        $domainPart = explode('.', $domain);
-        if (count($domainPart) > 6) {
-            $domain = implode('.', array_reverse(array_slice(array_reverse($domainPart), 6)));
-        }
+    $services = [
+        '.*.*.*.*.xip.io',
+        '.*.*.*.*.nip.io',
+        '-*-*-*-*.nip.io',
+    ];
+
+    if (isset($config['tunnel_services'])) {
+        $services = array_merge($services, (array) $config['tunnel_services']);
+    }
+
+    $patterns = [];
+    foreach ($services as $service) {
+        $pattern = preg_quote($service, '#');
+        $pattern = str_replace('\*', '.*', $pattern);
+        $patterns[] = '(.*)' . $pattern;
+    }
+
+    $pattern = implode('|', $patterns);
+
+    if (preg_match('#(?:' . $pattern . ')\z#u', $domain, $matches)) {
+        $domain = array_pop($matches);
     }
 
     if (strpos($domain, ':') !== false) {
-        $domain = explode(':',$domain)[0];
+        $domain = explode(':', $domain)[0];
     }
 
     return $domain;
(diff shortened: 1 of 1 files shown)


irazasyed:
Example usage with localhost.run.

```bash
ssh -R 80:localhost:80 laravel@ssh.localhost.run
```

This should generate a url like: `https://laravel-<random-string>.localhost.run`

If you have this service configured in your valet config and open the tunnel link, it'll be able to process your request and match with the project "laravel" in your parked dir.

irazasyed:
Some more services we can use for reference:
- localtest.me
- localxpose.io
- ssi.sh (self-hosted)

You can also just setup an `A` dns record for your own domain and have wildcard SSL support with services like Cloudflare.

irazasyed:
Some more services we can use for reference:
- localtest.me
- localxpose.io
- ssi.sh (self-hosted)

You can also just setup a wildcard dns record for your own domain and have it a configure with valet to process requests. It's very useful for internal projects :)

drbyte:
I like this. 
Initial thoughts:
- I wonder whether `tunnel_services` is the best name? I've gone back and forth about the word `tunnel`. Undecided.
- Is there a way to do this to make it all even more generic so there's no `config` entry needed?

irazasyed:
@drbyte Thanks for the comment.

1. Feel free to suggest a name but other than tunnel, `expose`, `share` or `wildcard_dns` are some of the names I can think of right now.

2. Please elaborate? Config entry is optional, by default the existing xip.io and nip.io services will work fine. It's just making it flexible so anyone that wishes to use any other services can add an entry and use it. But I think we can definately consider adding all the popular services to the default list if that's what you meant? It's up to the mainteners, tho.

drbyte:
1. Maybe `wildcard_dns_services` makes sense given the name of the function doing the resolving, and because it's allowing addition of additional services for which wildcards will (usually) need to be provided.

2. I was meaning: let's hardcode directly into the function the known services you've identified, 1-per-line like you've done with the xip/nip services already.

I'd like to keep the default config.json as minimalist as possible.
Also, I'd like to avoid encouraging people to edit the file manually unless really necessary ... for two reasons: support, and because any json syntax errors due to manual edits will currently cause Valet to wipe the config file and replace it.

irazasyed:
1. But these are not wildcard DNS services tho.

2. That's doable but the option to customize is still needed as we can't rely on or expect maintainers to constantly keep the list updated plus if there's a custom internal tunnel setup, then obviously we'd want to keep that private for internal use.

You can keep your config file to the default and never touch it if you don't need these services, this is for advanced users that wish to use some of these servives like I said before :)

irazasyed:
Maybe a command can be added to add/remove these services from the config but that's out of scope of this PR I suppose.

irazasyed:
Bump! No maintainers for Valet?

drbyte (approved):

chimit:
@irazasyed, could you, please, add some instructions on how to use it?

I added 
```
"tunnel_services": [
    ".serveo.net",
    "-*.localhost.run",
    "-*-*-*-*.sslip.io"
]
```
into `~/.config/valet/config.json` and restarted Valet. Then I ran `ssh -R 80:localhost:80 my-valet-site@ssh.localhost.run` and got a URL like `https://832b6d7b2c87c4.lhr.life` which I assume points to localhost and not to my Valet project and returns `404 - Not Found`.

If I follow your instructions and call `https://my-valet-site-832b6d7b2c87c4.localhost.run` I get `no tunnel here :(`.

PR #957, opened 2020-06-02
Add 'unpark' alias to 'forget' command

drbyte:
Sometimes I find myself wondering how to un-park a folder, and it takes me awhile to find `forget` in the list.
Adding `unpark` as an alias may be helpful.


mattstauffer:
🙌 I love it!

PR #958, opened 2020-06-02
Protect against un-parking Valet's core Sites folder

### LINKED ISSUE #956: Guard that `valet forget` cannot be run within `~/.config/valet/Sites`
A friend recently accidentally ran `valet forget` from within the `~/.config/valet/Sites` directory and it broke a lot of things. Let's make sure that `valet forget` can only reasonably run within directories that are forgettable, and if a directory is not forgettable, it should just bail out.

drbyte:
Fixes #956

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -180,6 +180,11 @@ class Configuration
      */
     function removePath($path)
     {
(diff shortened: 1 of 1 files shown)


mattstauffer:
🙌

PR #967, opened 2020-06-26
fetch-share-url return list of urls instead of the first

sidux:
Right now when calling the `fetch-share-url` command, it returns on the Url of the first ngrok.
This PR makes it return the urls of all the opened tunnels.

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -2,38 +2,51 @@
 
 namespace Valet;
 
+use Httpful\Exception\ConnectionErrorException;
 use Httpful\Request;
 use DomainException;
 
 class Ngrok
 {
-    var $tunnelsEndpoints = [
-        'http://127.0.0.1:4040/api/tunnels',
-        'http://127.0.0.1:4041/api/tunnels',
-    ];
+    const STARTING_PORT = 4040;
+
+    const TUNNELS_ENDPOINT = 'http://127.0.0.1:{port}/api/tunnels';
 
     /**
      * Get the current tunnel URL from the Ngrok API.
      *
-     * @return string
+     * @return array
      */
-    function currentTunnelUrl($domain = null)
+    function currentTunnelsUrls($domain = null)
     {
         // wait a second for ngrok to start before attempting to find available tunnels
         sleep(1);
 
-        foreach ($this->tunnelsEndpoints as $endpoint) {
-            $response = retry(20, function () use ($endpoint, $domain) {
+        $responses = [];
+        $port = self::STARTING_PORT;
+        $loop = true;
+        while ($loop) {
+            $endpoint = str_replace('{port}', $port, self::TUNNELS_ENDPOINT);
+            try {
+                Request::get($endpoint)->whenError(
+                    static function () use (&$loop) {
+                    $loop = false;
+                })->send();
+            } catch (ConnectionErrorException $e) {
+                break;
+            }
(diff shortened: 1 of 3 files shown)


drbyte:
1. Why?
2. Doesn't this break the `pbcopy` that puts the URL onto the clipboard?

sidux:
1. Let's say you want to open an ngrok an get back the url, this will work fine only when having one ngrok opened, if you have two, you will be able to fetch only the url of the first one.
An alternative solution to this issue, is to return the Url of the last opened ngrok, so when opening and fetching two times you will get each time the correct url.

2. No it's doesn't break it, it's returns multiple lines so all lines will be copied, if you have one opened ngrok it will return 1 line so it's the same as before.

drbyte:
I'm still puzzled ... because last year when I made some updates to fix a bug specifically around supporting multiple tunnels, it was correctly pulling back only the tunnel of the URL for the project that was just started based on the directory being served, even when multiple tunnels were open.

So, I guess I'm still not understanding your use-case ... ie: the actual scenario you're doing that exposed this situation to need a "fix".

ie: if it's already pulling back the right ngrok URL for the tunnel that has just started ... then why are you wanting to get back different URLs, when it's already pulling back the "current" one?

sidux:
My use case is that i had 2 tunnels in the **same** directory, right now the `$domain` defaults to `Site::host(getcwd())` which resolves the issue when you open multiple tunnels in **different** projects/directories.
The possible workaround is two create 2 local domains for the same projet if you want to open two tunnels and get the two different urls from the fetch command by providing the domain param.

drbyte:
> My use case is that i had 2 tunnels in the **same** directory

Thanks. I would have been less puzzled if this were explained at the start ;)

Another question I've been pondering: Is this entire PR primarily intended to get `fetch-share-url` to return an array? Or is it actually to get `pbcopy` to return the multiple URLs? What's your workflow here?  ie: if this were merged, how would you document the "change" in order to explain to every Valet user what's different, why, and how to actually use the multiple values returned? Are you doing some automated parsing of the response somewhere in a 2nd app? Is it just the clipboard contents you're using?
(This question speaks to whether this PR needs changes/updates to improve clarity, or to reduce possible confusion, and what documentation is needed.)

> But still even if you have a different project for each tunnel, you will be able to handle only 2, since you scan only 2 ports (4040, 4041) are checked.

I haven't seen any situation where people are actually using more than 2 simultaneous processes.
Especially since ngrok's free-plan limitations (and first paid tier) limit the number of simultaneous ngrok processes to 1 ... and the higher paid tiers limit it to 2.
Can you show me a case where 3 or more are active?
(This question speaks to whether this PR is overkill because it starts at 1 and just iterates until error, or in terms of maintenance whether it leaves ambiguity in the code about how many tunnels to support in Valet even if ngrok doesn't support them.)

sidux:
The method `Ngrok::currentTunnelsUrls` does return an array, but the `output` function takes an array also so it's just multiple lines instead of one.

So I'd say the PR is intended to get `fetch-share-url` to output the list of all tunnels url's (string separated with "\n")

In my case execute it like this `valet fetch-share-url | tail -n 1` to get the last opened tunnel.
The second case is just the standard pbcopy but I just have two lines in it instead of one one, it lists all the opened tunnels.

So maybe for clarity it will be better to use the name `fetch-share-urls` and deprecate the old command `fetch-share-url`

When i test `valet share` multiple times it does open every time a new tunnel without me having to change anything (tested with 3 tunnels) so I guess ngrok removed the limit.

PR #987, opened 2020-10-28
Add support for PHP 8 & move tests to GitHub Actions

m1guelpf:
This PR adds support for using Valet with PHP 8 and PHP 8.1. It also fixes a deprecation warning triggered when you have optional parameters before required ones, by defaulting the boolean to `false`.


mattstauffer:
Restarting the build after it tanked out after 2.5 hours.

mattstauffer:
Looks like this will probably require upgrading PHPUnit as well.

m1guelpf:
Added the `--ignore-platform-req=php` flag since, as far as I've been able to gather, it's the indicated way to get PHPUnit up and running in PHP 8. This does require us to update to Composer v2, which shouldn't be an issue since lockfiles and stuff are backwards-compatible.

mattstauffer:
Looks like this relies on a polyfill from Symfony that is failing on PHP 5.6 and PHP 7.

image

We may have to drop support for PHP's 5.6 and 7.0 going forward unless anyone else knows the best way around; having not looked into it much, my guess is that Symfony just didn't polyfill whatever this is to older versions.

drbyte:
I'd be totally content dropping PHP 5.6 and 7.0.  But I can't speak for all those using this package.

That said, it's probably possible to configure a matrix of lower dependencies in the `travis.yml` for those older PHP versions.
I don't have the time to build that out for Travis right now, but I've seen lots of packages trying to bridge this gap by doing that.

Aside: I'd also be fine with us switching away from Travis to Github Actions. We could maintain 2 separate test files in that case, one for current and one for older/legacy. Just a thought. (ya, same can be done in Travis ... it's just another dependency instead of GH Actions)

mattstauffer:
I'd love to move to GH actions and I guess I'm also OK if this PR works in 5.6/7.0 in real life but just not Travis. Anyone have easy facility to test it out?

drbyte:
Starting point to work with: https://github.com/m1guelpf/valet/pull/1

m1guelpf:
There we go! Tests are now passing for all PHP versions: https://github.com/m1guelpf/valet/actions/runs/343850543

mattstauffer:
Fantastic! Thanks so much @m1guelpf!

drbyte:
Thanks @m1guelpf !

cli/valet.php
```diff
@@ -165,8 +165,7 @@
     /**
      * Stop serving the given domain over HTTPS and remove the trusted TLS certificate.
      */
-    $app->command('unsecure [domain] [--all]', function ($domain = null, $all) {
-
+    $app->command('unsecure [domain] [--all]', function ($domain = null, $all = false) {
```
drbyte, line 168: It looks like symfony prefers `null` here instead of `false`

I know this was trying to fix:
```
Deprecated: Required parameter $all follows optional parameter $domain 
in ~/.composer/vendor/laravel/valet/cli/valet.php on line 168
```

But using `false` triggers this:
```
Fatal error: Uncaught Symfony\Component\Console\Exception\LogicException: Cannot set a default value when using InputOption::VALUE_NONE mode. in ~/.composer/vendor/laravel/valet/vendor/symfony/console/Input/InputOption.php:159
Stack trace:
#0 ~/.composer/vendor/laravel/valet/vendor/mnapoli/silly/src/Command/Command.php(63): Symfony\Component\Console\Input\InputOption->setDefault(false)
#1 ~/.composer/vendor/laravel/valet/vendor/mnapoli/silly/src/Application.php(108): Silly\Command\Command->defaults(Array)
#2 ~/.composer/vendor/laravel/valet/cli/valet.php(181): Silly\Application->command('unsecure [domai...', Object(Closure))
#3 {main}
  thrown in ~/.composer/vendor/laravel/valet/vendor/symfony/console/Input/InputOption.php on line 159
```

drbyte: Switching to `null` makes both these messages go away.

m1guelpf: Oops, looks like I missed this when running the testsuite somehow 😳

drbyte: It might be a factor of my local environment having mixed PHP versions and mixed composer installs. Perhaps the test suite isn't triggering it because it's actually installing only a single specific compatible PHP version each time.

Feel free to PR the update 👍

m1guelpf: Opened #988

soilSpoon:
> shouldn't

@m1guelpf I think it would be better to upgrade phpunit. Is there any reason why I shouldn't upgrade phpunit?

soilSpoon:
> Added the `--ignore-platform-req=php` flag since, as far as I've been able to gather, it's the indicated way to get PHPUnit up and running in PHP 8. This does require us to update to Composer v2, which shouldn't be an issue since lockfiles and stuff are backwards-compatible.

@m1guelpf I think it would be better to upgrade phpunit. Is there any reason why I shouldn't upgrade phpunit?

drbyte:
> Is there any reason why I shouldn't upgrade phpunit?

Requires a complete syntax rewrite, and breaks backward compatibility.

soilSpoon:
> > Is there any reason why I shouldn't upgrade phpunit?
> 
> Requires a complete syntax rewrite, and breaks backward compatibility.

However, it is currently not supported by php 7.1 or lower. This version below looks good to give up support or just fix bugs and upgrade for the next version. Do you have plans for the next major version?

## Reference
[PHP Supported Versions](https://www.php.net/supported-versions.php)
[PHPUnit Supported Versions](https://phpunit.de/supported-versions.html)
[Laravel Support Policy](https://laravel.com/docs/8.x/releases#support-policy)

PR #992, opened 2020-11-14
Modify brew packages path to be based on brew prefix

soilSpoon:
In the logic to check if php is already installed, /usr/local/bin/php is modified because it does not recognize the php installed by brew on macOS ARM.

## Edit
add other packages path(etc, log)
- nginx
- Dnsmasq

### Reference
https://docs.brew.sh/Installation#untar-anywhere

PR #996, opened 2020-11-23
Point PHP-FPM error_log to user directory

drbyte:
Point PHP-FPM error_log to user directory

Old:  `/usr/local/var/log/php-fpm.log`
New: `~/.config/valet/Log/php-fpm.log`

(We already put `~/.config/valet/Log/nginx-error.log` in this same directory.)

This avoids ARM Mac differences in directory paths (eg: `/usr/local/` isn't a default anymore).
It also makes it easier for the User to find/access these logs.

Addendum to #992 discussion at https://github.com/laravel/valet/pull/992#pullrequestreview-533730869

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -38,7 +38,7 @@ class PhpFpm
             $this->brew->ensureInstalled('php', [], $this->taps);
         }
 
-        $this->files->ensureDirExists('/usr/local/var/log', user());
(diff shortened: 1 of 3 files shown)


drbyte:
Hmmm ... maybe needs some discussion about other log files as well:
https://github.com/laravel/valet/blob/21f0549bd54e4ff4bf186ae8ff504140d38a213b/cli/valet.php#L481-L487

... noting that redirecting those to another place requires adding more configuration that Valet doesn't already manage. Kinda prefer to avoid adding the complexity.

PR #997, opened 2020-11-23
prefer site name over domain name when detecting the site path

gehrisandro:
Prefer path matching the domain name over a path matching only the domain name.

### DIFF
--- a/server.php
+++ b/server.php
@@ -126,6 +126,7 @@ if (strpos($siteName, 'www.') === 0) {
  * Inspects registered path directories, case-sensitive.
  */
 $valetSitePath = null;
+$valetSitePathByDomain = null;
 $domain = array_slice(explode('.', $siteName), -1)[0];
 
 foreach ($valetConfig['paths'] as $path) {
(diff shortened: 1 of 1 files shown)


server.php
```diff
@@ -126,6 +126,7 @@ function valet_default_site_path($config)
  * Inspects registered path directories, case-sensitive.
  */
 $valetSitePath = null;
+$valetSitePathByDomain = null;
```
drbyte, line 129: Can this be done without introducing the additional variable?

gehrisandro: You are right, it is possible.
I changed my PR but I'm not sure if the code is more readable without this extra variable.

drbyte: I understand your point about readability. It's the eternal tension between complexity, readability, cognitive load, and plain old comments! :)

This is one of the few cases where I feel like I prefer using a brief comment instead of a variable. Partly because I didn't feel like the variable was enough to explain why the two "if" clauses can't be combined (and prevent me from re-making that "optimization" again in the future!)

(And, apologies for the breaking change caused by that PR!)

drbyte (approved):

wells:
Pretty please merge.

mattstauffer:
Tagged: https://github.com/laravel/valet/releases/tag/v2.13.2

mattstauffer:
Thanks so much!

PR #1003, opened 2020-11-27
Update brew list command to remove deprecation warning

soilSpoon:
`brew list` does not support TTY. So a raises a warning or error when used with pipe.
This PR solves the above situation.

Ex) 
```
❯ /opt/homebrew/bin/brew list | grep php
Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.
```
```
❯ /usr/local/bin/brew list | grep go
Warning: Calling `brew list` to only list formulae is deprecated! Use `brew list --formula` instead.
go
```

## Reference
[Homebrew/brew#8851](https://github.com/Homebrew/brew/pull/8851)
[Homebrew/brew#8229](https://github.com/Homebrew/brew/pull/8229)

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -48,7 +48,7 @@ class Brew
      */
     function installed($formula)
     {
-        return in_array($formula, explode(PHP_EOL, $this->cli->runAsUser('brew list | grep '.$formula)));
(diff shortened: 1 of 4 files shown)


drbyte (approved):

drbyte:
Thanks for catching this. 

We should probably release this fix quickly as it may be affecting some installs, as mentioned in the linked Valet issue.

PR #1004, opened 2020-11-28
Update test suite to phpunit 9.5

### LINKED ISSUE #1002: Update supported php versions
Currently, php 7.1 or lower is not supported. And PHPUnit is also not supported under 7.1 (currently only PHPUnit8 and PHPUnit9 are supported), and Laravel 6, the lowest supported Laravel, only supports php 7.2.5 or higher.

Using such a supported language or library may lead to security flaws and performance may be poor.

I think it is correct to discard the dead version by changing the major version.

## Reference
[PHP Supported Versions](https://www.php.net/supported-versions.php)
[PHPUnit Supported Versions](https://phpunit.de/supported-versions.html)
[Laravel Support Policy](https://laravel.com/docs/8.x/releases#support-policy)
[#987 comment](https://github.com/laravel/valet/issues/987#issuecomment-734712515)

drbyte:
Update test suite to phpunit 9.5 syntax
Refactored to use polyfill for older PHP versions via `yoast/phpunit-polyfills`

Note: this includes 2 important differences from usual phpunit test suites:
- instead of extending `PHPUnit\Framework\TestCase` we extend `Yoast\PHPUnitPolyfills\TestCases\TestCase`
- instead of handling fixtures via `setUp()` and `tearDown()` we use `set_up()` and `tear_down()` respectively

----

Comment regarding formatting: I chose to use the FQDN in the `extends` syntax of the class declaration instead of using `use` so that it is more quickly apparent that we're doing something slightly different than usual phpunit syntax, particularly in regards to the set_up() / tear_down() methods that appear immediately following the `extends` line.

----

IMO this closes #1002

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,9 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false">
+         stopOnFailure="false"
+         beStrictAboutTestsThatDoNotTestAnything="false"
+>
     <testsuites>
         <testsuite name="Valet Test Suite">
             <directory suffix="Test.php">./tests</directory>
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,11 +29,6 @@ jobs:
 
       - name: Install dependencies
         run: composer install --no-interaction --prefer-dist
-        if: matrix.php != 8.0
-
-      - name: Install dependencies
-        run: composer install --no-interaction --prefer-dist --ignore-platform-req=php
-        if: matrix.php == 8.0
 
       - name: Execute tests
         env:
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vendor/
 composer.lock
 error.log
 .idea
+.phpunit.result.cache
--- a/cli/stubs/etc-phpfpm-error_log.ini
+++ b/cli/stubs/etc-phpfpm-error_log.ini
@@ -1,4 +1,4 @@
-# php-fpm error logging directives
+; php-fpm error logging directives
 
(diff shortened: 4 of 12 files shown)


phpunit.xml
```diff
@@ -7,7 +7,9 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false">
+         stopOnFailure="false"
+         verbose="true"
```
soilSpoon, line 11: verbose is included in the github workflow.
Do you need to add it here?

drbyte: Fair point. Thanks.

drbyte:
@mattstauffer I have no strong opinions whether to use this `yoast` package vs the `symfony` package in # 1009.

I'll note two factors I discovered while exploring both options:

1. The caveat with the symfony package is you can NOT run `phpunit`. You must run `simple-phpunit` instead. That tripped me up a number of times when working locally. This is my primary frustration with it and I wish it could be avoided.

2. The symfony package uses a tiny fraction more RAM to do its tests, and in the background it re-runs composer to clone the phpunit version it is trying to support for a given PHP version. This can make its setup take a tiny bit longer. 
Both are negligible.

soilSpoon:
@drbyte @mattstauffer It has the same drawbacks as above, but symfony project is more mature than yoast.

I'm not sure if maturity or good performance suits the valet project better.

Please good choice

mattstauffer:
Hm.. I very much like what @drbyte wrote about the values of the Yoast package. However, 42 million installs vs 4000 installs is a *big* difference.

That said, I think the PHPUnit Bridge package does more than just the polyfills, and I'm inclined to pull in the smallest surface area we can for this.

I'm leaning toward pulling in Yoast and taking the risk on it, knowing we have Symfony (and #1009) as a backup.

PR #1008, opened 2020-11-29
Improvements to "valet use" command

drbyte:
Improvements to "valet use" command

- `valet use` can now display the currently-linked PHP version, for quick inspection
- properly detects if the requested version is already installed, and skips re-installing/re-starting/re-configuring
    - allows `--force` to re-configure anyway
- smarter treatment of short 'php' name when it's aliased to another specific installed version

All in all, should be faster and a little more user-friendly.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -41,14 +41,23 @@ class Brew
     }
 
     /**
-     * Determine if the given formula is installed.
+     * Ensure the formula exists in the current Homebrew configuration
      *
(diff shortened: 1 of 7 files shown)


drbyte:
Merge note: In my testing, even though tests are touched, merging 1004, 1006, 1007 in order first before this one results in no merge conflicts.

mattstauffer:
I read it, but I'm going to have to take your word that it's good :) Thanks for your hard work on this!!

PR #1018, opened 2020-12-30
Site path found, break from loops

mrk-j:
I don't know exactly since when my issue exists but I believe #997 has something to do with it. If I have 'parked' my `~/PhpstormProjects` directory filled with Laravel projects and have a linked directory inside one of those Laravel projects as a subdomain the parked directory is used instead of the linked directory. This PR prevents that.

To be clear:
```
cd ~/PhpstormProjects
valet park

cd ~/PhpstormProjects/project/subdomain
valet link subdomain.project
```

Then `subdomain.project.test` should link to `~/PhpstormProjects/project/subdomain`, but now it links to `~/PhpstormProjects/project`. After this PR is merged it links to `~/PhpstormProjects/project/subdomain`.

### DIFF
--- a/server.php
+++ b/server.php
@@ -137,10 +137,11 @@ foreach ($valetConfig['paths'] as $path) {
             // match dir for lowercase, because Nginx only tells us lowercase names
             if (strtolower($file) === $siteName) {
                 $valetSitePath = $path.'/'.$file;
-                break;
+                break 2;
             }
             if (strtolower($file) === $domain) {
                 $valetSitePath = $path.'/'.$file;
+                break 2;
             }
         }
         closedir($handle);
--- a/server.php
+++ b/server.php
@@ -137,14 +137,18 @@ foreach ($valetConfig['paths'] as $path) {
             // match dir for lowercase, because Nginx only tells us lowercase names
             if (strtolower($file) === $siteName) {
                 $valetSitePath = $path.'/'.$file;
-                break 2;
+                break;
             }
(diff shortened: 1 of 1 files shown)


server.php
```diff
@@ -137,10 +137,11 @@ function valet_default_site_path($config)
...
                 $valetSitePath = $path.'/'.$file;
-                break;
+                break 2;
             }
             if (strtolower($file) === $domain) {
                 $valetSitePath = $path.'/'.$file;
+                break 2;
             }
         }
         closedir($handle);
```
drbyte, line 147: While `break 2` is simple and expressive, that means the `closedir()` doesn't get called.
I haven't checked lately: does PHP always properly close all opened file handles at end of execution?

mrk-j: You're right, I updated the code to break from the deepest loop when a match is found and break from the outside loop when `$valetSitePath` is not `null` anymore.

mrk-j:
In some cases my updated PR seems to break, I need to check this later.

mrk-j:
It seems to work after all, I believe it works as it should now.

drbyte (approved):
I'm satisfied with this.

drbyte:
I've been using this code in my own environment for a couple weeks, without any negative issues.

@mattstauffer I think this can be merged.

mrk-j:
I’ve been using it too, and I have not encountered any side-effects.

mattstauffer:
Thanks @mrk-j!

PR #1025, opened 2021-01-18
Fix valet bin dir linking for ARM / M1 macs

### LINKED ISSUE #1023: Incorrect bin directory for M1 Macs
## Clear description of your problem

Homebrew added support for M1 Macs. As part of that, they're now using `/opt/homebrew` rather than `/usr/local`.

Valet still uses the old path in its code.

This causes a couple of issues:
- The `valet` binary doesn't get symlinked into the user's `bin`
- Valet constantly asks for your password because the path in the sudoers file that `valet trust` generates points to the wrong place.

## Expected behavior

- The path to the correct `bin` directory should be used.
- The valet binary should get symlinked to the correct bin directory.
- The valet trust command should write the correct path and stop asking for passwords.

## Steps to Reproduce
1. Fresh install of MacOS on an M1 mac.
2. Install Homebrew
3. Install Valet
4. Run `valet trust` 
5. Open a new terminal so your sudo doesn't carry over from a second ago.
6. Run a valet command, see that it asks for your password.
7. Symlink `/opt/homebrew/bin/valet`
8. Edit `/etc/sudoers.d/valet` to point to `/opt/homebrew/bin/valet`
9. New terminal again.
10. Run a valet command, but this time it won't ask for a password.

## Possible solution

The path is hardcoded here:
https://github.com/laravel/valet/blob/35676120a707feacaed5bea5ce4311f73c2dcc85/cli/Valet/Valet.php#L11

This'll probably need to be injected.

I imagine it'll need to be a bit more than a "is this an M1, use /opt/homebrew, otherwise /usr/local" because of people that installed Homebrew before the native M1 support, which still would have used /usr/local.

Maybe you can check if the `/opt/homebrew` directory exists?

There's a few other references to `/usr/local`. https://github.com/laravel/valet/search?q=%2Fusr%2Flocal

I'm happy to take a stab at a PR when I get time, but I don't know when that'll be. Maybe this will be a simple thing for someone who knows Valet's internals better.

Hope that was helpful, thanks!

drbyte:
Fixes #1023

Ref #992

/cc @jasonvarga

### DIFF
--- a/cli/Valet/Valet.php
+++ b/cli/Valet/Valet.php
@@ -8,7 +8,7 @@ class Valet
 {
     var $cli, $files;
 
-    var $valetBin = '/usr/local/bin/valet';
+    var $valetBin = BREW_PREFIX.'/bin/valet';
 
     /**
      * Create a new Valet instance.
@@ -88,7 +88,7 @@ class Valet
     {
         $this->files->ensureDirExists('/etc/sudoers.d');
 
-        $this->files->put('/etc/sudoers.d/valet', 'Cmnd_Alias VALET = /usr/local/bin/valet *
+        $this->files->put('/etc/sudoers.d/valet', 'Cmnd_Alias VALET = '.BREW_PREFIX.'/bin/valet *
 %admin ALL=(root) NOPASSWD:SETENV: VALET'.PHP_EOL);
     }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):

jasonvarga:
Worked well on my M1 here, thanks!

drbyte:
> Worked well on my M1 here, thanks!

Great. Thanks for testing it. 

/cc @mattstauffer Ready for merge.

mattstauffer:
<3

mattstauffer:
Tagged in v2.13.17

Issue #790, opened 2019-06-26
Errors when using "valet share"


AnthonyWebDev:
Hi guys,

As i always do i wanted to share the application I am making to a customer so i did a "valet share" but got some errors.
Here are the logs : https://pastebin.com/3GjCaZAD
Do you have any idea ? 
Already updated composer and tried to uninstall / re install valet.
Thanks for the help.

drbyte:
> `ERROR:  Tunnel 'command_line' specifies invalid address 'newlinedisplay.test:80': address newlinedisplay.test:80: unexpected '[' in address`

1. For diagnostic purposes, make the following edit to find out the command that's actually being run.
Temporarily edit `~/.composer/vendor/laravel/valet/valet`, line 55 to change `sudo -u` to just `echo`:
```diff
-  sudo -u "$(logname)" "$DIR/bin/ngrok" http "$HOST.$TLD:$PORT" -host-header=rewrite ${*:2}
+  echo "$(logname)" "$DIR/bin/ngrok" http "$HOST.$TLD:$PORT" -host-header=rewrite ${*:2}
```
Then run `valet share` again (from your `newlinedisplay` dir of course).  It will output the info it resolves for that line, and it will also output a Connection Refused message, which you can ignore and press CTRL+C to get control back.

What's the output you get?
It should be something like this:
```text
username /Users/username/.composer/vendor/laravel/valet/bin/ngrok http projectname.test:80 -host-header=rewrite
```

2. The error message you posted mentions `unexpected '[' in address`. Where's the `[` coming from?

3. Have you let Valet manage all your nginx configs? Or have you edited those files yourself? (`~/.config/valet/Nginx/newlinedisplay.test`)

AnthonyWebDev:
Hi @drbyte as always you rock the issues 👍 
1 / Here is the result : 
<img width="796" alt="Capture d’écran 2019-06-26 à 19 27 52" src="https://user-images.githubusercontent.com/45395092/60201443-92d4d600-9848-11e9-8f71-03e475a7c61e.png">

2/ I have no idea ...

3/ I didn't edit anything by myself. I also uninstalled and re installed completely homebrew and valet (and let nginx / dnsmasq etc ... being installed by valet)

drbyte:
Hmmm...
Well, let's undo that edit at least.

I'm puzzled about where ngrok is getting the `[` from.

I'd try securing it since that creates a new nginx config:
```sh
cd newlinedisplay
valet secure
valet share
```
[...]

PR #1029, opened 2021-01-22
Fix issue with `valet share` on Hyper

drbyte:
Related: https://github.com/laravel/valet/issues/790#issuecomment-765559963

Changing from `info()` to just `output()` avoids ANSI char output for color display (which Vercel Hyper is misinterpreting), and this particular line doesn't "need" colored output particularly.


mattstauffer:
Wow. You're awesome, and.. what a weird error.

PR #1042, opened 2021-03-12
Add some self-repair suggestions to the Issues template

drbyte:
Offers some suggestions for self-repair before posting an issue.


mattstauffer:
Great ideas!!

PR #1050, opened 2021-03-31
Small changes to allow Valet to be used in non-CLI contexts

mattstauffer:


cli/includes/helpers.php
```diff
@@ -4,21 +4,22 @@
...
 
 /**
- * Define the ~/.config/valet path as a constant.
+ * Define constants
  */
-define('VALET_LOOPBACK', '127.0.0.1');
-define('VALET_HOME_PATH', $_SERVER['HOME'].'/.config/valet');
-define('VALET_SERVER_PATH', realpath(__DIR__ . '/../../server.php'));
-define('VALET_STATIC_PREFIX', '41c270e4-5535-4daa-b23e-c269744c2f45');
+if (! defined('VALET_LOOPBACK')) {
```
mattstauffer, line 13: I *hope* just checking for a single constant is enough to know whether this has been booted up (right now, the goal of this is *only* to make Valet work as a dependency of a Laravel/PHP app that's running *served by Valet*). But I'm open if you think I should do individual checks, @drbyte

cli/includes/helpers.php
```diff
@@ -4,21 +4,22 @@
...
+    define('VALET_LOOPBACK', '127.0.0.1');
+    define('VALET_HOME_PATH', $_SERVER['HOME'] . '/.config/valet');
+    define('VALET_SERVER_PATH', realpath(__DIR__ . '/../../server.php'));
+    define('VALET_STATIC_PREFIX', '41c270e4-5535-4daa-b23e-c269744c2f45');
 
-define('VALET_LEGACY_HOME_PATH', $_SERVER['HOME'].'/.valet');
+    define('VALET_LEGACY_HOME_PATH', $_SERVER['HOME'] . '/.valet');
```
mattstauffer, line 19: As I mentioned in a chat, I'm also going to be moving the codebase to follow Tlint and PSR-12 styles.

mattstauffer: Well, PSR-12 for sure. Not 100% sure about Tlint; the only one I'm doing for sure is spaces around the concatenation operator.

mattstauffer:
@drbyte It won't let me assign you as the reviewer, but would you give me a sanity check? Thanks!

cli/includes/helpers.php
drbyte: I think this is fine for now. 
At some point extracting a "valet_is_booted()" (or similar) function may have benefit. But that may become clearer at a later point.

drbyte:
👍

cli/includes/helpers.php
mattstauffer: 👌

mattstauffer:
Thanks!

cli/includes/helpers.php
mattstauffer: haha, sweet :)

cli/includes/helpers.php
gbuckingham89: I'm a bit late to the party here, but I'm not sure `VALET_LOOPBACK` is the right constant to be checking for. 

If like Matt, I require the `laravel/valet` package into another Laravel app _(I understand it's not really a supported use case)_ that is being served by Valet, I still see warnings about `VALET_HOME_PATH` and `VALET_STATIC_PREFIX` already being defined.

I'm relatively new to Valet, but diving in, it looks like when a site is served through `/server.php`, the only two constants that get set are `VALET_HOME_PATH` and `VALET_STATIC_PREFIX` - not `VALET_LOOPBACK`.

Would you open to a PR that changes this so we check for `VALET_HOME_PATH` instead of `VALET_LOOPBACK`?

PR #1059, opened 2021-05-15
Add Nette framework driver

Majkie:


cli/drivers/NetteValetDriver.php
```diff
@@ -0,0 +1,65 @@
...
+        return $sitePath.'/www/index.php';
+    }
+
+    /**
+     * Check if .neon config files are present.
+     *
+     * @param $sitePath
+     * @return bool
+     */
+    private function checkNanoConfigFiles($sitePath)
```
rodrigopedra, line 57: Shouldn't this be named `checkNeonConfigFiles`?

PR #1062, opened 2021-05-24
Add `-diff` for binary file.

crynobone:
https://git-scm.com/docs/gitattributes#_marking_files_as_binary

> Git usually guesses correctly whether a blob contains text or binary data by examining the beginning of the contents. However, sometimes you may want to override its decision, either because a blob contains binary data later in the file, or because the content, while technically composed of text characters, is opaque to a human reader.


drbyte (approved):

mattstauffer:
Thanks @crynobone!

PR #1040, opened 2021-02-23
Fix subdomains pointing to root path

lpheller:
Closes #1032

This PR reverts the changes from #1018 and fixes the problem with subdomains described in #1032.

The "breaking" change was in [server.php#L144](https://github.com/laravel/valet/blob/master/server.php#L144)
```php
            if (strtolower($file) === $domain) {
                $valetSitePath = $path.'/'.$file;
                break;
            }
```

The `break` statement stopped the execution of the loop too early, as `$domain` is not what we are looking for at this very moment, when using a linked subdomain.

I actually was able to reproduce the problem @mrk-j described in #1018 too and solved this by re-adding his outer break statement, but removing the early breaks. 

So, this should fix both problems described in #1018 and #1032.

Issue #1032, opened 2021-02-06
Subdomain is pointing to root domain, even though they have different folders linked.


natecorkish:
## Clear description of your problem

I have a root folder called Onlysort, which contains 2 sub-folders: app and homepage. The app is a Laravel installation and the homepage is a jigsaw installation. I have the homepage linked to `Onlysort` and the app linked to `app.onlysort` now, when I visit both urls, it takes me to the homepage link (Onlysort.test)

<img src="https://i.imgur.com/J6GSYLH.png">

## Expected behavior

app.onlysort.test to visit the app folder
onlysort.test to visit the homepage folder

## Diagnosis

`<details>
<summary>sw_vers</summary>
<pre>ProductName:	macOS
ProductVersion:	11.1
BuildVersion:	20C69</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 2.13.19</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "paths": [
        "/Users/nate/.config/valet/Sites"
    ]
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "laravel/valet": "^2.13"
    }
}</pre>
</details>
<details>
<summary>composer global diagnose</summary>
<pre>Changed current directory to /Users/nate/.composer
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.0.7 => 2.0.9)
Composer version: 2.0.7
PHP version: 8.0.0
PHP binary path: /usr/local/Cellar/php/8.0.0_1/bin/php
OpenSSL version: OpenSSL 1.1.1h  22 Sep 2020
cURL version: 7.74.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1i
zip extension: OK</pre>
</details>
<details>
<summary>composer global outdated</summary>
<pre>Changed current directory to /Users/nate/.composer
[...]

PR #1068, opened 2021-06-10
Fix subdomain using root domain link

stidges:
I just updated my local Valet installation, and it looks like the fix implemented in #1040 does not work for me. After putting some debug `var_dump`s in the `server.php` file in the `get_valet_site_path` function - in the loop and right before the return statement - it gave me the following output when trying to access `a.domain.test`:

```
// -> /Users/stidges/.config/valet/Sites/a.domain
// -> /Users/stidges/.config/valet/Sites/domain

// -> RESULT: /Users/stidges/.config/valet/Sites/domain
```

I think this is caused by the root domain coming _after_ the subdomain, so it overwrites the `$valetSitePath`. 

This PR breaks out of the loop as soon as the `strtolower($file) === $siteName` statement passes, as you're sure to have an exact match at that point. This would still allow for a fallback to the root domain.

### DIFF
--- a/server.php
+++ b/server.php
@@ -139,7 +139,7 @@ function get_valet_site_path($valetConfig, $siteName, $domain)
         $dirs = [];
 
         while (false !== ($file = readdir($handle))) {
-            if (is_dir($path.'/'.$file) && ! in_array($file, ['.', '..', '.DS_Store'])) {
+            if (is_dir($path.'/'.$file) && ! in_array($file, ['.', '..'])) {
                 $dirs[] = $file;
             }
         }
--- a/server.php
+++ b/server.php
@@ -135,7 +135,12 @@ function get_valet_site_path($valetConfig, $siteName, $domain)
                 if (in_array($file, ['.', '..', '.DS_Store'])) continue;
 
                 // match dir for lowercase, because Nginx only tells us lowercase names
-                if (strtolower($file) === $siteName || strtolower($file) === $domain) {
+                if (strtolower($file) === $siteName) {
+                    $valetSitePath = $path.'/'.$file;
+                    break;
+                }
+
+                if (strtolower($file) === $domain) {
                     $valetSitePath = $path.'/'.$file;
                 }
             }
--- a/server.php
+++ b/server.php
@@ -128,32 +128,39 @@ if (strpos($siteName, 'www.') === 0) {
 function get_valet_site_path($valetConfig, $siteName, $domain)
 {
     $valetSitePath = null;
+
     foreach ($valetConfig['paths'] as $path) {
-        if ($handle = opendir($path)) {
-            $dirs = [];
-            while (false !== ($file = readdir($handle))) {
-                if (! is_dir($path.'/'.$file)) continue;
-                if (in_array($file, ['.', '..', '.DS_Store'])) continue;
+        $handle = opendir($path);
+
+        if ($handle === false) {
+            continue;
+        }
+
+        $dirs = [];
+
+        while (false !== ($file = readdir($handle))) {
+            if (is_dir($path.'/'.$file) && ! in_array($file, ['.', '..', '.DS_Store'])) {
                 $dirs[] = $file;
             }
-            closedir($handle);
+        }
 
-            // Note: strtolower used below because Nginx only tells us lowercase names
-            foreach ($dirs as $dir) {
-                if (strtolower($dir) === $siteName) {
-                    // early return when exact match for linked subdomain
-                    return $path.'/'.$dir;
-                }
+        closedir($handle);
 
-                if (strtolower($dir) === $domain) {
-                    // no early return here because the foreach may still have some subdomains to process with higher priority
-                    $valetSitePath = $path.'/'.$dir;
-                }
+        // Note: strtolower used below because Nginx only tells us lowercase names
+        foreach ($dirs as $dir) {
+            if (strtolower($dir) === $siteName) {
+                // early return when exact match for linked subdomain
+                return $path.'/'.$dir;
             }
 
-            if ($valetSitePath) {
-                return $valetSitePath;
+            if (strtolower($dir) === $domain) {
+                // no early return here because the foreach may still have some subdomains to process with higher priority
+                $valetSitePath = $path.'/'.$dir;
             }
+        }
+
+        if ($valetSitePath) {
+            return $valetSitePath;
         }
     }
 }


lpheller:
Hi @stidges,
I can't reproduce the behavior you describe. Could you give some details about the setup you are using? Which domains / sites are parked, which are linked etc.? 
My setup still works with your code, so everything seems ok. 🙏🏻

mattstauffer:
OK, I tried the workflow above in older Valet and it works as expected on `2.15.0` (`b.domain.test` returns "Two!") after I linked it in the `two` directory.

Updating Valet to the `2.15.1`..

`b.domain.test` now shows "One" even though it's linked to the "Two" directory.
 
So I *can* reproduce this. Updating to this branch to see if it fixes it...
 
Yep. Once I apply these changes it fixes. 

I'll wait on @drbyte review, because I didn't dig into #2040 at all so I don't know what implications this has, but I can verify I was able to show that @stidges use case worked in 2.15.0, broke in 2.15.1, and worked again with this change.

lpheller:
Yes, I can see the error in my code now. I see I only fixed a very specific part and broke it for other scenarios.

@mattstaufer let me give you a short recap how we got here:
Prior to #1040 a subdomain linked inside of a folder would resolve in the base domain linked to a separate folder, because in #1018 the loop was broken as soon as a matching domain was found.

In #1040, I reverted this change and tried to find a fix to the problem described in #1018, but unwittingly reversed the change introduced in #997 (which was the PR that lead to the problem in #1018).

So with this PR's combined, I think we finally got everything covered, as we are basically back to #997 but still include this  fallback prevent the original issue from #1018 (and the problem from #1032 is still fixed by this PR as well):
https://github.com/laravel/valet/blob/43da6f5065b0ca168d5222818a26178861a85b03/server.php#L149-L151

<details>

<summary>
Tested scenarios
</summary>
I've tested the following scenarios from this PR.

| Site           | URL                         | Path                            |
|----------------|-----------------------------|---------------------------------|
| a.domain       | http://a.domain.test        | ~/Sites/one            | 
| b.domain       |http://b.domain.test        | ~/Sites/two            |
| domain         |http://domain.test          | ~/Sites/one            |


As well as the scenario described in #1032
| Site           | URL                         | Path                            |
|----------------|-----------------------------|---------------------------------|
| app.project     |http://app.project.test      | ~/Sites/project/app    |
| project     |http://project.test      | ~/Sites/project/main    |


</details>

Regarding the code change in this PR, I'd suggest to use a return statement instead of a break, since this code now lives in a function, and this would make it even more obvious that we have found what we are looking for. But that doesn't change the functionality at all.

```php
if (strtolower($file) === $siteName) {
     return $valetSitePath = $path.'/'.$file;
}

**Comment by mattstauffer** [comment]
@drbyte LMK when you have a chance to review.

@stidges Could you make the `return` change suggested above, if you agree with it?

**Comment by drbyte** [comment]
While I like the "early `return`" suggestion (especially because, as mentioned, it is now inside a function, making the `break` less necessary in most cases), it has one problem: we still have a `readdir()` file handle open that ought to be closed for thoroughness. (I suspect the PHP internals handles cleanup properly at the end of execution, but it's good to be complete anyway.)

**Comment by drbyte** [comment]
@lpheller @stidges shouldn't the 2nd `if (strtolower($file) ===` also have a `break` added to it?

Perhaps it's moot, but it would save a few CPU cycles.

**Comment by lpheller** [comment]
@drbyte adding it for the second one as well would probably reopen this whole vicious circle again, since a similar thing was added in #1018 https://github.com/laravel/valet/pull/1018/files#diff-a35003f202d86e283e3a2378727b7094fff8c81c768368f5acc07258cd56c0e0R144 and caused other bugs like described in #1032.

**Comment by drbyte** [comment]
Yup, you're right. I was just testing a few scenarios and see that. It needs to be allowed to match all possible subdomains first.

**Comment by drbyte** [comment]
@lpheller I've made a PR to @stidges branch. It seems to test fine for me. Wanna check it in your scenarios?

https://github.com/stidges/valet/pull/1/files

It closes the open dir handles earlier by storing the available entries (with `$dir` which is more intuitive/accurate than `$file`), then does the foreach using that array, with the early return on a matching subdomain, but leaving the top tld for final fallback. And adds some comments to help us avoid getting into this cycle again.

**Comment by lpheller** [comment]
@drbyte Looks great! I tested it in all the scenarios I set up and could not find any problems. And the readability is also much better with these changes  🙏🏻

**Comment by drbyte** [comment]
@stidges when you merge the PR I made to your branch, those changes should automatically appear in this PR, and then @mattstauffer can complete the final merge into Valet.

**Comment by lpheller** [comment]
>and then mattstauffer can complete the final merge into Valet. 

Hopefully once and for all 😅

Thank you @drbyte and @mattstauffer for your ongoing work on this great tool!

**Comment by mattstauffer** [comment]
Freaking lovely. Thanks everyone!

**Comment by stidges** [comment]
@drbyte @lpheller Thanks for looking into this! I have one more suggestion for the function so that it is a bit easier to follow - I'll push that right now so you can take a look to see what you think :)

File: server.php [hunk 1]
```diff
@@ -128,23 +128,40 @@ function valet_default_site_path($config)
...
+        $handle = opendir($path);
+
+        if ($handle === false) {
+            continue;
+        }
+
+        $dirs = [];
+
+        while (false !== ($file = readdir($handle))) {
+            if (is_dir($path.'/'.$file) && ! in_array($file, ['.', '..', '.DS_STORE'])) {
```
drbyte, line 142: `.DS_STORE` should go back to `. DS_Store`

stidges: It looks like the current master version of server.php says `.DS_STORE` without the space, is this incorrect?

stidges: Excuse me, I misunderstood :) fixed it

drbyte: Correct. My bad with sloppy copy/paste 🤣

lpheller: I suspect the space is a typo, but it probably shouldn't all be capitalized.

drbyte: Looks properly corrected now. 👍

stidges: Looking into this further, I wonder if we need this check at all, since the `is_dir` check comes first, which returns `false` for `.DS_Store`?

drbyte: True: `.DS_Store` is a file, not a directory.

drbyte: If you're gonna remove it, do you mind rolling back your local working branch to the Flattening commit, before removing it? And then force-push the commit to the PR, to keep the history cleaner?

stidges: Sure thing! Squashed them together :)

drbyte (approved):

mattstauffer:
Thanks again y'all!

PR #1079, opened 2021-07-29
Prevent 502 errors when using AJAX

mattkingshott:
When using the Laravel DebugBar with the `capture_ajax` flag set to `true`, developers may encounter 502 Bad Gateway errors. This is due to DebugBar's AJAX request monitoring exceeding the FastCGI buffer limit.

This PR increases the size of the buffer. While it is not guaranteed to fix the problem, the increased buffer size should solve the issue for most developers. Further discussion on this subject can be [found in this issue on the DebugBar repo](https://github.com/barryvdh/laravel-debugbar/issues/251).

### DIFF
--- a/cli/stubs/fastcgi_params
+++ b/cli/stubs/fastcgi_params
@@ -18,3 +18,5 @@ fastcgi_param SERVER_NAME  $server_name;
 fastcgi_param HTTPS   $https if_not_empty;
 fastcgi_param REDIRECT_STATUS  200;
 fastcgi_param HTTP_PROXY  "";
+fastcgi_buffer_size 512k;
+fastcgi_buffers 16 512k;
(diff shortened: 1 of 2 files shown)


drbyte:
Is it intentional that this only applies to unsecured sites? ie: only on sites for which `valet secure` has NOT been run?

The change you're proposing is in the config file used only for unsecured sites. 
If one wishes this to be applied to both secured and unsecured sites, note that both types of site stubs have an `include fastcgi_params` directive, meaning the proposed changes could be placed in that file instead.

Also, is the temp_file_write_size and busy_buffers part really needed?  Forcing unneeded settings on every Valet user can lead to unexpected outcomes.

mattkingshott:
Thanks for the comments @drbyte. It wasn't intentional no, so I've done the following...

I've moved the required parameters to `fastcgi_params` and removed the unnecessary ones.

drbyte:
Wondering: How does doing this affect RAM demands on sites?  eg: what are the new requirements for using Valet?

mattkingshott:
I’ve no idea to be honest, but is it likely to be an issue given that Valet is a dev environment and not for use in production?

Either way, I’ve not noticed any difference with it enabled or not.

mattstauffer:
There are a lot of people in [this thread](https://github.com/barryvdh/laravel-debugbar/issues/2510) either talking about this or emoji responding positively to the folks talking about it, so I'm going to consider it tested enough, and merge it. Thanks!

PR #1082, opened 2021-08-06
Update ngrok binary to latest version v2.3.40

devfrey:
Today when trying valet share the following error popped up:
> $ valet share --region=eu
Your ngrok agent version "2.3.35" is no longer supported. Only the most recent version of the ngrok agent is supported without an account.
Update to a newer version with `ngrok update` or by downloading from https://ngrok.com/download.
Sign up for an account to avoid forced version upgrades: https://ngrok.com/signup.
ERR_NGROK_120

This PR updates the ngrok binary to the latest stable version.

Maybe we should also look into supporting the M1 binary for non-Intel macs. But that's outside of the scope of this PR.


drbyte:
I'm puzzled why your ngrok isn't auto-updating itself.

Mine does this:
Screen Shot 2021-08-06 at 1 18 54 PM
and then about 3 seconds later this:
Screen Shot 2021-08-06 at 1 18 59 PM
and then on subsequent restarts of ngrok via `valet share` it shows normally like this:
Screen Shot 2021-08-06 at 1 20 20 PM

That said, I've no objection to upgrading the default binary again. It's been awhile.

drbyte (approved):

drbyte:
> Mine requires me to hit CTRL-U to update; did you do that, Chris?
> 
> image

Good point. Mine is set to auto-update via my `~/.ngrok2/ngrok.yml` file, but the CTRL+U option is fine too.

```yml
authtoken: (redacted)

#docs: https://ngrok.com/docs

region: us

json_resolver_url: ""
dns_resolver_ips: []

# root_cas options: trusted, host, or path-to-a-CA-file
root_cas: trusted

# auto-update
update: true
```

PR #1083, opened 2021-08-06
Added ability to use a `.valetphprc` file with `valet use` in a project to define the required PHP version

MeMattStone:
This extension to the functionality of the `valet use` command adds a check for a `.valetrc` file when no version is passed as an argument, which contains the PHP version for the project and works in a similar manner to the `nvm use` command in Node Version Manager to save developers time and avoid mistakes when working across projects that have different PHP version requirements.

e.g. `.valetrc` in a project root could contain:
```
php@7.3
```

When `valet use` is run without specifying a PHP version you could expect one of the following outputs:

**Unlinking the current version for the specified version**
```bash
Found '/Users/johnnyappleseed/Sites/example/.valetrc' with version: php@7.3
Unlinking current version: php@7.4
Linking new version: php@7.3
Updating PHP configuration...
Restarting php@7.3...
Restarting nginx...
Valet is now using php@7.3.

Note that you might need to run composer global update if your PHP version change affects the dependencies of global packages required by Composer.
```

**Specified version is already linked**
```bash
Found '/Users/johnnyappleseed/Sites/example/.valetrc' with version: php@7.3
Valet is already using php@7.3.
```

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -495,7 +495,15 @@ You might also want to investigate your global Composer configs. Helpful command
      */
     $app->command('use [phpVersion] [--force]', function ($phpVersion, $force) {
         if (!$phpVersion) {
-            return info('Valet is using ' . Brew::linkedPhp());
+            if (file_exists($path = getcwd() . '/.valetrc')) {
+                if ($phpVersion = trim(file_get_contents($path))) {
+                    info('Found \'' . $path . '/.valetrc\' with version <' . $phpVersion . '>');
+                } else {
+                    return info('.valetrc does not contain a valid PHP version' . $phpVersion);
+                }
+            } else {
+                return info('Valet is using ' . Brew::linkedPhp());
+            }
         }
 
         PhpFpm::validateRequestedVersion($phpVersion);
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -495,14 +495,14 @@ You might also want to investigate your global Composer configs. Helpful command
      */
     $app->command('use [phpVersion] [--force]', function ($phpVersion, $force) {
         if (!$phpVersion) {
-            $path = getcwd() . '/.valetrc';
+            $path = getcwd() . '/.valetphprc';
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for this. 

- Question: Given that this is specifically geared around which PHP version to use, perhaps `.valetrc` is too generic? Would `.valetphprc` be suitable?

- I suppose if someone is actively using this feature then they'll be willing to forego the ability to check which PHP version is currently set and allow calling `valet use` to just switch them instead of reporting as it would if the rc file didn't exist.  It may be an edge case, but is a limitation created by this.

cli/valet.php
```diff
@@ -495,7 +495,18 @@
...
         if (!$phpVersion) {
-            return info('Valet is using ' . Brew::linkedPhp());
+            $path = getcwd() . '/.valetrc';
+            $linkedVersion = Brew::linkedPhp();
+            if (!file_exists($path)) {
+                return info(sprintf('Valet is using %s.', $linkedVersion));
+            }
+
+            $phpVersion = trim(file_get_contents($path));
+            info('Found \'' . $path . '\' with version: ' . $phpVersion);
```
drbyte, line 505: Instead of `with version:` perhaps `requesting version:` or `specifying version:` might be more accurate?

MeMattStone:
I did originally consider using `.phprc` but this would've caused confusion with the PHP Runtime Configuration environment variable with is referred to as PHPRC. I like `.valetphprc` as you're right this file is very specific for valet and the PHP version so I'll make this change.

I did consider adding an extra flag to check the currently linked version but as you've pointed out the file negates the need to do this so i omitted this feature.

cli/valet.php
MeMattStone: Yes this makes sense, I'll update this PR shortly with these changes

mattstauffer:
Huh. This is clever and simple. I like it!

mattstauffer:
Worked great for me, looks very simple code path wise, gonna tag it.

PR #1087, opened 2021-08-27
Add support for parked domains to some commands

antonioribeiro:
Some commands doesn't support parked domains today:

- secure
- unsecure
- open

This PR will do sucessfully secure a parked domain:

```
> valet secure sitename.dev.a17.io
Restarting nginx...
The [sitename.dev.a17.io] site has been secured with a fresh TLS certificate.
```

Instead of adding the .test TLD to it and securing an inexistent domain:

```
The [sitename.dev.a17.io.test] site has been secured with a fresh TLS certificate.
```

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -948,4 +948,22 @@ class Site
 
         return $this->valetHomePath().'/Certificates'.$url.$extension;
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


drbyte:
Problem: with these changes, calling `valet secure foo` now no longer secures `foo.test`

PR #1088, opened 2021-08-29
Add support for parked domains to some commands

antonioribeiro:
Some commands doesn't support parked domains today:

- secure
- unsecure
- open

This PR will do successfully secure a parked domain:

```
> valet secure sitename.dev.a17.io
Restarting nginx...
The [sitename.dev.a17.io] site has been secured with a fresh TLS certificate.
```

Instead of adding the .test TLD to it and securing an inexistent domain:

```
The [sitename.dev.a17.io.test] site has been secured with a fresh TLS certificate.
```


antonioribeiro:
Moving @drbyte comment here

_Problem: with these changes, calling valet secure foo or valet secure foo.test now no longer secures foo.test_

antonioribeiro:
@drbyte , yes you are right, I had a precedence bug on my code, it's fixed.

antonioribeiro:
And I wrongly sent it via my master branch, so I just closed the other one and opened this one with a proper feature branch.

drbyte:
It still doesn't add a TLD.

```sh
cd ~/Sites
mkdir foo
cd foo
valet secure
```
Output:
```
Restarting nginx...
The [foo] site has been secured with a fresh TLS certificate.
```
Note that the TLD isn't added, and therefore also the `~/.config/valet/Nginx` site config file created is `foo`, not `foo.test`, and in the `~/.config/valet/Certificates` directory the TLD also isn't included.

Same is happening with `valet secure` as shown above, but also with `valet secure foo` or `valet secure foo.test`


## Side technical notes:
- `valet secure` should secure the current directory project, using its directory name and adding the TLD.
- `valet secure foo` should secure `foo.test` regardless what directory one's shell is currently in

antonioribeiro:
Well, not doing the same here for me:

https://user-images.githubusercontent.com/3182864/131266754-57005bc2-08d8-4635-b591-081f4baf3eaf.mov

And if you look at the code, it will behave differently only if you have a parked folder AND the site you are trying to secure is a parked site, any other executions will have the exact same behaviour as before because they all fallback to the original code:

``` php
return ($domain ?: $this->host(getcwd())).'.'.$this->config->read()['tld'];
```

If we compare the codes

```php

$url = ($domain ?: Site::host(getcwd())).'.'.Configuration::read()['tld'];

return ($domain ?: $this->host(getcwd())).'.'.$this->config->read()['tld'];
```

mattstauffer:
@antonioribeiro I'm having a little trouble following this. 

When I think of a domain that's *outside of the normal*, I'm thinking *linked* domains, not *parked domains*.

Could you tell me how, exactly, you have a parked domain that's not using the standard `.test` TLD?

antonioribeiro:
Hey @mattstauffer, I believe I'm using an undocumented feature 😆

```
# Create a parked folder
mkdir ~/code/.valet/parked

# Move into it
cd ~/code/.valet/parked

# Create a link for your motf domain 
ln -s ~/code/tighten/website tighten.dev.a17.io

# Run this only the first time to configure Valet for your parked domain
valet park ~/code/.valet/parked
```

Then you probably have your domain `valet parked` now:

```
+------------------------------------+-----+------------------------------------------------+---------------------------------------------------------+
| Site                               | SSL | URL                                            | Path                                                    |
+------------------------------------+-----+------------------------------------------------+---------------------------------------------------------+
| tighten.dev.a17.io                 |     | http://tighten.dev.a17.io.test                 | /Users/antonioribeiro/code/client/site                  |
+------------------------------------+-----+------------------------------------------------+---------------------------------------------------------+
```

The feature is: most people think they need to use http://tighten.dev.a17.io.test to access it, but you can **also** access it using the name of the folder as the domain http://tighten.dev.a17.io.

But you cannot secure `tighten.dev.a17.io`, that's where this PR comes in handy:

```
valet secure tighten.dev.a17.io
```

And you just secured a custom domain on Valet:

Screen Shot 2022-04-06 at 18 14 30

It will not show as `SSL` in `valet parked`, but it may be also doable.

mattstauffer:
Thanks so much for the update!

I think I want to keep this feature undocumented and not official, because if we were having to do a lot of work to support it that would be rough and I'd just choose to cut it.

But... this is a pretty minor change. So let's do it :)

antonioribeiro:
Thanks for merging it, @mattstauffer! ❤️

I believe a real support for custom domains on Valet is something that the community needs, and is probably using workarounds to circumvent the lack of support, as I did.

Our (at [AREA 17](https://area17.com/)) use case for this is pretty simple: some third party services, like https://checkout.com (a payment provider), doesn't allow `.test` as a valid domain when configuring callback URLs, as it cannot be part of global DNS, but also because it's good to enforce the same level of security while in development as we would have in production, to find and fix problems sooner than later, so they require us to use domains under a public (and known) TLD (.com, .org, .net, .io, .com.br).

Yes we could solve it by using something like ngrok or Expose to have both public domain and a SSL cert, but they also come with some overhead, as we now need to make sure daemon is always running, and if we restart it we might need to get back to the service and change the callback URLs to the new domain (and potentially our .env file), as the daemon may hang/close automatically after some time, unless you pay for the service (one more variable). Those services have a better use when working with API, webhooks and other features that requires the (local) host to be reachable from outside when testing the application.

A Valet custom domain could solve all this because once you configure, you don't have to think about it anymore, it's there, it just works.

mattstauffer:
Rolling this back. I tested it and couldn't reproduce DrByte's issue but now I can, and others can too--see #1239. I'll re-evaluate this later.. clearly I need to get more hands-on familiar with this undocumented thing you're working with.

PR #1091, opened 2021-09-22
Change logo to work on Dark Theme

xiCO2k:


mattstauffer:
Screenshots for PRs like this are helpful.

This is what it looks like without this PR:

image

This is what it looks like with this PR:

image

I appreciate the idea but don't particularly like the implementation. I'll think about it and see if anyone can possibly propose a better solution. Thanks!

xiCO2k:
Let me try add more border on this one

xiCO2k:
Now looks like this on Dark mode:

image

On the Light Mode:
image

I personally prefer the other one since it matches the other repos.

Thanks for reviewing.

mattstauffer:
Thanks!

PR #1111, opened 2021-11-29
Set Brew PHP latest version to 8.1

caelumek:
Set Brew PHP latest version to 8.1 as PHP 8.1 is released and command `valet use php@8.0` now sets 8.1 and it is impossible to set 8.0.


driesvints (approved):

mattstauffer:
Thanks @caelumek!

Issue #1011, opened 2020-12-04
Update to rely on Illuminate Collections instead of Tightenco/Collect


mattstauffer:
Any deprecation issues we need to consider? It'll force everyone to update their installers but they'll have to anyway. Any other global packages that would stop being compatible to be installed at the same time if we did this?

soilSpoon:
`illuminate/collections` exists since `^8.0`. It only supports `PHP` `^7.3` and `^8.0`.
But valet supports `PHP` `^5.6` and above.

wturrell:
FYI, Valet is now impacted by this: https://github.com/tighten/collect/issues/261 with PHP 8.1

driesvints:
Started working on this here: https://github.com/laravel/valet/pull/1114

driesvints:
Hi there, we're cleaning up the Valet issue tracker and are moving all feature requests/support questions to the Discussions tab. We'll be using the issue tracker solely for bugs with Valet from now on. You're welcome to continue the discussion on in the Discussions tab. Thanks!

PR #1114, opened 2021-11-30
Switch to Illuminate Collections

driesvints:
This PR switches Valet to use the new Illuminate Collections instead. This will require a new major release. You might want to split off the current master branch into a `2.x` branch before merging this.

Because we're switching to a new Illuminate component introduced in Laravel 8, we'll have to drop old PHP and Laravel versions.

Valet 2.x can still be maintained simultaneously if needed for people on lower Laravel/PHP versions.

Closes https://github.com/laravel/valet/issues/1011

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [5.6, '7.0', 7.1, 7.2, 7.3, 7.4, '8.0', '8.1']
+        php: [7.3, 7.4, '8.0', '8.1']
 
     name: PHP ${{ matrix.php }}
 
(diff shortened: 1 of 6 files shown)


driesvints:
One thing that @mattstauffer noted to me is that this'll leave people working on apps on older and newer PHP versions simultaneously in the dark since they won't be able to switch to all versions easily anymore. Ideally we should find a way to offer that somehow.

Unfortunately I don't see how we can do that since requiring both tighten/collect and illuminate/collections can't be done due to namespace conflicts. We could maybe delay all of this until Laravel 10 has been released since tighten/collect will still be maintained for Laravel 8 & 9. When Laravel 10 is released we'll need to update the illuminate/container dependency on Valet and then it might be a good time to phase out all PHP versions below PHP 7.3.

driesvints:
@mattstauffer I don't think we use any tighten/collect namespaces in Valet? They're already Illuminate namespaces (or mostly the `collect` helper? Or am I misunderstanding you?

mattstauffer:
Oh. That could be the case 🤦 didn't even check.

driesvints:
Gonna close this one for now.

PR #1151, opened 2021-12-06
Update constraints

driesvints:
This PR updates the versions in the composer.json file to use the proper caret symbol like we do in all of our other libraries. This will ensure SemVer updates are properly handled. Further more, I've limited the allowed PHP versions to only the ones we support at the time. The current constraint allowed Valet to be installed on any PHP version that will be released in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,11 +24,11 @@
         }
     },
     "require": {
-        "php": ">=5.6",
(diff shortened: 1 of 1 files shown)


mattstauffer:
Thanks @driesvints!

PR #1152, opened 2021-12-09
Clean valet.sock when restarting phpfpm

mattstauffer:
I've noticed that I'm seeing more and more occasions when `valet.sock` sticks around when it shouldn't. I'm hoping this will help.


mattstauffer:
Looking at this again, I'm actually concerned this wouldn't work, because I think restarting PHPFPM doesn't actually create a new `valet.sock`... so I need to find another solution.

Issue #1153, opened 2021-12-13
Valet Secure - lifetime too long since Chrome 58 - Max CA lifetime is 368 days


icyrizard:
- Valet Version: 2.16.2
- PHP Version: php7.4

### Description:
Valet secure creates certificates that are too long for Chrome v58, see max allowed life time of these certificates here: https://chromium.googlesource.com/chromium/src/+/HEAD/net/docs/certificate_lifetimes.md

If the certificate is longer than 368 days you will get the following error for each request:
~~~
net::ERR_CERT_VALIDITY_TOO_LONG
~~~

Life time is hardcoded and set to  `730 days` in `valet secure -> Site::secure`. This should be configureable.

https://github.com/laravel/valet/blob/d4aad6f309b787a15b9748d49ad3e2ccdcf8708f/cli/Valet/Site.php#L518

!! Note that this is NOT a duplicate of https://github.com/laravel/valet/issues/1103

### Steps To Reproduce:
1. Run valet
2. Run valet park in your toplevel folder
3. Run valet secure <folder name>
4. Run `valet parked` and confirm the following,

Max sure the SSL certificate column has an X as seen below:

~~~bash
+-------------------+-----+-------------------------------+---------------------------------------+
| Site              | SSL | URL                           | Path                                  |
+-------------------+-----+-------------------------------+---------------------------------------+
| config-service    |  X  | https://config-service.test   | /Users/richard/Work/config-service    |
+-------------------+-----+-------------------------------+---------------------------------------+
~~~

### Diagnosis

```
<details>
<summary>sw_vers</summary>
<pre>ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H1419</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 2.16.2</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "loopback": "127.0.0.1",
    "paths": [
        "/Users/richard/Work"
    ]
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "laravel/valet": "^2.16",
        "squizlabs/php_codesniffer": "*",
        "phpmd/phpmd": "*",
        "friendsofphp/php-cs-fixer": "2.19",
        "laravel/installer": "^4.0"
    },
    "config": {
	    "platform-check": false
    }
}</pre>
</details>
<details>
[...]

Issue #1103, opened 2021-10-12
Root CA expires after 730 days without a convenient way to renew it


jaylinski:
## Clear description of your problem
The generated root CA `LaravelValetCASelfSigned.pem` expires after 730 days, which will result in the error `cURL error 60: SSL certificate problem: unable to get local issuer certificate` when sending a PHP cURL request to a Valet secured domain.

### Option 1

Valet auto-rotates (renews) the root CA automatically and informs the user to recreate a CA bundle for cURL by running `brew reinstall ca-certificates`.

### Option 2

Valet provides a `valet renew` command that creates a new root CA and creates new certs for existing sites, adds them to the Keychain, then executes `brew reinstall ca-certificates` in order to regenerate a CA certificate bundle from the keychain.

## Current behavior
The root CA expires and things break with `cURL error 60: SSL certificate problem: unable to get local issuer certificate`.

There is no convenient way to create a new root CA and child-certificates for existing sites.

The current workaround is to delete the `valet/CA`-folder and re-secure existing sites.

## Steps to Reproduce

1. Install Valet
2. Wait 731 days and run `brew reinstall ca-certificates` (the CA bundle for cURL won't contain the expire Valet Root CA anymore)
3. Make a https request via PHP cURL to a site secured with Valet
4. Get the error `cURL error 60: SSL certificate problem: unable to get local issuer certificate`
5. Spend a lot of hours in order to figure out how certificate validation on macOS with cURL and PHP works 🙃 

### Output of these steps
/

## Possible solution

Valet either auto-rotates the root CA (by checking its age) or provides a command to manually renew the root CA.

I can provide a PR for one of these options, if desired.

## Diagnosis
<details>
    <summary>sw_vers</summary>
    <pre>ProductName:	macOS
ProductVersion:	11.6
BuildVersion:	20G165</pre>
</details>
<details>
    <summary>valet --version</summary>
    <pre>Laravel Valet 2.16.0</pre>
</details>

driesvints:
Hey @jaylinski, I think "Valet either auto-rotates the root CA (by checking its age)" is would be a good pragmatic solution here. Are you willing to send in a PR for that? 

Also https://github.com/laravel/valet/pull/1154 got merged so the expiry date is half of what it was before.

jaylinski:
> Are you willing to send in a PR for that?

@driesvints Yes, I'll send in a PR this week.

PR #1154, opened 2021-12-13
Add 'expireIn=' option and use in the openssl command

icyrizard:
This pr is meant to overcome an issue in concerning Chrome v58 and up.

### The tests I've conducted are:

**Test the default setting**
1. Clean linked directories, `$ valet unpark`
2. Run without `expireIn` should go to default. `$ valet secure <my_laravel_project>`
3. See if Chrome is happy - yes it is - validity is set default to 368 days now.

**Test with a too high `expireIn`**

1. Clean linked directories, `$ valet unpark`
2. Run without secure, `$ valet secure --expireIn=730 <my_laravel_project>`
3. See if Chrome is not happy, throws same error `net::ERR_CERT_VALIDITY_TOO_LONG`

**Test with a lower `expireIn`**
1. Clean linked directories, `$ valet unpark`
2. Run with shorter `expireIn`, `$ valet secure --expireIn=120 <my_laravel_project>`
3. See if Chrome is happy.

**Extra test is done by checking Mac's Keychain Access for default option, 368 days counting from December 13 2021**
image

The expireIn option that can be passed to `$ valet secure <domain> --expireIn=<nDays>` can be used to overcome the aforementioned issue.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -464,9 +464,10 @@ class Site
      *
      * @param  string  $url
      * @param  string  $siteConf  pregenerated Nginx config file contents
+     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
      * @return void
      */
-    public function secure($url, $siteConf = null)
+    public function secure($url, $siteConf = null, $caExpireInDays = 368)
     {
         $this->unsecure($url);
 
@@ -476,9 +477,8 @@ class Site
 
         $this->files->ensureDirExists($this->nginxPath(), user());
 
-        $this->createCa();
-
-        $this->createCertificate($url);
+        $this->createCa($caExpireInDays);
+        $this->createCertificate($url, $caExpireInDays);
 
         $this->files->putAsUser(
             $this->nginxPath($url), $this->buildSecureNginxServer($url, $siteConf)
@@ -488,9 +488,10 @@ class Site
     /**
      * If CA and root certificates are nonexistent, create them and trust the root cert.
      *
+     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
      * @return void
      */
-    public function createCa()
+    public function createCa($caExpireInDays)
     {
         $caPemPath = $this->caPath('LaravelValetCASelfSigned.pem');
         $caKeyPath = $this->caPath('LaravelValetCASelfSigned.key');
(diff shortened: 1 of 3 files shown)


driesvints:
Hey @icyrizard, can you please refrain from linking to an issue in commit messages? It messes up the history/conversation in GitHub. See the log in https://github.com/laravel/valet/issues/1153

icyrizard:
@driesvints Ah I see, my bad, I honestly thought that was a useful feature before you mentioned this. I'll squash the commits into one.

mattstauffer:
Thanks @icyrizard! This looks great. Per @driesvints's suggestion, I'd also like to see if we can use some of this same code to allow people to update expired/expiring certificates per #1103

stancl:
Not sure if this was meant to fix it or if this caused it, but I just reinstalled my local setup due to a PHP version upgrade and started getting this after I tried to `secure` a site:

```php
➜ samuel@MacBook-Pro  ~/Sites/helpdesk git:(master) ✗ valet secure
Enter PEM pass phrase:

In Nginx.php line 138:
                                                                                                                                                                                                                                                                                                                                                                                      
  Nginx cannot start; please check your nginx.conf [1: nginx: [emerg] cannot load certificate "/Users/samuel/.config/valet/Certificates/helpdesk.test.crt": BIO_new_file() failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/Users/samuel/.config/valet/Certificates/helpdesk.test.crt','r') error:2006D080:BIO routines:BIO_new_file:no such file)  
  ].                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                      

secure [--expireIn EXPIREIN] [--] [<domain>]
```

Was pretty confusing to debug, but after adding some `dd()` calls to the vendor valet files, I found that `$caExpireInDays` is `null` inside `secure()`, despite the parameter having a default value of `368`. Which was leading to openssl complaining about the number of days missing:

```
bad number of days: invalid\n
usage: x509 args\n
```

And therefore only creating the `csr` request and not the actual `crt` certificate that the nginx file tries to include.

**Passing the number of days manually using `--expireIn=368` fixes everything.**

My guess is that the issue is caused by this line having `null` as a default:
https://github.com/laravel/valet/blob/8ff684087b29a88655ef686edfba571164f3e75a/cli/valet.php#L182

So if you have time right now, please test if you can reproduce this and if setting `$expireIn` to `368` ^ here then fixes the issue. Alternatively, you can revert this and re-merge when it's fixed in the implementation.

I wanted to open an issue pointing this out and saying that passing the `--expireIn` flag manually fixes this so that others don't have to go through the same painful debugging, but I saw that a related PR (this one) was merged 3 hours ago which made finding the (likely) root cause much easier.

Either way, it would be great to patch this at least somehow and tag a new release, since right now running `valet secure` or just updating your PHP version (seems like that triggers a valet reinstall/reconfigure) sort of bricks the entire nginx configuration.

PR #1155, opened 2021-12-13
Fix certificate generation

stancl:
Fixes issue described in https://github.com/laravel/valet/pull/1154#issuecomment-992900342

I just tested this multiple times and can confirm that making it `null` makes `valet secure` break nginx config, but making it `368` in the source code makes `valet secure` work as expected.


mattstauffer:
Thanks for the PR!

PR #1161, opened 2021-12-15
Avoid expiring of root certificate authority

### LINKED ISSUE #1103: Root CA expires after 730 days without a convenient way to renew it

jaylinski:
### Description

Currently Laravel Valet creates a root certificate authority with the same validity period as normal certificates.

It is correct to limit normal certificates to a maximum of 396 days in order to adhere to certificate baseline-requirements as defined in https://github.com/cabforum/servercert/blob/main/docs/BR.md (2020-09-01).

But the validity of root CAs is not limited, thus it makes sense to create the root CA with a longer validity period. The Lets Encrypt Root CA `ISRG Root X1` has a validity period of 20 years, so that should also be the default period of the Laravel Valet Root CA.

This resolves #1103 without having to implement auto-rotation of expired root CAs. (Because the amount of people that use the same machine and OS for over 20 years is probably zero.)

There is only one downside: existing users won't benefit from this change, because they would have to manually delete the root CA from the keychain in order for Valet to create a new CA. But since I was the first one to complain about this issue, I guess this doesn't matter that much.

### Screenshot

Bildschirmfoto 2021-12-16 um 17 18 32

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -464,10 +464,15 @@ class Site
      *
      * @param  string  $url
      * @param  string  $siteConf  pregenerated Nginx config file contents
-     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
+     * @param  int  $certificateExpireInDays  The number of days the self signed certificate is valid.
+     *                                        Certificates SHOULD NOT have a validity period greater than 397 days.
+     * @param  int  $caExpireInYears  The number of years the self signed certificate authority is valid.
+     *
+     * @see https://github.com/cabforum/servercert/blob/main/docs/BR.md
+     *
      * @return void
      */
-    public function secure($url, $siteConf = null, $caExpireInDays = 368)
+    public function secure($url, $siteConf = null, $certificateExpireInDays = 396, $caExpireInYears = 20)
     {
         $this->unsecure($url);
 
@@ -477,8 +482,10 @@ class Site
 
         $this->files->ensureDirExists($this->nginxPath(), user());
 
-        $this->createCa($caExpireInDays);
-        $this->createCertificate($url, $caExpireInDays);
+        $caExpireInDate = (new \DateTime())->diff(new \DateTime("+{$caExpireInYears} years"));
+
+        $this->createCa($caExpireInDate->format('%a'));
+        $this->createCertificate($url, $certificateExpireInDays);
 
         $this->files->putAsUser(
             $this->nginxPath($url), $this->buildSecureNginxServer($url, $siteConf)
@@ -488,7 +495,7 @@ class Site
     /**
      * If CA and root certificates are nonexistent, create them and trust the root cert.
      *
-     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
+     * @param  int  $caExpireInDays  The number of days the self signed certificate authority is valid.
      * @return void
      */
     public function createCa($caExpireInDays)
@@ -526,7 +533,7 @@ class Site
      * Create and trust a certificate for the given URL.
      *
      * @param  string  $url
-     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
+     * @param  int  $caExpireInDays  The number of days the self signed certificate is valid.
      * @return void
      */
     public function createCertificate($url, $caExpireInDays)


cli/Valet/Site.php
```diff
@@ -464,10 +464,13 @@ public function secured()
...
      * @param  string  $siteConf  pregenerated Nginx config file contents
-     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
+     * @param  int  $certificateExpireInDays  The amount of days the self signed certificate is valid.
+     *                                        Certificates SHOULD NOT have a validity period greater than 397 days.
+     * @param  int  $caExpireInYears  The amount of years the certificate authority is valid.
+     * @see https://github.com/cabforum/servercert/blob/main/docs/BR.md
      * @return void
      */
-    public function secure($url, $siteConf = null, $caExpireInDays = 368)
+    public function secure($url, $siteConf = null, $certificateExpireInDays = 396, $caExpireInYears = 20)
```
jaylinski, line 473: I'm not sure why `368` was picked by @icyrizard, according to the baseline-requirements it can be 396 days:

https://github.com/cabforum/servercert/blob/main/docs/BR.md (2020-09-01)

cli/Valet/Site.php
```diff
@@ -477,8 +480,11 @@ public function secure($url, $siteConf = null, $caExpireInDays = 368)
 
         $this->files->ensureDirExists($this->nginxPath(), user());
 
-        $this->createCa($caExpireInDays);
-        $this->createCertificate($url, $caExpireInDays);
+        $caExpireInterval = (new \DateInterval(sprintf('P%dY', $caExpireInYears)));
+        $caExpireInDate = (new \DateTime())->diff((new \DateTime())->add($caExpireInterval));
+
+        $this->createCa($caExpireInDate->format('%a'));
```
jaylinski, line 486: Simpler alternative: `$this->createCa($caExpireInYears * 365)` (but this misses leap years)

mattstauffer: Good note--love this even if it's bit less simple

cli/Valet/Site.php
mattstauffer: Yah, I checked the original PR: https://github.com/laravel/valet/pull/1154 

And there's no explanation given, so I'm totally fine for moving it to 396. Good catch!

cli/Valet/Site.php
```diff
@@ -477,8 +482,11 @@ public function secure($url, $siteConf = null, $caExpireInDays = 368)
...
```
mattstauffer, line 488: What if we simplify this a bit?

```suggestion
        $caExpireInDate = (new DateTime())->diff(new DateTime("+{$caExpireInYears} years"));
        $this->createCa($caExpireInDate->format('%a'));
```

cli/Valet/Site.php
```diff
@@ -488,7 +496,7 @@ public function secure($url, $siteConf = null, $caExpireInDays = 368)
     /**
      * If CA and root certificates are nonexistent, create them and trust the root cert.
      *
-     * @param  int  $caExpireInDays  The mount of days the self signed certificate is valid.
+     * @param  int  $caExpireInDays  The mount of days the self signed certificate authority is valid.
```
mattstauffer, line 499:
```suggestion
     * @param  int  $caExpireInDays  The number of days the self signed certificate authority is valid.
```

mattstauffer (changes requested):
I just suggested a tiny refactor. Great work here!

cli/Valet/Site.php
jaylinski: Much better, yes. :+1:

cli/Valet/Site.php
jaylinski: :+1: Done.

I also corrected this for the function `Site::createCertificate()`.

mattstauffer:
Fantastic. Thank you so much @jaylinski!

PR #1157, opened 2021-12-15
PHPUnit: Convert deprecations to exceptions

bram-pkg:
On PHP 8.1 there are a lot of deprecation warnings.

PR #1163, opened 2021-12-20
Replace HTTPFul with Guzzle

bram-pkg:
As commented on #1157, HTTPful contains some deprecated calls and replacing it with Guzzle is better.

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -3,7 +3,8 @@
 namespace Valet;
 
 use DomainException;
-use Httpful\Request;
+use GuzzleHttp\Client;
+use GuzzleHttp\Utils;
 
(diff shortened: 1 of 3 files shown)


driesvints:
@bram-pkg fyi: we fixed a small issue with this PR here: https://github.com/laravel/valet/pull/1164

driesvints:
@bram-pkg seems that there's more issues with this PR:

```
Fatal error: Uncaught Error: Call to undefined method GuzzleHttp\Utils::jsonDecode() in /Users/driesvints/.composer/vendor/laravel/valet/cli/Valet/Valet.php:81
```

Guzzle v6 doesn't have this. I'll try to fix this.

driesvints:
Released v2.18.3 with a fix for this.

Issue #1177, opened 2022-01-06
`valet use` does not stop previous php version service


cristiancalara:
- Valet Version: 2.18.4
- PHP Version: 8.1.1

### Description:
After I switch the PHP version using `valet use php@8.0`, the previous PHP@8.1 service is still running https://share.getcloudapp.com/12u0DEPN. On restart, the PHP@8.0 service will have a status of 'error' when running `sudo brew services restart` https://share.getcloudapp.com/xQuz099d

At this point, the version of PHP linked is 8.0 but the running PHP version is 8.1 (that's also the case when running `phpinfo()`).  If you run `valet use php@8.0` (I usually run this as I see an error that's caused by the site running on 8.1, so I'd expect that valet is also using 8.1), you get the warning saying that this version is already linked, and you should use `--force`

### Steps To Reproduce:
1. `valet use php`
2. `valet use php@8.0`
3. `sudo brew services list` - will show both the previous and the current service still running
4. restart computer
5. open a site `phpinfo()` - you should see PHP 8.1 (so the version used when running `valet use php`)
6. run `valet use php@8.0` - 
7. you should see a warning saying that this version is already linked - this is a bit confusing for me, as I'd expect to have php8.1 linked as shown in `phpinfo()` i.e. the linked and running version to be the same

### Debugging
I've looked a bit into this, and it seems that the intention in the code was to actually stop all the PHP services - https://github.com/laravel/valet/blob/master/cli/Valet/PhpFpm.php#L205 . But they are not actually closed because the `$this->brew->getRunningServices()` call does not return the services that are ran using `sudo` (as it's the case with PHP).

If I update the code at https://github.com/laravel/valet/blob/master/cli/Valet/Brew.php#L381 and add `sudo` https://share.getcloudapp.com/WnuqEAzd  before the command, the issue above is resolved. 

### Notes
- The issue above is not really critical, as you can switch between versions and things will get fixed. But I thought it might be worth raising as after looking through the code I saw that tried to actually stop the currently running PHP services.
- While on my tests, it was always the `php` version that was launched successfully and the older one having the error - I'm not 100% sure this is consistent, as I'm not sure of the order `launchctl` uses when opening the services. 

### Diagnosis

<details>
<summary>sw_vers</summary>
<pre>ProductName:	macOS
ProductVersion:	11.6
BuildVersion:	20G165</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 2.18.4</pre>
</details>
[...]

Issue #1102, opened 2021-10-11
Delete valet.sock on `valet start`, `valet stop`, and `valet restart`


mattstauffer:
## Clear description of your problem

In many circumstances `~/.config/valet/valet.sock` still persists when it shouldn't, leaving me with the wrong version of Valet running.

## Expected behavior

If Valet is running a version of PHP locally, I expect that same version to be what's running on web sites that are being served.

## Current behavior

Often I'll find that, when I restart my computer, or in other scenarios--still trying to exactly pin it down--I'll have PHP 8 running on the command line, managed by Valet, but a site will open and give a Composer error because I'm running PHP 7.1 or something.

## Steps to Reproduce

1. I still can't figure out exactly how to repro or what the source of this is, frustratingly!!

## Possible solution

Every time we run `valet stop` or `valet start` or `valet restart`, delete the `valet.sock` file if it exists.

We already do this when running `valet use`.

## Diagnosis

Not gonna run the whole thing, but I'm on v2.15.0 on one machine and v2.16.0 on the other and it's happening on both.

nicoverbruggen:
Heya @mattstauffer and @drbyte, I think I have some (hopefully) helpful info related to this particular issue.

I’m currently preparing for the release of [phpmon 4.1](https://github.com/nicoverbruggen/phpmon/tree/develop), which will use Valet’s own `valet use` command to switch between different versions of PHP! This was primarily done to avoid potential issues caused by the app invoking the `brew link` and `brew services` commands on its own.

I've been doing some testing over the past week and I bumped into an issue today after rebooting my machine, and I think it's a potential bug in Valet. (I was going to make a new issue but this particular issue *is the very thing I'm bumping into as well*.)

On my Mac, I have the latest version of Valet running and the `php` and `php@8.0` formulae (from Homebrew Core) are installed.

I have noticed that sometimes, whenever I reboot my machine, sites are being *served* using PHP 8.1, even though I switched to PHP 8.0 before rebooting. It is important to note that PHP 8.0 remains linked correctly after the reboot. 

(Here’s the wonky part: I cannot reproduce this *every* time. Switching versions and rebooting does not always seem to trigger this particular behaviour, so I'm a bit puzzled. I suspect this is a timing issue...)

After switching to `php@8.0` (using `valet use php@8.0` in the terminal) I went to inspect the output of `sudo brew services list`, and I noticed the following :

```
➜  ~ sudo brew services list
Name    Status  User File
[...]

PR #1178, opened 2022-01-08
Stop root homebrew services on php version switch

cristiancalara:
Before the change, when running `valet use` the code intended to stop currently running PHP services. But the `getRunningServices` method only returned non-root running services. As PHP services started by Valet are run using `sudo` (so running as root), they were not returned and subsequently not stopped.

This change is intended to fix the above and stop PHP services that are started by Valet on a PHP version switch.

Described the actual issue I was facing in https://github.com/laravel/valet/issues/1177 

A few notes on the implementation:
- I initially tried to make the change as small as possible, and tried to update the current command to be something like `{ sudo brew services list && sudo -u "cristian" brew services list } | grep started | awk '{ print $1; }'` i.e. basically to get the output of both root and non-root services at once. While the command above worked on my command line, it throws an error `"sh: {sudo brew services list && sudo -u "cristian" brew services list }: command not found"` when run via the code. I didn't investigate this further, as I thought that using "advanced" features like the "&&" and "{}" was not worth it. I considered that the readability of this approach was also not that great, so I scraped this implementation.
- On this new implementation, I was not sure of how to name the two new methods. Ended up with `getRunningServicesAsRoot` and `getRunningServicesAsUser`. I considered using `LaunchDaemons` and `LaunchAgents` as this is how they are called in Homebrew https://share.getcloudapp.com/nOu9NlWJ - but decided against as I think the "average" brew user (like myself) would probably not know this. So went with "AsRoot" and "AsUser". Would welcome any thoughts on this.

PS. Noticed that the Diagnose command is also using `brew services list`. Was thinking that maybe it would be a good idea to also add `sudo brew services list`? As Valet runs most of the services using sudo, so they won't be on the `brew services list` command output.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -384,7 +384,7 @@ class Brew
 
     /**
      * Get the currently running brew services as root.
-     * i.e. /Library/LaunchDaemons (started at boot)
+     * i.e. /Library/LaunchDaemons (started at boot).
      *
      * @return \Illuminate\Support\Collection
      */
@@ -395,7 +395,7 @@ class Brew
 
     /**
      * Get the currently running brew services.
-     * i.e. ~/Library/LaunchAgents (started at login)
+     * i.e. ~/Library/LaunchAgents (started at login).
      *
      * @return \Illuminate\Support\Collection
      */
@@ -410,7 +410,7 @@ class Brew
      * @param  bool  $asUser
      * @return \Illuminate\Support\Collection
      */
-    public function getRunningServices(bool $asUser = false)
+    public function getRunningServices($asUser = false)
     {
         $command = 'brew services list | grep started | awk \'{ print $1; }\'';
         $onError = function ($exitCode, $errorOutput) {
@@ -421,7 +421,7 @@ class Brew
 
         return collect(array_filter(explode(PHP_EOL, $asUser
             ? $this->cli->runAsUser($command, $onError)
-            : $this->cli->run('sudo ' . $command, $onError)
+            : $this->cli->run('sudo '.$command, $onError)
         )));
     }
 
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -371,20 +371,58 @@ class Brew
     }
 
     /**
-     * Get the currently running brew services.
+     * Get all the currently running brew services.
      *
      * @return \Illuminate\Support\Collection
      */
-    public function getRunningServices()
+    public function getAllRunningServices()
     {
-        return collect(array_filter(explode(PHP_EOL, $this->cli->runAsUser(
-            'brew services list | grep started | awk \'{ print $1; }\'',
-            function ($exitCode, $errorOutput) {
-                output($errorOutput);
+        return $this->getRunningServicesAsRoot()
+            ->concat($this->getRunningServicesAsUser())
+            ->unique();
+    }
 
-                throw new DomainException('Brew was unable to check which services are running.');
-            }
-        ))));
+    /**
+     * Get the currently running brew services as root.
+     * i.e. /Library/LaunchDaemons (started at boot)
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    public function getRunningServicesAsRoot()
+    {
+        return $this->getRunningServices();
+    }
+
+    /**
+     * Get the currently running brew services.
+     * i.e. ~/Library/LaunchAgents (started at login)
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    public function getRunningServicesAsUser()
+    {
+        return $this->getRunningServices(true);
+    }
+
+    /**
+     * Get the currently running brew services.
+     *
+     * @param  bool  $asUser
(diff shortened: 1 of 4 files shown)


cli/Valet/Brew.php
```diff
@@ -370,21 +370,59 @@ function ($exitCode, $errorOutput) use ($formula) {
...
-                throw new DomainException('Brew was unable to check which services are running.');
-            }
-        ))));
+    /**
+     * Get the currently running brew services.
+     *
+     * @param  bool  $asUser
+     * @return \Illuminate\Support\Collection
+     */
+    public function getRunningServices($asUser = false)
```
mattstauffer, line 413: My only concern with this PR is this syntax--I'm not a big fan of boolean parameters. I'm trying to think about how we can rework this to instead be a syntax that's clearer when you're calling it. I'll keep thinking, but other than that this PR looks great--thank you for making it!

mattstauffer: Here's an idea from @driftingly:

```php
    public function getCommandLineArgs($prefix = ''): array
    {
        $command = $prefix.'brew services list | grep started | awk \'{ print $1; }\'';
        $onError = function ($exitCode, $errorOutput) {
            output($errorOutput);

            throw new DomainException('Brew was unable to check which services are running.');
        };

        return [$command, $onError];
    }

    public function getRunningServicesAsRoot(): Collection
    {    
        return $this->formatCommandLineOutput($this->cli->run(...$this->getCommandLineArgs('sudo ')));
    }

    public function getRunningServicesAsUser(): Collection
    {
        return $this->formatCommandLineOutput($this->cli->runAsUser(...$this->getCommandLineArgs()));
    }

    public function formatCommandLineOutput($output): Collection
    {
        return collect(array_filter(explode(PHP_EOL, $output)));
    }
```

(I added the `prefix` concept because it was missing from his refactor). What do you think?

cristiancalara: Hey @mattstauffer, sorry for the delay on this, I somehow missed the notifications on this. 

I actually tried a few approaches to avoid the boolean param but in the end, decided it was not worth it. I actually went as far as creating a new class that would actually just be a wrapper for the list command, that would return the output as a collection that could then be filtered by started, stopped, error, etc. But decided against it as maybe adding a new class for only 3-4 lines of code didn't feel right. What do you think?

For the solution above, I actually tried a similar approach but decided against it as I couldn't find any good name for the functions. `getCommandLineArgs` and `formatCommandLineOutput` feels a bit too general? i.e. `Brew@getCommandLineArgs($prefix)` does not offer any info about the fact that this is actually only about the list command and it only gets "started" services? So I'd prefer having the boolean param over needing to add these 2 new functions without a proper name. But honestly, both solutions are in the "not ideal" column for me so I'm fine with both. So if you guys think this approach is better, I'll make the update, should be an easy one. Let me know :)

nicoverbruggen:
This PR will probably end up fixing #1102 as well as a side-effect (thank you for working on this @cristiancalara!).

I'm not sure if this is a "better" solution but I do have a suggestion — during my work on phpmon a while ago I found out that there's actually a "safer" way to find out which services are running:

```
sudo brew services info --all --json
```

Instead of showing you a list you need to use `grep` and `awk` on, this outputs JSON that can be filtered and parsed. If you filter the array based on the name of the service and check the `running` property, you have a "safe" way of checking whether a particular service is running as root. 

It is more of a hassle than the one-liner you currently have set up, of course... But assuming you use this, I do believe all you need to do is run `sudo brew services info --all --json` as superuser and this should always give you accurate information. (I don't think you need to run it as the regular user this way.)

You could do something like: 

```php
collect(json_decode($outputOfJsonCommand)
    ->filter(function ($service) { return $service->running; })
    ->pluck('name')
    ->all()
```

to get an array of all running services.

Not sure if you want to end up doing this — perhaps the current method is just fine, figured I'd share what I'm doing in my project. Cheers!

cristiancalara:
Hey @nicoverbruggen 

I do think that using the command you shared might make things a bit more clear i.e. more clear than using `awk`. But we won't actually get rid of the "boolean param" thing. As you still need to run that command twice (once as a normal user and one as root) i.e. `sudo brew services info --all --json` and `brew services info --all --json` return different results. 

I think that this improvement could be made, but not sure it's worth adding to the current PR? Maybe the change could be made as a separate "improvement/refactoring" PR?

nicoverbruggen:
@cristiancalara Good point. 

It does pay off to run the command as a non-user right after a new installation of the brew formulae (`php`, `nginx`, etc.), since I believe some of the services might still run as non-root at that point. 

As soon as you start using `valet use php@version` all services _should_ run as root, though.

mattstauffer:
Gonna merge as-is so we can get this into the codebase; if we end up needing do do any refactors let's throw them in future smaller PRs. Thanks all!

driesvints:
@nicoverbruggen @cristiancalara did this end up fixing #1102 ?

driesvints:
Thanks @nicoverbruggen. Good luck with the new release of PHP Monitor 👍

Issue #1171, opened 2022-01-04
valet bin only have x86 version of ngrok


xeon-zolt:
- Valet Version: 2.18.3
- PHP Version: 7.4

### Description:
when running valet share on m1 mac with apple silicon it says 
`sudo: unable to execute /Users/xeon/.composer/vendor/laravel/valet/bin/ngrok: Bad CPU type in executable`

### Steps To Reproduce:
run valet share on non x86 cpu arch

xeon-zolt:
I need to change the valet bash script with a if else and need to include binary of ngrok in bin for m1 mac

driesvints:
I currently seem to be able to run ngrok with Valet fine expect I'm stumbling into a connection issue. I think this works fine with m1 Mac's right now. I'll need to figure out what's causing the connection issue before I can verify that.

In the meantime please try to update homebrew to the latest version and reinstall Valet completely.

xeon-zolt:
let me try that

xeon-zolt:
is ngrok shipped with Valet or downloaded via homebrew ?

driesvints:
Shipped with Valet: https://github.com/laravel/valet/blob/master/bin/ngrok

I'm not really sure where I can find the latest ngrok binary but it indeed seems we should update it.

mattstauffer:
https://github.com/laravel/valet/pull/1082

Here's what a PR looks like to update the binary, FWIW, in case we need to.

driesvints:
Seems like v2.3.40 that currently ships is already the latest version so that can't be it.

xeon-zolt:
for me, it worked after installing rosetta.
but ngrok ships two different binaries for mac
I think we can include both the versions in the valet and 
modify script like this around line 81 
```
    arch = $(uname -m)
    if [[ $arch == arm64 ]]; then
        sudo -u "$USER" "$DIR/bin/ngrok-arm" http "$HOST.$TLD:$PORT" -host-header=rewrite $PARAMS
    else 
        sudo -u "$USER" "$DIR/bin/ngrok" http "$HOST.$TLD:$PORT" -host-header=
rewrite $PARAMS"
    fi

driesvints:
@xeon-zolt yeah that sounds like an idea maybe. Is there a repo with these on GitHub somewhere?

xeon-zolt:
ngrok 2.0 is not opensource only binary file available on main website which can be fetched

driesvints:
Do you have an url?

[...]

PR #1182, opened 2022-01-18
Add ARM support for ngrok

driesvints:
This PR adds an ARM binary for ngrok to support Apple Silicon MacBooks. Unfortunately this will add around 30MB to the repo so not entirely sure this is wanted...

For Dusk where we do something similar with the chrome driver binaries, [we opted to remove them in the next major version](https://github.com/laravel/dusk/pull/873) and to let the user explicitly install them during the install process. That way the repos stay small. I'm not sure if there's any interest in taking that approach? If so I can look into it maybe.

Closes https://github.com/laravel/valet/issues/1171

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,3 +19,4 @@ phpunit.xml.dist export-ignore
 UPGRADE.md export-ignore
 
 /bin/ngrok -diff
+/bin/ngrok-arm -diff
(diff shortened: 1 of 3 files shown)


xeon-zolt:
LGTM

xeon-zolt:
#1171

mattstauffer:
I think let's bundle both for now and then consider adding a Dusk-style installer switch in the next major. Thanks @driesvints!

driesvints:
Reverting this: https://github.com/laravel/valet/pull/1258

PR #1183, opened 2022-01-18
Fix syntax error in ./valet

UdaraJay:
The extra " in line 86 breaks the installer and valet commands

FIXES:

```
/usr/local/bin/valet: line 95: unexpected EOF while looking for matching `"'
/usr/local/bin/valet: line 97: syntax error: unexpected end of file
```


UdaraJay:
Hey @driesvints, heads-up on this^

driesvints:
Thanks! Sorry about that

PR #1192, opened 2022-02-06
Multiple/Parallel PHP Version Support for Valet

NasirNobin:
This pull request introduces an optional `--site` flag on the `valet use` command, that allows users to run multiple versions of PHP in parallel on different sites. 

Btw, I build this feature to solve my very own problem. I need to constantly switch between PHP versions, as I need to switch between old and new projects. This is being very handy for me. 

### To isolate a site's PHP version, run: 

```bash
valet use php@8.0 --site=site.test
```

or pass the site directory's name
```bash
valet use php@8.1 --site=site2 
```

or `cd` into the site's directory then use `.` to use current directory:
```bash
valet use php@8.1 --site .
```

### Remove isolation for a site: 
```bash
valet use default --site .
```

<details>
<summary>Here's a video showing this feature up and running</summary>

https://user-images.githubusercontent.com/13833460/152670301-d479a21e-c82c-443e-b732-f46afc67a2a9.mp4
</details>

### Behind the scene:
- Updates the FPM config file `/etc/php/8.0/php-fpm.d/valet-fpm.conf` and updates the listen parameter to something like: `valet80.sock`
- Publishes an Nginx config file on the `~/.config/valet/Nginx` location for that site.
- Updates that specific Nginx config file and replaces the `fastcgi_pass` param with `valet80.sock`

### Few other notes: 
- System default linked PHP will keep running on `valet.sock` as expected. 
- Global PHP switching commands like `valet use php@7.4` would still work as expected. Alongside, it will perform some operations on the isolated site's Nginx config to make sure those isolated sites still work as expected. 
- Running `valet install` or `valet restart` would do a scan on isolated sites to determine what versions of PHP are being used.  So it can restart each version of PHP correctly. 
- SSL-enabled sites could be isolated with the usual command. (or isolated site can obtain SSL, as well)
- Running `valet unsecure` on an isolated site will still keep the isolation and will remove the SSL.

### Future todos: 
- Need to have another command to list all the isolated sites and their PHP versions. (would be handy to have that)
- ~~Ideally, we would need to have a way to remove isolation for a site (valet use default --site .)~~

This modification doesn't seem to break any existing test cases or any existing flows.

I tried to add tests for the new methods that were introduced in this PR.

PR #1198, opened 2022-03-12
[WIP] Multiple/Parallel PHP Version Support for Valet -- Updated

mattstauffer:
I'm building on #1192 and needed to make a PR from a local working branch so I can push code to it. More details soon.


mattstauffer:
@NasirNobin I made some updates to your PR, including moving the isolation into its own command, which you can see here.

I'll likely make a new branch where I attempt to do the symlink change we talked about in the PR, so it doesn't mess this one up.

if you have time and you're interested, please feel free to review this PR, or just wait until I work on the symlink change (which I'm not getting to tonight, I don't think).

cli/Valet/Brew.php
```diff
@@ -15,12 +15,10 @@ class Brew
         'php@7.2',
         'php@7.1',
         'php@7.0',
-        'php@5.6',
         'php73',
         'php72',
         'php71',
         'php70',
-        'php56',
```
driesvints, line 23: This probably requires a new major version release?

mattstauffer: @driesvints Yah.. good call, thank you.

PR #1201, opened 2022-03-13
Update `valet.sock` to be a symlink to an existing `valet{version}.sock` file

mattstauffer:
This PR modifies #1198, which is an update of #1192.

In #1192/#1198, `valet.sock` is always used by the global Valet PHP version, and then Valet sets up `valet80.sock` etc. for versions of PHP that aren't used globally but are used by isolated sites.

This PR makes it so that every PHP version used by Valet, global or not, sets up a `valet{version}.sock` file, and the global version is set by simply symlink-ing `valet.sock` to the version-specific `.sock` file.

This PR:
- Makes that symlink change
- Catches some work I missed in #1192 to the syntax in valet.php
- Cleans up `normalizePhpVersion` with a regex from @fuelingtheweb; again should've just applied to #1192
- Drops the tests that were covering the old pre-symlink behavior

Bugs in testing:
- [x] Somehow I saw `valet.sock` disappear twice in the middle of my testing, which broke all sites. Re-running `valet install` fixed it but I'm not sure what deleted it in the first place. I'll try to keep testing.

### DIFF
--- a/cli/Valet/Nginx.php
+++ b/cli/Valet/Nginx.php
@@ -179,13 +179,13 @@ class Nginx
     }
 
     /**
-     * Return a list of all sites with explicit Nginx configurations
+     * Return a list of all sites with explicit Nginx configurations.
      *
      * @return \Illuminate\Support\Collection
      */
     public function configuredSites()
     {
-       return collect($this->files->scandir(VALET_HOME_PATH.'/Nginx'))
+        return collect($this->files->scandir(VALET_HOME_PATH.'/Nginx'))
             ->reject(function ($file) {
                 return starts_with($file, '.');
             });
--- a/cli/Valet/Nginx.php
+++ b/cli/Valet/Nginx.php
@@ -177,4 +177,17 @@ class Nginx
         $this->brew->uninstallFormula('nginx nginx-full');
         $this->cli->quietly('rm -rf '.BREW_PREFIX.'/etc/nginx '.BREW_PREFIX.'/var/log/nginx');
     }
+
+    /**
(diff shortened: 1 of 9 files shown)


mattstauffer:
I haven't tested this yet. Will test before asking for review.

mattstauffer:
I'm also wondering... do we even *need* a symlink? It's a socket.. there may not actually be anything fancy going on here. It may just need a file to exist, in which case we could just delete the file and make a new one.. not even worry about symlinking?

mattstauffer:
Current thinking about the symlink situation:

If PHP8.0 is our primary version of PHP that's running
*and*
we have an isolated site on PHP8.0

They'll be pointing to different socket files that are both intending to represent the same PHP connection
And I think they don't like it. And I think that's why my `valet.sock` keeps disappearing. I'll be digging into it to see.

mattstauffer:
Solved this problem: https://github.com/laravel/valet/pull/1201#issuecomment-1066883737

I think at this point this PR is ready for review.

shengslogar:
This is 🔥! Won't pretend to know the ins-and-outs of how this works, but how hard would it be to add a CLI util to access that running PHP version? (Thinking along the lines of `php artisan tinker`...)

NasirNobin:
@shengslogar that's coming soon! checkout https://github.com/laravel/valet/pull/1216

> This is 🔥! Won't pretend to know the ins-and-outs of how this works, but how hard would it be to add a CLI util to access that running PHP version? (Thinking along the lines of `php artisan tinker`...)

PR #1204, opened 2022-03-15
Bump valet version for CLI

joelbutcher:
As of the `3.0.0-alpha` release yesterday `valet -V` returns the wrong value. This PR updates the global version for Valet. **Not:** this will likely need to be updated before tagging the next release (probably either as `3.0.0-alpha-2` or something similar)

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -32,7 +32,7 @@ if (is_dir(VALET_LEGACY_HOME_PATH) && ! is_dir(VALET_HOME_PATH)) {
  */
 Container::setInstance(new Container);
 
-$version = '2.18.10';
+$version = '3.0.0-alpha';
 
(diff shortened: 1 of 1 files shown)


mattstauffer:
No need for this PR, i'll update it myself for the next tag, but thanks for the note. Honestly I just didn't care to update it because I didn't think anyone would mind, but I was obviously wrong; my bad!

Out of curiosity, did you have any specific reason to want to check the version? Either way I'll update it for the next tag; I am just curious.

joelbutcher:
@mattstauffer oromarily for verifying that I have the correct version installed when testing the new isolation feature

driesvints:
@joelbutcher new alpha release should have the correct version now 👍

joelbutcher:
Awesome, cheers @driesvints

Issue #1205, opened 2022-03-15
Trouble with Valet 3.0 alpha due to Homebrew PHP formula's default `www` pools


nicoverbruggen:
Hey @mattstauffer and @NasirNobin,

First off, my apologies for this wall of text 😅...

I've been trying out the new changes in Valet 3.0's alpha, and I have started work on integrating site isolation into PHP Monitor. I absolutely love that you guys were able to add this to Valet — it's the one thing I've been missing all this time. I'm grateful you were able to add this to Valet. ❤️

I did run into a particular issue on my system that presents itself when multiple conflicting PHP services are running, and I wanted to report this — or at least see if you are having the same issue(s). I'm not entirely positive if this is a "oops, it's happening on my system" kinda issue or if it's going to be a common issue.

I believe that the default installation of a PHP formula creates a `www` pool configuration file in `/opt/homebrew/etc/php/{version}/php-fpm.d/www.conf`. (Correct me if I'm wrong — I found these files on my system, at least. I think this _may_ be a more recent addition, I'm not sure. Can you confirm that you have these files on your system as well?)

Anyway, my issue seems to stem from these `www` pool configuration files.

Since multiple PHP-FPM instances might be running if isolated sites are configured, it is possible that due to this default file, the various `www` pools are created, all trying to bind to the default port 9000. (Each PHP-FPM instance spins up all pools set up in `/opt/homebrew/etc/php/{version}/php-fpm.d`. Note that on Intel systems your Homebrew folder might be in a different spot.)

Only one PHP version's `www` pool will "win" port 9000 in this case, and all others will silently fail in the background. As a result, in `sudo brew services list` these services will now report themselves as failing. Note that Valet should, at this point in time, remain functional, since its `valet-fpm.conf` pools are not affected... (yet).

I think it might be prudent to disable these default `www` pools now that Valet uses multiple PHP versions, because... due to Homebrew thinking that these PHP-FPM `www` pools conflict, it is possible that some of the commands that Valet needs to execute, fail (ones that involve `brew services`). 

As far as I can tell, on my system, the various PHP-FPM processes did not spin up the respective `valet` pools if one of the `www` pools failed to bind to port 9000. This doesn't always seem to happen, so I'm guessing the order in which the pools are started is random or perhaps the pools are started in parallel? The point is, sometimes the Valet socket won't work because the PHP-FPM pool created for Valet won't start for some reason (again, I'm guessing it's due to the port conflict).

[...]

PR #1208, opened 2022-03-16
Fix Homebrew PHP default www pools issue on Valet 3.0 alpha

NasirNobin:
Previously Valet was disabling `www.conf` during the PHP version installation process. 

But [these lines of codes](https://github.com/laravel/valet/commit/eefc06b07f8ee57186ce898cacaaf33f9504d679#diff-d8e7b9c98bc6a045292f784be0d98860c26b463a1dd29f36256fed1d2ad62752L89) was probably deleted accidentally. Or intentionally? I'm not sure. 

Either way, I think bringing these lines back should fix https://github.com/laravel/valet/issues/1205.

### Steps to reproduce this error
1. Install a PHP version using `brew install php@x.x` or `valet use php@x.x` that is not present on the machine.
2. Homebrew would create a `/etc/php/x.x/php-fpm.d/www.conf` file automatically during the installation process.
3. That should create a conflict and would result in a 502 error in valet sites.

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -92,6 +92,12 @@ class PhpFpm
 
         $this->files->ensureDirExists(dirname($fpmConfigFile), user());
 
+        // rename (to disable) old FPM Pool configuration, regardless of whether it's a default config or one customized by an older Valet version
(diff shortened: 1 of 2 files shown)


mattstauffer:
It was absolutely a mistake on my part. Thanks so much for the quick fix!

PR #1213, opened 2022-03-21
Isolation compatibility with .valetphprc

### LINKED ISSUE #1209: `valet use`, when run in an isolated directory, should pull that directory's isolated PHP version
https://twitter.com/JacobDelcroix/status/1503887481149796352

Currently `valet use` , when run in a directory with a `.valetphprc` file, pulls the PHP version from that file and uses it to change the global PHP Valet to that version.

It should also (this is a feature request, not a bug, but I'm the maintainer so I get that privilege of using issues this way bwahahahaha) read from the isolated PHP version.

NasirNobin:
This PR adds two enhancements: 

1. Running `valet isolate` (without specifying a PHP version) on a directory where there's a `.valetphprc` file is present, Valet would recognize that. 
2. Running `valet use` (without specifying a PHP version) on an isolated site directory would change global PHP version to that isolated version. 

Fixes #1209

Tests: I'm not sure how do we test codes from `cli/valet.php` yet. I would be happy to add tests if I can get some guidance.

### DIFF
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -540,8 +540,7 @@ You might also want to investigate your global Composer configs. Helpful command
             $site = basename(getcwd());
         }
 
-        if (! $phpVersion) {
-            $phpVersion = Site::phpRcVersion($site);
+        if (is_null($phpVersion) && $phpVersion = Site::phpRcVersion($site)) {
             info("Found '{$site}/.valetphprc' specifying version: {$phpVersion}");
         }
 
--- a/cli/valet.php
+++ b/cli/valet.php
@@ -508,7 +508,15 @@ You might also want to investigate your global Composer configs. Helpful command
         if (! $phpVersion) {
             $site = basename(getcwd());
             $linkedVersion = Brew::linkedPhp();
-            $phpVersion = Site::phpRcVersion($site);
+
+            if ($phpVersion = Site::phpRcVersion($site)) {
+                info("Found '{$site}/.valetphprc' specifying version: {$phpVersion}");
+            } else {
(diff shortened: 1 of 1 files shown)


cli/valet.php
```diff
@@ -508,14 +508,25 @@
...
             }
 
-            $phpVersion = trim(file_get_contents($path));
-            info("Found '{$path}' specifying version: {$phpVersion}");
+            if (! $phpVersion) {
+                return info("Valet is using {$linkedVersion}.");
+            }
 
-            if ($linkedVersion == $phpVersion) {
+            if ($linkedVersion == $phpVersion && ! $force) {
```
NasirNobin, line 529: Previously, it wasn't respecting `--force` when comparing version against `.valetphprc`. I thought it would do that.

cli/valet.php
```diff
@@ -534,6 +540,11 @@
             $site = basename(getcwd());
         }
 
+        if (! $phpVersion) {
+            $phpVersion = Site::phpRcVersion($site);
+            info("Found '{$site}/.valetphprc' specifying version: {$phpVersion}");
+        }
```
NasirNobin, line 546:
```suggestion
        if (is_null($phpVersion) && $phpVersion = Site::phpRcVersion($site)) {
            info("Found '{$site}/.valetphprc' specifying version: {$phpVersion}");
        }
```

mattstauffer:
> Tests: I'm not sure how do we test codes from cli/valet.php yet. I would be happy to add tests if I can get some guidance.

There aren't any yet, but I've been meaning to add some. We don't need to hold up this PR for that but I'll make an issue.

mattstauffer:
Great work as always @NasirNobin!

PR #1216, opened 2022-03-22
PHP version isolation helper for command line

NasirNobin:
This PR introduces 3 new valet commands.  These will help users to run PHP and Composer commands with isolated site's PHP version on the CLI. 

1. `valet php ...` will proxy PHP Commands with isolated PHP version
2. `valet composer ...` will proxy Composer Commands with isolated PHP version
3. `valet which-php` outputs the PHP executable path for a site. For isolated site it would output the isolated PHP executable path. But non-isolated site will just output the linked default PHP path. The other two commands are dependent on this one to find the PHP executable. 

<details>
<summary> 🔽 Video demonstration </summary>

https://user-images.githubusercontent.com/13833460/159581055-4c1b7494-feb5-420a-946f-f89512597cc3.mp4
</details>

## Syntax
Though the [original idea](https://github.com/laravel/valet/discussions/1211) was to create a command like `valet run` that would proxy PHP calls to an isolated PHP version. But I kinda like the [laravel sail](https://github.com/laravel/sail/blob/1.x/bin/sail) style, so kept it this way. 

## Quick note on `which-php` command
Currently `valet which-php` uses PHP to find the correct executable path. Initially, I thought of doing this with vanilla bash. But then thought if it would bring any benefit, also would make it hard to test it. Some of the [codes](https://github.com/NasirNobin/valet/pull/8/files) are still here, in case if we want to explore that route. 

<details>
<summary>Symlinking PHP executables to make it a bit fast (No longer using it)</summary>

~~It's using a combination of `brew --cellar` and `brew info` commands to determine the correct PHP executable path, So each time user runs `valet php -v` it would go through the process of extracting the PATH, which is very slow. Ended up adding a caching mechanism that creates a symlink at `/opt/homebrew/bin/valetphp81` location, that directs toward the actual path for that version.~~
~~So next time a user runs `valet php -v` it would resolve the PHP path much faster from the symlink.~~
~~Please do suggest if you think there's an easier way to find the PHP executable for a given PHP version. I'll be happy to refactor.~~
</details>

## Update:
Refactored it to directly look for `BREW_PREFIX/opt/php@8.0/bin/php` location to find the executable binary file. It's already fast and convenient. So we are not creating any additional symlinks now. Thanks to @nicoverbruggen for the suggestion.

## Alternative Solution
Before starting to work on this PR, I was working on another external solution called [zsh-valet](https://github.com/NasirNobin/zsh-valet) to solve the very same problem in a more automated way, still in progress. I'll try to manage that as an external zsh package. 

--

Thanks to my colleague @rana01645, @MishukAdhikari & @shifat160 for helping me test this PR properly on both M1 and Intel Mac.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -251,16 +251,7 @@ class Brew
 
         $resolvedPath = $this->files->readLink(BREW_PREFIX.'/bin/php');
 
-        /**
-         * Typical homebrew path resolutions are like:
-         * "../Cellar/php@7.4/7.4.13/bin/php"
-         * or older styles:
-         * "../Cellar/php/7.4.9_2/bin/php
-         * "../Cellar/php55/bin/php.
-         */
-        preg_match('~\w{3,}/(php)(@?\d\.?\d)?/(\d\.\d)?([_\d\.]*)?/?\w{3,}~', $resolvedPath, $matches);
-
-        return $matches;
+        return $this->parsePhpPath($resolvedPath);
     }
 
     /**
@@ -289,48 +280,10 @@ class Brew
 
         return $this->supportedPhpVersions()->first(
             function ($version) use ($resolvedPhpVersion) {
-                $resolvedVersionNormalized = preg_replace('/[^\d]/', '', $resolvedPhpVersion);
-                $versionNormalized = preg_replace('/[^\d]/', '', $version);
-
-                return $resolvedVersionNormalized === $versionNormalized;
+                return $this->isPhpVersionsEqual($resolvedPhpVersion, $version);
             }, function () use ($resolvedPhpVersion) {
                 throw new DomainException("Unable to determine linked PHP when parsing '$resolvedPhpVersion'");
             });
-    }
-
-    /**
-     * Get PHP executable path for a given PHP Version.
-     *
-     * @param  string|null  $phpVersion
-     * @param  bool  $skipCache
-     * @return string
-     */
-    public function whichPhp($phpVersion = null, $skipCache = false)
-    {
-        if (! $phpVersion) {
-            return BREW_PREFIX.'/bin/php';
-        }
-
-        $versionInteger = preg_replace('~[^\d]~', '', $phpVersion);
-        $symlinkedValetPhpPath = BREW_PREFIX."/bin/valetphp{$versionInteger}";
-
-        // If the symlinked Valet PHP path exists, then we can use that
-        if (! $skipCache && $this->files->isLink($symlinkedValetPhpPath)) {
-            $phpExecutablePath = $this->files->readLink($symlinkedValetPhpPath);
-
-            // Still make sure that the version of the executable exists
-            if ($this->files->exists($phpExecutablePath)) {
-                return $phpExecutablePath;
-            }
-        }
-
-        // If the symlinked Valet PHP path doesn't exist, then we need to look for the correct executable path
-        // Create a symlink to the Valet PHP version, so next time Valet won't have to look for the executable path
-        if ($phpExecutablePath = $this->getPhpExecutablePath($phpVersion)) {
-            $this->files->symlinkAsUser($phpExecutablePath, $symlinkedValetPhpPath);
-        }
-
-        return $phpExecutablePath ?: BREW_PREFIX.'/bin/php';
     }
 
     /**
@@ -339,32 +292,34 @@ class Brew
      * @param  string  $phpVersion
      * @return string
      */
-    public function getPhpExecutablePath($phpVersion)
+    public function getPhpExecutablePath($phpVersion = null)
     {
-        $phpExecutablePath = null;
+        if (! $phpVersion) {
+            return BREW_PREFIX.'/bin/php';
+        }
 
-        $cellar = $this->cli->runAsUser("brew --cellar $phpVersion"); // Example output: `/opt/homebrew/Cellar/php@8.0`
-        $details = json_decode($this->cli->runAsUser("brew info --json $phpVersion"), true);
-        $phpDirectory = data_get($details, '0.linked_keg');
+        // Check the default `/opt/homebrew/opt/php@8.1/bin/php` location first
+        if ($this->files->exists(BREW_PREFIX."/opt/{$phpVersion}/bin/php")) {
+            return BREW_PREFIX."/opt/{$phpVersion}/bin/php";
+        }
 
-        if (is_null($phpDirectory) && $installed = data_get($details, '0.installed')) {
-            $phpDirectory = data_get(collect($installed)->where('installed_as_dependency', false)->last(), 'version');
+        // Check the `/opt/homebrew/opt/php71/bin/php` location for older installations
+        $phpVersion = str_replace(['@', '.'], '' , $phpVersion); // php@8.1 to php81
+        if ($this->files->exists(BREW_PREFIX."/opt/{$phpVersion}/bin/php")) {
+            return BREW_PREFIX."/opt/{$phpVersion}/bin/php";
+        }
 
-            if (is_null($phpDirectory)) {
-                $phpDirectory = data_get(collect($installed)->last(), 'version');
+        // Check if the default PHP is the version we are looking for
+        if ($this->files->isLink(BREW_PREFIX."/opt/php")) {
+            $resolvedPath = $this->files->readLink(BREW_PREFIX."/opt/php");
+            $matches = $this->parsePhpPath($resolvedPath);
+            $resolvedPhpVersion = $matches[3] ?: $matches[2];
+            if ($this->isPhpVersionsEqual($resolvedPhpVersion, $phpVersion)) {
+                return BREW_PREFIX."/opt/php/bin/php";
             }
         }
 
-        if (isset($phpDirectory) && $this->files->exists(trim($cellar).'/'.$phpDirectory.'/bin/php')) {
-            $phpExecutablePath = trim($cellar).'/'.$phpDirectory.'/bin/php';
-        }
-
-        // When PHP Version is installed directly though shivammathur/homebrew-php, it usually stores the binaries in this directory
-        if (is_null($phpExecutablePath) && $this->files->exists(BREW_PREFIX."/opt/{$phpVersion}/bin/php")) {
-            $phpExecutablePath = BREW_PREFIX."/opt/{$phpVersion}/bin/php";
-        }
-
-        return $phpExecutablePath;
+        return BREW_PREFIX.'/bin/php';
     }
 
     /**
@@ -531,5 +486,42 @@ class Brew
                 output($errorOutput);
             }
         );
+    }
+
+    /**
+     * Parse homebrew PHP Path
+     *
+     * @param  string  $resolvedPath
+     *
+     * @return mixed
+     */
+    public function parsePhpPath($resolvedPath)
+    {
+        /**
+         * Typical homebrew path resolutions are like:
+         * "../Cellar/php@7.4/7.4.13/bin/php"
+         * or older styles:
+         * "../Cellar/php/7.4.9_2/bin/php
+         * "../Cellar/php55/bin/php.
+         */
+        preg_match('~\w{3,}/(php)(@?\d\.?\d)?/(\d\.\d)?([_\d\.]*)?/?\w{3,}~', $resolvedPath, $matches);
+
+        return $matches;
+    }
+
+    /**
+     * Check if two PHP versions are equal
+     *
+     * @param string $resolvedPhpVersion
+     * @param string $version
+     *
+     * @return bool
+     */
+    public function isPhpVersionsEqual($resolvedPhpVersion, $version)
+    {
+        $resolvedVersionNormalized = preg_replace('/[^\d]/', '', $resolvedPhpVersion);
+        $versionNormalized = preg_replace('/[^\d]/', '', $version);
+
+        return $resolvedVersionNormalized === $versionNormalized;
     }
 }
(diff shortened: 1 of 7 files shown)


cli/Valet/Brew.php
```diff
@@ -298,6 +298,75 @@ function ($version) use ($resolvedPhpVersion) {
...
+                $phpDirectory = data_get(collect($installed)->last(), 'version');
+            }
+        }
+
+        if (isset($phpDirectory) && $this->files->exists(trim($cellar).'/'.$phpDirectory.'/bin/php')) {
+            $phpExecutablePath = trim($cellar).'/'.$phpDirectory.'/bin/php';
+        }
+
+        // When PHP Version is installed directly though shivammathur/homebrew-php, it usually stores the binaries in this directory
+        if (is_null($phpExecutablePath) && $this->files->exists(BREW_PREFIX."/opt/{$phpVersion}/bin/php")) {
```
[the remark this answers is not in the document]
NasirNobin: On the [initial commit](https://github.com/laravel/valet/pull/1216/commits/794fbc0c577931494ca9b16fe6ec5df24098c580#diff-679e80b2a36f0b50866ae7b3d72d113ca4f2e647fa1a21bdc15255cb19d4ac65R309) I was simply looking for `BREW_PREFIX/opt/php@8.0/bin/php`, and returning it if that exists. 

But for some reason, I found that the symlink on that location was not pointing to the correct (updated) PHP binary when I was running tests on my machine. So kinda choose to rewrite this to check the other way around. Which was a mistake on my tests now I'm seeing it again. 

I'll re-run tests and will refactor to this way. It should make the whole thing a lot simpler than currently it is. Thanks a lot! @nicoverbruggen

cli/Valet/Brew.php
```diff
@@ -463,4 +487,39 @@ function ($exitCode, $errorOutput) {
             }
         );
     }
+
+    /**
+     * Parse homebrew PHP Path.
+     *
+     * @param  string  $resolvedPath
+     * @return mixed
```
NasirNobin, line 495:
```suggestion
     * @return array
```

cli/Valet/Brew.php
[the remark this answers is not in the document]
NasirNobin: @nicoverbruggen updated the PR. If you are available and interested please feel free to re-review the PR. Thanks!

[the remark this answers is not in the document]
nicoverbruggen: This looks great to me, nice! 👌

cli/Valet/Brew.php
```diff
@@ -289,15 +280,48 @@ public function linkedPhp()
...
+                return $this->isPhpVersionsEqual($resolvedPhpVersion, $version);
             }, function () use ($resolvedPhpVersion) {
                 throw new DomainException("Unable to determine linked PHP when parsing '$resolvedPhpVersion'");
             });
     }
 
+    /**
+     * Extract PHP executable path from PHP Version.
+     *
+     * @param  string  $phpVersion
```
mattstauffer, line 292: Could we specify this should be a `php@8.1` style, since sometimes we're only passing around `8.1`?

```suggestion
     * @param  string  $phpVersion  For example, "php@8.1"
```

cli/valet.php
```diff
@@ -561,6 +561,45 @@
...
+        if (! $phpVersion) {
+            $path = getcwd().'/.valetphprc';
+            if (file_exists($path)) {
+                $phpVersion = trim(file_get_contents($path));
+            }
+        }
+
+        $phpVersion = $phpVersion ? PhpFpm::normalizePhpVersion($phpVersion) : null;
+
+        return output(Brew::getPhpExecutablePath($phpVersion));
```
mattstauffer, line 580: I don't think the ternary check is necessary here, because you already checked it for false-ness on line 571.

Also.. what if we run `normalizePhpVersion` at the top of `getExecutablePath`? Then A) you don't have to run it here or anywhere else, and B) you don't have to add my suggested line in the comments there

```suggestion
        return output(Brew::getPhpExecutablePath(PhpFpm::normalizePhpVersion($phpVersion)));
```

cli/valet.php
```diff
@@ -561,6 +561,45 @@
...
+    })->descriptions('Get the PHP executable path for a given site', [
+        'site' => 'The site to get the PHP executable path for',
+    ]);
+
+    /**
+     * Proxy PHP commands with isolated site's  PHP executable file.
```
mattstauffer, line 586:
```suggestion
     * Proxy commands through to an isolated site's version of PHP.
```

cli/valet.php
```diff
@@ -561,6 +561,45 @@
...
+     */
+    $app->command('php [command]', function ($command) {
+        warning('It looks like you are running `cli/valet.php` directly; please use the `valet` script in the project root instead.');
+    })->descriptions("Proxy PHP commands with isolated site's PHP executable", [
+        'command' => "Command to run with isolated site's PHP executable",
+    ]);
+
+    /**
+     * Proxy Composer commands with isolated site's PHP executable.
```
mattstauffer, line 595:
```suggestion
     * Proxy commands through to an isolated site's version of Composer.
```

valet
```diff
@@ -81,16 +75,36 @@ then
...
+elif [[ "$1" = "php" ]]
+then
+    $(php "$DIR/cli/valet.php" which-php) "${@:2}"
+
+    exit
+
+# Proxy Composer commands with the "php" executable on the isolated site
+elif [[ "$1" = "composer" ]]
+then
+    $(php "$DIR/cli/valet.php" which-php) $(which composer) "${@:2}"
```
mattstauffer, line 95: Are we going to run into any trouble where we're running a lower version of PHP than Composer is globally updated to expect? If so, what's the error we see?

valet
```diff
@@ -81,16 +75,36 @@ then
     ARCH=$(uname -m)
 
     if [[ $ARCH == 'arm64' ]]; then
-        sudo -u "$USER" "$DIR/bin/ngrok-arm" http "$HOST.$TLD:$PORT" -host-header=rewrite $PARAMS
+        "$DIR/bin/ngrok-arm" http "$HOST.$TLD:$PORT" -host-header=rewrite $PARAMS
     else
-        sudo -u "$USER" "$DIR/bin/ngrok" http "$HOST.$TLD:$PORT" -host-header=rewrite $PARAMS
+        "$DIR/bin/ngrok" http "$HOST.$TLD:$PORT" -host-header=rewrite $PARAMS
```
mattstauffer, line 80: Could you explain to me what the change to these lines, and also moving lines 23:28, are for? Sorry, i don't full understand either, and I figured it'll be faster for you to share than for me to dig into it on my own. Thanks!

cli/Valet/Brew.php
```diff
@@ -463,4 +487,39 @@ function ($exitCode, $errorOutput) {
...
+     * @param  string  $resolvedPhpVersion
+     * @param  string  $version
+     * @return bool
+     */
+    public function isPhpVersionsEqual($resolvedPhpVersion, $version)
+    {
+        $resolvedVersionNormalized = preg_replace('/[^\d]/', '', $resolvedPhpVersion);
+        $versionNormalized = preg_replace('/[^\d]/', '', $version);
+
+        return $resolvedVersionNormalized === $versionNormalized;
```
mattstauffer, line 523:
```suggestion
     * @param  string  $versionA
     * @param  string  $versionB
     * @return bool
     */
    public function arePhpVersionsEqual($versionA, $version)
    {
        $versionANormalized = preg_replace('/[^\d]/', '', $versionA);
        $versionBNormalized = preg_replace('/[^\d]/', '', $versionB);

        return $versionANormalized === $versionBNormalized;
```

cli/Valet/Brew.php
```diff
@@ -289,15 +280,48 @@ public function linkedPhp()
...
+        if ($this->files->exists(BREW_PREFIX."/opt/{$phpVersion}/bin/php")) {
+            return BREW_PREFIX."/opt/{$phpVersion}/bin/php";
+        }
+
+        // Check if the default PHP is the version we are looking for
+        if ($this->files->isLink(BREW_PREFIX.'/opt/php')) {
+            $resolvedPath = $this->files->readLink(BREW_PREFIX.'/opt/php');
+            $matches = $this->parsePhpPath($resolvedPath);
+            $resolvedPhpVersion = $matches[3] ?: $matches[2];
+            if ($this->isPhpVersionsEqual($resolvedPhpVersion, $phpVersion)) {
```
mattstauffer, line 317:
```suggestion

            if ($this->arePhpVersionsEqual($resolvedPhpVersion, $phpVersion)) {
```

cli/Valet/Brew.php
```diff
@@ -289,15 +280,48 @@ public function linkedPhp()
 
         return $this->supportedPhpVersions()->first(
             function ($version) use ($resolvedPhpVersion) {
-                $resolvedVersionNormalized = preg_replace('/[^\d]/', '', $resolvedPhpVersion);
-                $versionNormalized = preg_replace('/[^\d]/', '', $version);
-
-                return $resolvedVersionNormalized === $versionNormalized;
+                return $this->isPhpVersionsEqual($resolvedPhpVersion, $version);
```
mattstauffer, line 283:
```suggestion
                return $this->arePhpVersionsEqual($resolvedPhpVersion, $version);
```

tests/BrewTest.php
```diff
@@ -395,6 +395,71 @@ public function test_restart_linked_php_will_pass_through_linked_php_formula_to_
...
+    }
+
+    public function test_it_can_compare_two_php_versions()
+    {
+        $this->assertTrue(resolve(Brew::class)->isPhpVersionsEqual('php71', 'php@7.1'));
+        $this->assertTrue(resolve(Brew::class)->isPhpVersionsEqual('php71', 'php@71'));
+        $this->assertTrue(resolve(Brew::class)->isPhpVersionsEqual('php71', '71'));
+
+        $this->assertFalse(resolve(Brew::class)->isPhpVersionsEqual('php71', 'php@70'));
+        $this->assertFalse(resolve(Brew::class)->isPhpVersionsEqual('php71', '72'));
```
mattstauffer, line 460:
```suggestion
        $this->assertTrue(resolve(Brew::class)->arePhpVersionsEqual('php71', 'php@7.1'));
        $this->assertTrue(resolve(Brew::class)->arePhpVersionsEqual('php71', 'php@71'));
        $this->assertTrue(resolve(Brew::class)->arePhpVersionsEqual('php71', '71'));

        $this->assertFalse(resolve(Brew::class)->arePhpVersionsEqual('php71', 'php@70'));
        $this->assertFalse(resolve(Brew::class)->arePhpVersionsEqual('php71', '72'));
```

mattstauffer (changes requested):
Fantastic work, as always. A few notes/requests :)

valet
NasirNobin:
```bash
if [[ "$EUID" -ne 0 ]]
then
    sudo USER="$USER" --preserve-env "$SOURCE" "$@"
    exit
fi
```

Occurred to me, we just need these lines for this particular PHP call `php "$DIR/cli/valet.php" "$@"`. Other commands from this file don't require a root user. So, We moved this chunk of code from the top of this file to the bottom.

Also, we definetly don't want to proxy `php` and `composer` commands with the root user. So moving that line to the bottom kinda made sense to me. 

Now that we moved that line to the bottom. `valet share`, `valet php` and `valet composer` commands should not come from `root` user. So this `sudo -u "$USER"` shouldn't be necessary anymore. 

--

btw, please do let me know, if you think we should be explicit about the `user` we are using to run all these commands, I'll undo this and update my PR accordingly.

valet
NasirNobin: In that case, **any valet command** would fail with a similar error.  

```bash
valet -V

Fatal error: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.1.0". You are running 7.0.33. in /Users/user/.composer/vendor/composer/platform_check.php on line 24
```

NasirNobin: btw, if you are asking about the other way around. Like if isolated `php@7.0` would be able to run composer commands when the global composer is updated to `php@8.1`. In that case, I can see it works fine.

valet
mattstauffer: No, that makes a ton of sense--I just have never *fully* followed which user we need in which context, so it's helpful to have your explanation. :) Thank you!

cli/valet.php
```diff
@@ -561,6 +561,43 @@
...
 
+    /**
+     * Get the PHP executable path for a site.
+     */
+    $app->command('which-php [site]', function ($site) {
+        $host = Site::host($site ?: getcwd()).'.'.Configuration::read()['tld'];
+        $phpVersion = Site::customPhpVersion($host);
+
+        if (! $phpVersion) {
+            $path = getcwd().'/.valetphprc';
```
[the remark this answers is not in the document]
NasirNobin: Will fix the `$site` specific `.valetphprc` issue. 

Also I'm very much agreed to the `Site::phpRcVersion($sitename)` method. In fact, I even have [another PR](https://github.com/laravel/valet/pull/1213/files) that kinda has these lines as duplicates.

cli/valet.php
```diff
@@ -533,6 +534,11 @@
             $site = basename(getcwd());
         }
 
+        if (! $phpVersion) {
```
mattstauffer, line 537: Could we save this change for a separate PR? I think you're already building it in #1213, right?

mattstauffer (changes requested):
Two minor changes and then I think we're good to go!

cli/valet.php
```diff
@@ -533,6 +534,11 @@
...
+            $phpVersion = Site::phpRcVersion($site);
+            info("Found '{$site}' specifying version: {$phpVersion}");
+        }
```
NasirNobin, line 540:
```suggestion

```

mattstauffer:
Thanks so much for your work here @NasirNobin!

Issue #771, opened 2019-04-19
Add Drupal 8.7 JSON:API support


chrfritsch:
When currently requesting jsonapi with a GET on https://thunder.test/jsonapi/node/article, I get the following error.

```
{
    "jsonapi": {
        "version": "1.0",
        "meta": {
            "links": {
                "self": {
                    "href": "http://jsonapi.org/format/1.0/"
                }
            }
        }
    },
    "errors": [
        {
            "title": "Bad Request",
            "status": "400",
            "detail": "The following query parameters violate the JSON:API spec: 'q'.",
            "links": {
                "via": {
                    "href": "https://thunder.test/jsonapi/node/article"
                },
                "info": {
                    "href": "http://jsonapi.org/format/#query-parameters"
                }
            }
        }
    ]
}
```

I tried to apply the fix from https://github.com/weprovide/valet-plus/issues/268, but that didn't work for me.

drogers98:
Also hitting this issue, with no fix yet.

drogers98:
So, not optimum, but changing line 60-62 of DrupalValetDriver.php from:
`
if (!isset($_GET['q']) && !empty($uri) && $uri !== '/') {
          $_GET['q'] = $uri;
        }
`
to:
`
if (!isset($_GET['Q']) && !empty($uri) && $uri !== '/') {
          $_GET['Q'] = $uri;
        }
`
Fixed this for me. Think it is to do with jsonapi expecting camel/upper case. Not 100% sure what else this would affect, if any, so not submitting a patch, but this did the trick for me locally for now.

remember to run "valet restart" and clear drupal cache, if you go this route.

wimleers:
It sounds like Laravel Valet is rewriting `example.com/foobar` to `example.com/index.php?q=foobar` — i.e. it sounds like it's causing routing to not use clean URLs anymore internally?

dustinleblanc:
I've temporarily made my local driver look like this:
```php
<?php

class LocalValetDriver extends DrupalValetDriver
{
    /**
     * Get the fully resolved path to the application's front controller.
     *
     * @param  string  $sitePath
     * @param  string  $siteName
     * @param  string  $uri
     * @return string
     */
    public function frontControllerPath($sitePath, $siteName, $uri)
    {
        $sitePath = $this->addSubdirectory($sitePath);
        $matches = [];
        if (preg_match('/^\/(.*?)\.php/', $uri, $matches)) {
            $filename = $matches[0];
[...]

PR #1218, opened 2022-03-24
fix: Drupal JSON:API support for non-standard url

BurningDog:
This is related to issue https://github.com/laravel/valet/issues/771 which was fixed by https://github.com/laravel/valet/pull/705

Unfortunately, as [this comment in the PR shows](https://github.com/laravel/valet/pull/705/files#r284138311), this will break for anyone not using the default path setting for [Drupal's JSON:API module](https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module/api-overview), like me. In Drupal I changed the default `/jsonapi` path to `/api` and this broke Valet's integration.

Specifically, when making any API call to a JSON:API endpoint, I got the same error as in the original issue:

```
The following query parameters violate the JSON:API spec: 'q'.
```

In the original issue there was already [a fix proposed](https://github.com/laravel/valet/issues/771#issuecomment-485533478), which is to change the following in `DrupalValetDriver.php`:

`if (!isset($_GET['q']) && !empty($uri) && $uri !== '/') { $_GET['q'] = $uri; }`
to:
`if (!isset($_GET['Q']) && !empty($uri) && $uri !== '/') { $_GET['Q'] = $uri; }`

That's all this current PR does. It doesn't break anything existing, as Drupal otherwise does not care about the case-sensitivity of the `q` querystring parameter.

### Alternative solutions

There are a few other possibilities mentioned in https://www.drupal.org/project/jsonapi/issues/2984044 but they require patching Drupal core, or having Nginx config specific to Drupal (which isn't applicable to Valet).

Out of interesting, here's the official [Nginx config for Drupal](https://www.nginx.com/resources/wiki/start/topics/recipes/drupal/)

### DIFF
--- a/cli/drivers/DrupalValetDriver.php
+++ b/cli/drivers/DrupalValetDriver.php
@@ -57,8 +57,8 @@ class DrupalValetDriver extends ValetDriver
     {
         $sitePath = $this->addSubdirectory($sitePath);
 
-        if (! isset($_GET['q']) && ! empty($uri) && $uri !== '/' && strpos($uri, '/jsonapi/') === false) {
-            $_GET['q'] = $uri;
+        if (! isset($_GET['Q']) && ! empty($uri) && $uri !== '/' && strpos($uri, '/jsonapi/') === false) {
+            $_GET['Q'] = $uri;
         }
 
         $matches = [];


mattstauffer:
It's wild that this solves it, but, thanks for the PR!! :D

EustaceCheng:
I upgrade valet to 3.1.3, and rebuild drupal, it works. thanks!
Is there has preview before submit add-content form?

Issue #1220, opened 2022-03-28
valet change folder ownership


PATROMO:
After fresh brew php installation:
<img width="587" alt="image" src="https://user-images.githubusercontent.com/318564/160414248-a424f367-0484-4d87-93d7-18797e84e1b9.png">

After valet usage:
<img width="614" alt="image" src="https://user-images.githubusercontent.com/318564/160414313-680a1b61-3554-4bd3-a1ba-1f2852d89eff.png">

This causes problems when upgrading the php version in the future. 
<img width="427" alt="image" src="https://user-images.githubusercontent.com/318564/160414977-0dcc05ac-4b37-4e42-ae12-8f6f539fafaa.png">

nicoverbruggen:
Hi, all! 

I wanted to chime in on this one, since I've [written about this](https://github.com/nicoverbruggen/phpmon/issues/85) as well when I was dealing with this issue in PHP Monitor and even shipped a feature/workaround that "fixes" this, called "[Fix Homebrew Permissions](https://github.com/nicoverbruggen/phpmon/issues/86)".

I don't think anything can be done about the ownership (it's a consequence of running `php` as `root`). However, reinstating the original permissions when you run `valet stop` might be a good idea. 

If implemented, you could then feasibly run `valet stop && brew upgrade php`, for example, without Homebrew then creating `.reinstall` folders, which would be awesome.

### Technical details

In phpmon, I restore the ownership like so...

`chown -R <username>:staff /path/to/directory` (default Homebrew permissions give the current user (accessible via `whoami`, and the `wheel` group access IIRC)

... for the following directories:

```swift
let directories = [
    "\(brewDir)/Cellar/\(formula)"
    "\(brewDir)/opt/\(formula)"
    "\(brewDir)/var/homebrew/linked/\(formula)"
];
```

So this has to happen for all installed PHP versions/formulae as well as for `nginx` and `dnsmasq`. I was going to take a look at this one myself and open a PR but I haven't had the time yet, unfortunately.

driesvints:
@PATROMO What are the reasons for running php as root? (Thanks for chiming in @nicoverbruggen)

nicoverbruggen:
I believe running all those services as `root` is required in order to successfully bind to port 80 on localhost, IIRC.

PATROMO:
@nicoverbruggen thanks for explaining :) 
Yes, we need root to bind port 80. But I think that the folders should not change the ownership. :)

[...]

PR #1223, opened 2022-03-29
When stopping Homebrew services, restore original folder permissions

nicoverbruggen:
As discussed in #1220, cleaning up after stopping Homebrew services that run as `root` is probably the best solution for now. This PR now causes permissions to be reset whenever a Homebrew service is stopped.

What's changed:

- Stopping any Homebrew service now restores the appropriate permissions for the Homebrew directories
- Stopping nginx now also happens via $this->brew (to avoid code duplication)

The tests were updated to reflect this change.

There is one thing I'm uncertain of — whether the group owner needs to be `wheel` or `admin`. Either is fine, so in these commits I opted to go for `admin`, since that seems to be the default group ownership Homebrew assigns by default.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -232,7 +232,6 @@ class Brew
                     BREW_PREFIX . "/var/homebrew/linked/$service",
                 ];
 
-                // if the services aren't running: we'll restore the original permissions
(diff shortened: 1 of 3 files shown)


mattstauffer:
Fantastic--thanks @nicoverbruggen!

PR #1229, opened 2022-04-03
Add PHP version to `valet links` output table

michaellehmkuhl:
Adds a new column to the `valet links` CLI command output to show which PHP is version running on each site. This gives users who run multiple sites with isolated PHP versions on each a quick view of which PHP version is running on each site.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -7,6 +7,7 @@ use PhpFpm;
 
 class Site
 {
+    public $brew;
     public $config;
     public $cli;
     public $files;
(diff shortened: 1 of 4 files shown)


NasirNobin:
There's a dedicated command `valet isolated` that lists all isolated sites with the PHP versions. I'm guessing this PR will be a duplicate of that. 

CleanShot 2022-04-04 at 10 43 48

michaellehmkuhl:
@NasirNobin it's true that `valet isolated` exists, but it shows only a list of isolated sites, rather than a full list of all sites like this PR.

mattstauffer:
Thanks @michaellehmkuhl!!

PR #941, opened 2020-04-28
feat(stub): allow for custom secure site stub file

jjpmann:
Fixes #941

PR #1238, opened 2022-04-13
use getStub to allow custom stub files

jjpmann:
Simplifying PR https://github.com/laravel/valet/pull/941 with this new one.

This will allow all stub files to be pulled from local VALET_HOME/stubs folder in case users want to create custom changes


mattstauffer:
Thanks for keeping this PR going for the last two years! I really appreciate your persistence!

Issue #1241, opened 2022-04-16
Command valet phplcomposerlwhich-php not working anymore globally


JacobDelcroix:
### Discussed in https://github.com/laravel/valet/discussions/1240

<div type='discussions-op-text'>

<sup>Originally posted by **JacobDelcroix** April 16, 2022</sup>
Laravel Valet 3.1.3

Hello,

I think a new bug has been introduced recently (maybe 3.1.2) with these commands:
- valet php
- valet composer
- valet which-php

**valet php command log**
```
➜ ~ valet php -v
/Users/jacob/.composer/vendor/laravel/valet/valet: line 88: Deprecated:: command not found
```

**valet which-php command log**
```
➜ ~ valet which-php

Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /Users/jacob/.composer/vendor/laravel/valet/cli/Valet/PhpFpm.php on line 327

Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /Users/jacob/.composer/vendor/laravel/valet/cli/Valet/PhpFpm.php on line 327

Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /Users/jacob/.composer/vendor/laravel/valet/cli/Valet/PhpFpm.php on line 327
/opt/homebrew/bin/php
```

This bug appear when using these commands outside of a project (it should use default php version just like before).. all is working fine when you use these inside a project.
</div>

NasirNobin:
I can reproduce this. Most probably introduced in this [release](https://github.com/laravel/valet/releases/tag/v3.1.2). I'll try to debug and submit a quick fix soon.

PR #1242, opened 2022-04-16
Remove preg_replace deprecation warning

NasirNobin:
Resolves https://github.com/laravel/valet/issues/1241

`Site::getSites()` is now adding "PHP Version" output. Introduced in https://github.com/laravel/valet/pull/1229. On that process Site:getPhpVersion() is passing `null` to `PhpFpm::normalizePhpVersion()` that is breaking  commands like `valet which-php` This PR should fix that.


mattstauffer:
Thanks @NasirNobin!

PR #1244, opened 2022-04-18
Confirm tld command

driesvints:
This PR adds a confirmation to the tld command to warn about its usage and unexpected results. It also indicates that this isn't an officially supported feature and should be used with caution.


mattstauffer:
Thanks @driesvints!!

Issue #1233, opened 2022-04-06
Add test that cover `cli/valet.php`


mattstauffer:
Currently all of our tests cover the PHP classes that do the core work of Valet, but not the actual entry-point `cli/valet.php`. It'd be great to have some tests covering that.

PR #1256, opened 2022-05-24
Test Valet commands

driesvints:
This is a first attempt at https://github.com/laravel/valet/issues/1233. I've split up the app init for Valet into a separate `cli/app.php` file which can be re-used in the test suite. This will run the actual commands using the `ApplicationTester` class by Symfony.

Unfortunately I couldn't get the first commands I'm trying to test to work and would appreciate any help with this.


mattstauffer:
@driesvints Just pushed up my branch and compared and I much prefer how you're working here, so as long as we're able to get yours working in CI I'd rather go with your solution. I'll try to make some time for it this week!

driesvints:
@mattstauffer nice. I've just merged master into my PR so it's up to date again.

tests/CliTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+use Symfony\Component\Console\Tester\ApplicationTester;
+
+/**
+ * @requires PHP >= 8.0
+ */
+class CliTest extends Yoast\PHPUnitPolyfills\TestCases\TestCase
+{
+    public function prepTestConfig()
```
mattstauffer, line 10: This can move into `setUp` if we want, just wanted to separate out this functionality while I was proving it out.

mattstauffer:
@driesvints OK, I got it fixed so it's working in CI. Basically, it changes `VALET_HOME_PATH` to be a local folder, preps that folder, and then that triggers the commands being registered in `app.php` (they weren't before because `VALET_HOME_PATH` didn't exist).

tests/CliTest.php
driesvints: Yeah I think moving this into setUp is the right call.

driesvints:
Nice work!

cli/app.php
```diff
@@ -0,0 +1,731 @@
...
+    /**
+     * Get all the current secured sites.
+     */
+    $app->command('secured', function () {
+        $sites = collect(Site::secured())->map(function ($url) {
+            return ['Site' => $url];
+        });
+
+        table(['Site'], $sites->all());
+    });
```
EvanGoss, line 233:
```suggestion
    })->descriptions('Display all of the currently secured sites');
```

mattstauffer:
OK, I think this is working.

The main output of ALL of this code and ALL of these commits are: 

Before this, we could only test our individual classes, but not the commands themselves.

Now, we can actually test commands. 

For example:

https://github.com/laravel/valet/pull/1256/files#diff-38057577cc680f5c7b83f3688b196a18ed53d744d984789303e8503373acd629

mattstauffer:
@driesvints Please feel free to take a look before I merge. I kinda... took this one and ran with it 😆 

My only disappointment is the need to add the `writer()` line at the top of each command. Couldn't figure out how to get away without that, but I'm gonna see if I can let my brain rest on it for a day or two to see if I come up with anything.

That said, this is good to merge if you like it, Dries.

driesvints:
Looks good to me! Nice job @mattstauffer 👍

mattstauffer:
Thanks @driesvints!

PR #1257, opened 2022-05-26
Fix wrong host-header call

ThunderBirdsX3:
Fix wrong host-header call from `-host-header` to `--host-header`


ThunderBirdsX3:
And i update one more function that call `ngrok` in machine if it exists.

driesvints:
I've sent in a separate PR that reverts the arm binary and also incorporates this change. Thanks.

https://github.com/laravel/valet/pull/1258

PR #1267, opened 2022-06-22
Extending the WordPress driver to run sites in subdirectories

zetrider:
The WordPress site can be installed in a subdirectory of the root folder. The current driver change allows you to deploy sites in subdirectories.

### DIFF
--- a/cli/drivers/WordPressValetDriver.php
+++ b/cli/drivers/WordPressValetDriver.php
@@ -3,6 +3,60 @@
 class WordPressValetDriver extends BasicValetDriver
 {
     /**
+     * @var string real wordpress site path.
+     */
+    public $wpSitePath = '';
+
+    /**
+     * Determine is a WordPress site.
+     *
+     * @param  string  $path
+     * @return bool
+     */
+    public function isWordPress($path)
+    {
+        return file_exists($path . '/wp-config.php') || file_exists($path . '/wp-config-sample.php');
+    }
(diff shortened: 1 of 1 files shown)


mattstauffer:
I'm not a WordPress dev, so I've asked for help:

https://twitter.com/stauffermatt/status/1598343856126394368

rosswintle:
Hi folks, I'm just looking into helping Matt review and test this. Great to see the WordPress driver being improved.

Installing WordPress in a subdirectory can mean one of a few different things.

Could you please give an example of what this PR lets you do, specifically stating both the directory/path and the root URL of the site so that we can review and test what you are expecting to work?

Thanks

ihorvorotnov:
@rosswintle FYI running WordPress core from a subdir (installed as a package via Composer - e.g. Roots and similar setups) works just fine out of the box, using this for years.

seemly:
I don't know why, but this PR didn't work for me.

I did manage to find a repo that does work for subdirectory-based WordPress multisite functionality. Well, with the exception of the missing return types:

https://github.com/Objectivco/WordPressMultisiteSubdirectoryValetDriver/blob/master/WordPressMultisiteSubdirectoryValetDriver.php

mattstauffer:
As a non-WP dev, I can't act on this myself; I'm not seeing consensus above, or any continued conversation, so I'm going to close it.

PR #1284, opened 2022-09-30
Enable local network sharing

thinkverse:
This PR enables a rudimentary version of local network sharing with the fix mentioned in (https://github.com/laravel/valet/issues/440#issuecomment-579830736). This pull request was suggested in (https://github.com/laravel/docs/pull/8267) by @driesvints.

I've tested the fix locally and it does enable local network sharing, though it's rudimentary at best since assets don't load properly given the URL doesn't match.

For example, if I have `http://laravel.test` and access is via local network sharing `http://192.168.1.111/laravel.test`, the site does load. But the asset URLs will use `http://laravel.test` and thus not load. I'm not sure if that is due to the `@vite` directive or something else. The same thing happens with the `asset` helper, which might be because of the `APP_URL`.

I did manage to get the CSS, JavaScript, and HMR working with Vite by starting the Vite server using the local IP address passed in as the host.

```bash
npm run dev -- --host 192.168.1.111
```

Since Vite adds the IP address in the `script` and `link` files, the IP address used for the host must be an address that can be accessed by all devices over the network, The safest route is to use the same IP address of the device you're trying to expose.

Not sure how to go about fixing the issue of built assets not loading due to the URL not matching though. 🤷‍♂️

### DIFF
--- a/server.php
+++ b/server.php
@@ -104,6 +104,21 @@ $valetConfig = json_decode(
 );
 
 /**
+ * If the HTTP_HOST is an IP address, check the start of the REQUEST_URI for a
+ * valid hostname, extract and use it as the effective HTTP_HOST in place
+ * of the IP. It enables the use of Valet in a local network.
+ */
+if (preg_match('/^([0-9]+\.){3}[0-9]+$/', $_SERVER['HTTP_HOST'])) {
+    $uri = ltrim($_SERVER['REQUEST_URI'], '/');
+
+    if (preg_match('/^[-.0-9a-zA-Z]+\.'. $valetConfig['tld'] .'/', $uri, $matches)) {
+        $host = $matches[0];
+        $_SERVER['HTTP_HOST'] = $host;
+        $_SERVER['REQUEST_URI'] = str_replace($host, '', $uri);
+    }
+}
+
+/**
  * Parse the URI and site / host for the incoming request.
  */
 $uri = rawurldecode(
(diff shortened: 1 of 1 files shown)


thinkverse:
Also, could anyone confirm or deny if this breaks Laravel Valets' wildcard DNS support? I don't use `x|nip.io` services so sadly cannot check that myself.

mattstauffer:
I'm good with this modification, but I'll need someone who uses nip.io to test it out for me first.

Put out a request: https://twitter.com/stauffermatt/status/1598332500056621062

F0rsaken:
I tested it on my setup and it worked, but there is a problem with Drivers. `BasicValetDriver` has a line

```php
$_SERVER['SERVER_ADDR'] = '127.0.0.1';
```

that overrides server address, making a redirect to `http://site-name.test`. I don't know why there is a `127.0.0.1` override, but we can make fallback like this, where we can replace the ovveride condition with a proper one:

```php
$serverAddr = $_SERVER['SERVER_ADDR'];

// condition when to override the server address if necessary
if (!$serverAddr || $serverAddr == 'localhost') {
    $_SERVER['SERVER_ADDR'] = '127.0.0.1';
}
```

same goes for every driver that do this. I can make a PR, but I would need to know when to enable this override 😄

mattstauffer:
@F0rsaken Unfortunately I don't know yet whether your code there would or wouldn't be enough... I'm *guessing* it would be? I'll test it out and see if I can figure out.

mattstauffer:
OK. I made a modification similar to yours @F0rsaken, but I didn't worry about the `localhost` check.

In theory that means this PR should work. I'll test it out now or shortly.

mattstauffer:
Well.. this isn't perfect, but it at least is a start, with a bit of a potential fix, and it makes the docs at least a bit less inaccurate. lol. I just ran it and it worked, but with the same issues mentioned w/r/t the assets.

Thanks @thinkverse!

PR #1289, opened 2022-10-23
Fix bedrockvaletdriver

EHLOVader:
I'd experienced an issue with the bedrock valet driver for awhile specific to the custom wp-admin route provided by plugins, like the hidden backend feature in iThemes security. 

I finally dug into it and found that the driver is out of sync with the WordPress driver and had a few headers missing from the other drivers. So I tested updating it to match the WordPress driver and it fixed my problem. 

This uses the built in routing available instead of assuming that the folder has an index.php or is part of WordPress core. 

If accepted please kindly add [hacktoberfest-accepted] label too, thanks.


EHLOVader:
I'm not seeing the same output from the failed check. Why does it see spaces where there were none?

mattstauffer:
Thanks @EHLOVader!

PR #1294, opened 2022-11-01
Enabling securing custom domains, by not adding default TLD to $domain when explicitly provided

mgussekloo:
Update the domain() function to include the default TLD only if the $domain variable is not given explicitly. If the $domain variable is given, just return that instead of sticking the default TLD on top. This feature lets you secure custom domains, e.g. "valet secure www.mywebsite.com" (if you also add www.mywebsite.com to your hosts file and create a symlink or folder in the Valet site directory, of course).

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -1088,7 +1088,7 @@ class Site
         //     return $parked['site'];
         // }
 
-        return ($domain ?: $this->host(getcwd())).'.'.$this->config->read()['tld'];
+        return ($domain ?: $this->host(getcwd()).'.'.$this->config->read()['tld']);
     }
 
     /**


drbyte:
Unfortunately as currently proposed, this would be a big change in how (many) people are used to using `valet secure` ... because it would require *always* providing a TLD when calling it. 

ie: `valet secure foo` currently secures `foo.test`, but this PR would make it only secure `foo`, which wouldn't have the intended results.

driesvints:
Yeah this indeed seems like a breaking change to me. Think it’s best to leave things be.

mgussekloo:
I've made a new attempt, that checks for a dot in the domain and includes an update to SiteTest to make sure the common usecase stays supported. Would it be ok to make a new pull request?

PR #1297, opened 2022-11-02
When formatting provided site name, only remove .tld if it's at the end

### LINKED ISSUE #1295: test in site name fails
### Description:
In line with our regular site setup we have the same locally with valet:
portal.test-work.test
welcome.test-work.test
The moment it goes wrong is when I try to use the isolate option. Then I get the error that portal-work isn't found as site name. That is correct as it is indeed portal.test-work. The probleem seems to be that .test is removed not just at the end, but just as it is found in the site name.

### Steps To Reproduce:
1. add a link to your project with a specific site name in line with ours e.g. portal.test-work
2. try to isolate the site with any php version
3. you should get the error that the site name isn't found in the list of sites

### LINKED ISSUE #1298: test in site name fails with isolation
### Description:
In line with our regular site setup we have the same locally with valet:
portal.test-work.test
welcome.test-work.test
The moment it goes wrong is when I try to use the isolate option. Then I get the error that portal-work isn't found as site name. That is correct as it is indeed portal.test-work. The probleem seems to be that .test is removed not just at the end, but just as it is found in the site name.
searches on internet doesn't bring any solutions to the table

### Steps To Reproduce:
add a link to your project with a specific site name in line with ours e.g. portal.test-work
try to isolate the site with any php version
you should get the error that the site name isn't found in the list of sites

### Diagnosis

[valet-diagnose.txt](https://github.com/laravel/valet/files/9927983/valet-diagnose.txt)

ErikDohmen:
(copied from Issue #1298 for context):

**Description:**
TL;DR: If your site name contains your TLD (e.g. if your TLD is `.test` and your site name contains `.test`) it will be removed inappropriately.

In line with our regular site setup we have the same locally with valet:
- portal.test-work.test
- welcome.test-work.test

The moment it goes wrong is when I try to use the `isolate` option. Then I get an error: `portal-work isn't found as site name`. That is correct as it is indeed `portal.test-work`. The problem seems to be that `.test` is removed not just at the end, but just as it is found in the site name.

Searches on internet doesn't bring any solutions to the table.

**Steps To Reproduce:**
- add a link to your project with a specific site name in line with ours e.g. `portal.test-work`
- try to isolate the site with any php version
- you should get the error that the site name isn't found in the list of sites

**Solutions in this PR:**

- [x] Check if tld is at the end of the site name and, if so, remove it
- [x] Add corresponding test

Fixes #1295, Fixes #1298.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -208,7 +208,10 @@ class Site
             $directory = $this->host(getcwd());
         }
 
-        $directory = str_replace('.'.$tld, '', $directory); // Remove .tld from sitename if it was provided
+        $length = strlen('.'.$tld);
+        if (substr($directory, -$length) === '.'.$tld) {
+            $directory = substr($directory, 0, -$length); // Remove .tld from sitename if it was provided
+        }
 
         if (! $this->parked()->merge($this->links())->where('site', $directory)->count() > 0) {
             throw new DomainException("The [{$directory}] site could not be found in Valet's site list.");
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -208,9 +208,9 @@ class Site
             $directory = $this->host(getcwd());
         }
 
-        $length = strlen('.'.$tld);
-        if (substr($directory, -$length) === '.'.$tld) {
-            $directory = substr($directory, 0, -$length); // Remove .tld from sitename if it was provided
+        // Remove .tld from the end of sitename if it was provided
+        if (ends_with($directory, '.'.$tld) {
+            $directory = substr($directory, 0, -(strlen('.'.$tld)));
         }
 
         if (! $this->parked()->merge($this->links())->where('site', $directory)->count() > 0) {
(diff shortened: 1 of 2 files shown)


cli/Valet/Site.php
```diff
@@ -208,7 +208,10 @@ public function getSiteUrl($directory)
             $directory = $this->host(getcwd());
         }
 
-        $directory = str_replace('.'.$tld, '', $directory); // Remove .tld from sitename if it was provided
+        $length = strlen('.'.$tld);
+        if (substr($directory, -$length) === '.'.$tld) {
+            $directory = substr($directory, 0, -$length); // Remove .tld from sitename if it was provided
+        }
```
mattstauffer, line 214:
```suggestion
        // Remove .tld from the end of sitename if it was provided
        if (ends_with($directory, '.'.$tld) {
            $directory = substr($directory, 0, -(strlen('.'.$tld)));
        }
```

mattstauffer (changes requested):
Just one small refactor and then this is good to go!

ErikDohmen:
@mattstauffer, thanks for the suggestion. I added it and added one more missing ) in your code suggestion to make it work.

PR #1302, opened 2022-11-15
Secure custom domains

mgussekloo:
I have attempted a new pull request, one that includes a test and takes the feedback from [#1294](https://github.com/laravel/valet/pull/1294) into account. Apologies if this is simply an unwanted feature.

Update the domain() function to include the default TLD only if the $domain variable does not include a dot. This lets you secure custom domains, e.g. "valet secure [www.mywebsite.com](http://www.mywebsite.com/)". (You would also have to add [www.mywebsite.com](http://www.mywebsite.com/) to your hosts file and create a symlink or folder in the Valet site directory, of course.)

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -1078,12 +1078,11 @@ class Site
      *
      * @return string
      */
-    public function domain($domain)
+    public function domain($domain='')
     {
     	if (str_contains($domain, '.')) {
     		return $domain;
     	}
-
         return ($domain ?: $this->host(getcwd())) .'.'.$this->config->read()['tld'];
     }
 
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -1080,15 +1080,11 @@ class Site
      */
     public function domain($domain)
     {
-        // if ($this->parked()->pluck('site')->contains($domain)) {
-        //     return $domain;
-        // }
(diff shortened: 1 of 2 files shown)


mattstauffer:
I don't see any value in Valet securing domains that Valet doesn't serve. Valet is structured around serving everything under one TLD and I think we should keep the API constrained to that.

Thank you for the PR, though!

zkingdesign:
This would be really nice to have. @mgussekloo do you have a tutorial where you show how this can be used?

mgussekloo:
@zkingdesign Sure, it's nothing too special. 

- Make sure Valet site is reachable through the custom domain name (see [this tutorial from Stefan Bauer](https://stefanbauer.me/articles/how-to-use-custom-domains-with-different-tlds-on-laravel-valet))
- in the folder where i "park" my valet sites, i create a folder with the domain name e.g., the folder "www.myexample.com"
- add the domain to my /etc/hosts so www.myexample.com resolves to 127.0.0.1

Then, simply ask Valet to secure the custom domain, e.g. "valet secure www.myexample.com"
Make sure you modify Valet (see my pull request), else it will stubbornly secure "www.myexample.com.test" instead.

zkingdesign:
Thanks @mgussekloo for your help!

PR #1303, opened 2022-11-18
If a site is serving from public, check for individual PHP files in the public directory

silviokennecke:
Currently the public directory is not searched for the requested php file.

Example directory structure:
```
- my-project
|- public
 |- myscript.php
 |- index.php
```

A `GET http://my-project.test/myscript.php` would currently execute `my-project/public/index.php` because the `myscript.php` file is only searched at the `$sitePath`, which in this case is `my-project/`.

### DIFF
--- a/cli/drivers/BasicValetDriver.php
+++ b/cli/drivers/BasicValetDriver.php
@@ -68,6 +68,22 @@ class BasicValetDriver extends ValetDriver
             }
         }
 
+        $publicDynamicCandidates = [
+            $this->asActualFile($sitePath.'/public', $uri),
+            $this->asPhpIndexFileInDirectory($sitePath.'/public', $uri),
+            $this->asHtmlIndexFileInDirectory($sitePath.'/public', $uri),
+        ];
+
+        foreach ($publicDynamicCandidates as $candidate) {
+            if ($this->isActualFile($candidate)) {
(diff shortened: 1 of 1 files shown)


driesvints:
This doesn't seems correct to me. Surely your `$sitePath` should be `my-project/public`?

silviokennecke:
In my case, I parked the parent directory where all my projects are inside and the sitePath happens to be the absolute path to `my-project`

silviokennecke:
Another way to solve this could of course be, changing the `get_valet_site_path` function to ensure $sitePath is changed if a public directory is present. But that would require to also change all other drivers

mattstauffer:
Interesting. I'm leaning toward thinking that if we have `public/index.php`, *all* files should serve from there, which would mean your note about appending `public` to `$sitePath` would be our best option, even if it may be a breaking change.

Let me think about this and get back to you soon. Thanks for catching this!

silviokennecke:
Yes, in most cases, using only the index.php is totally fine, but in some cases this breaks stuff. In my case it was because a plugin added a php file in the public directory.

mattstauffer:
@silviokennecke Do you have the ability to test out the code I wrote in #1311 to make sure it solves your problem?

silviokennecke:
@mattstauffer, not until now. I'll try to test it this week

PR #1315, opened 2022-12-07
Add legacy drivers for the remainder of v3

### LINKED ISSUE #1314: Valet 3.2.0 breaks custom drivers that extend built-in drivers
In Valet 3.1.x, you could do this:

```php
class LocalValetDriver extends LaravelValetDriver {
```

or

```php
class LocalValetDriver extends WordPressValetDriver {
```

and this is what was recommended [in the documentation](https://laravel.com/docs/9.x/valet#local-drivers).

As of Valet 3.2.0 (and 3.2.1 does not fix this), this does not work.

> **Fatal error:** Uncaught Error: Class "WordPressValetDriver" not found in /Path/To/Site/LocalValetDriver.php

However, if you reference the full namespace `\Valet\Drivers\LaravelValetDriver` or `\Valet\Drivers\WordPressValetDriver`, it will work again.

I can update my custom drivers, but it really seems like this was an unintentional breaking change.

mattstauffer:
Closes #1314.


mattstauffer:
@markjaquith I just did a quick refactor; would you mind testing it out after the refactor? if it works for you after this, I'll merge it in. Thanks so much!

markjaquith:
> would you mind testing it out after the refactor?

Yep, at least with standard WordPress (didn't test Bedrock), 1684068 works for me.

mattstauffer:
Thanks again @markjaquith!

PR #1317, opened 2022-12-08
Fix legacy driver alias implementation

markjaquith:
- Loads legacy drivers when loading the server.
- Fixes WordPressValetDriver reference.


mattstauffer:
Thanks for catching both @markjaquith!

PR #1319, opened 2022-12-18
Extract Server class and refactor loading of drivers

mattstauffer:
This is a mega PR. Oops. :)

- Move server helpers to their own file, and then move them again into a Server class
- Add tons of type hints and return type hints
- Move all drivers but `Basic` and `Laravel` into a `Specific` directory so they can be loaded via glob
- Add a `beforeLoading` hook to simplify the `frontControllerPath` method of many drivers


driesvints:
@mattstauffer if we're going to add type hints might as well remove all docblock params and return types.

PR #1325, opened 2022-12-22
Add `set-ngrok-token` command

mattstauffer:
This allows users to set their ngrok auth token, now required for all usages of `valet share`.

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -75,7 +75,7 @@ class Ngrok
     /**
      * Set the Ngrok auth token.
      *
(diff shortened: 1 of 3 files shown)


mattstauffer:
@driesvints Is there anyone at Laravel who I could request to write documentation for a few features? Or is this something I should just take the time to document myself? Thanks!

driesvints:
@mattstauffer I can definitely check into that but currently I still have a few other things on my plate so not sure when I could get to it. If you could list the features with their PR's that you'd like to get documented feel free to send that my way.

PR #1328, opened 2022-12-26
Support isolated sites running PHP 7.4

mattstauffer:
Drop a few usages of PHP 8+ features so that users can isolate individual sites to PHP 7.4.

This still requires running PHP 8+ at the system level, but at least allows users to isolate certain sites to PHP 7.4.


mattstauffer:
I just ran some tests and this seems to work 👌

PR #1329, opened 2022-12-27
Add status command

mattstauffer:
Add a `status` command that checks to make sure Valet and its services are installed, running, and configured correctly.

### DIFF
--- a/cli/Valet/Configuration.php
+++ b/cli/Valet/Configuration.php
@@ -211,7 +211,7 @@ class Configuration
      */
     public function read(): array
     {
-        return json_decode($this->files->get($this->path()), true);
+        return json_decode($this->files->get($this->path()), true, 512, JSON_THROW_ON_ERROR);
     }
 
     /**
--- a/cli/Valet/Status.php
+++ b/cli/Valet/Status.php
@@ -5,7 +5,7 @@ namespace Valet;
 class Status
 {
     /**
-     * Create a new Status isntance.
+     * Create a new Status instance.
      *
      * @param  Filesystem  $files
      * @param  Brew  $brew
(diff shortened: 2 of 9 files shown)


mattstauffer:
@nicoverbruggen I haven't checked in with you lately.. would this status tool help you at all? I don't know if you're actively checking the status of the Brew services or anything else already?

nicoverbruggen:
@mattstauffer I actively check the Homebrew services, yes. 

The annoying part is that Valet's services run as admin, and some users will incorrectly run their services using their own user account (especially if they themselves installed PHP using `brew install php`): in that case, services might not be running correctly. 

So I tend to run both of these:

```
sudo brew services info --all --json
brew services info --all --json
```

and depending on what info I need I'll pick out the information from here.

This can also be used to find out if a user has a PHP service running incorrectly. (i.e. phpmon will shut down all PHP services running as the user account (as output of `brew services info --all --json`) and start the correct ones as root by running `sudo brew services start php@x.x`).

The services manager in PHP Monitor can be found [here](https://github.com/nicoverbruggen/phpmon/blob/e29ca9e1ad17b6bba9ef693df5f12ef38fc9f749/phpmon/Domain/App/ServicesManager.swift#L19).

mattstauffer:
@nicoverbruggen Thanks! Would this native Valet status command have any value to you, or are you happy with what you're currently doing?

nicoverbruggen:
@mattstauffer I'll stick with what I'm doing for now, yes!

mattstauffer:
Nico, the idea to check both the root and the user versions of the services was exactly what I needed to make this command more correct. Thank you!!

PR #1337, opened 2023-01-10
Add which-pecl and pecl commands

N-Silbernagel:
This PR adds a "valet pecl" command which behaves in line with the existing "valet php" and "valet composer" commands, two very helpful commands in day-to-day development.

One thing I alway need to do when setting a new php project with a new php version is to install extensions like xdebug. What I do right now is to get the pecl binary via "valet which-php", replacing php with pecl and using that path to install the extensions I need. 
Even tho my dicussion about it (https://github.com/laravel/valet/discussions/1331) did not go so well, I still went and implemented it, because this command would help me a lot in my process and I think other people would benefit from it as well.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -309,35 +309,18 @@ class Brew
      */
     public function getPhpExecutablePath($phpVersion = null)
     {
-        if (! $phpVersion) {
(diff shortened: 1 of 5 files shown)


N-Silbernagel:
closing because wrong git email

PR #1338, opened 2023-01-10
Add which-pecl and pecl commands

N-Silbernagel:
This PR adds a "valet pecl" command which behaves in line with the existing "valet php" and "valet composer" commands, two very helpful commands in day-to-day development.

One thing I alway need to do when setting a new php project with a new php version is to install extensions like xdebug. What I do right now is to get the pecl binary via "valet which-php", replacing php with pecl and using that path to install the extensions I need.
Even tho my dicussion about it (https://github.com/laravel/valet/discussions/1331) did not go so well, I still went and implemented it, because this command would help me a lot in my process and I think other people would benefit from it as well.

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -309,35 +309,18 @@ class Brew
      */
     public function getPhpExecutablePath($phpVersion = null)
     {
-        if (! $phpVersion) {
-            return BREW_PREFIX.'/bin/php';
-        }
+        return $this->findInBin($phpVersion, 'php');
+    }
 
-        $phpVersion = PhpFpm::normalizePhpVersion($phpVersion);
-
-        // Check the default `/opt/homebrew/opt/php@8.1/bin/php` location first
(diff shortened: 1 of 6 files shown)


mattstauffer:
Thanks for this PR! Unfortunately I think it's not a commonly-used or -needed tool, so I'd rather not add it to our maintenance burden.

I do want to note I think this is an elegant solution, and I wish I could merge it just for that reason!

N-Silbernagel:
I see. Thank you for the kind review

theutz:
Thanks for this PR. Just in case anyone else is trying to make this work, I'll leave some extra detail about a workaround.

You can programmatically determine the location of the proper pecl executable and store it as a variable, like so:

```bash
pecl="$(valet which-php | sed 's/php$/pecl')"
```

After you've done this, you can use `$pecl` instead of `pecl` when running the script. For instance:

```bash
$pecl install redis
```

PR #1341, opened 2023-01-13
Laravel v10 Support

driesvints:
I don't think we're actually testing Laravel v10 in here though.


mattstauffer:
@driesvints Yah, we're not running tests... we're also not using anything particularly clever with the container so I assume it'll be fine :)

mattstauffer (approved):
Looks good to me.

PR #1343, opened 2023-01-13
Fixes permission denied check when result starts with 'Permission denied'

matthewjohns0n:
When creating a certificate, if the result of the command starts with "Permission denied" it doesn't know that the permission was denied, so it doesn't try to run as root.

The current check will only work if the result contains "Permission denied" but will fail if the result _starts with_ "Permission denied" since that will return `0`.

I also wasn't sure which branch to PR this into, but it looks like `master` is typically where things get PR'd into. Let me know if it needs to be into a different branch and I can change it.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -620,7 +620,7 @@ class Site
         ));
 
         // If cert could not be created using runAsUser(), use run().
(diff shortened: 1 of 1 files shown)


mattstauffer:
Great, thanks! and yes, `master` was correct.

PR #1347, opened 2023-01-19
Add support for `.valetrc`

mattstauffer:


mattstauffer:
Checks failing because of some wonky issues with GH actions, but the tests seem to pass so I'm not bothered at the moment.

driesvints:
Hey @mattstauffer. It might be best to retrigger the builds before merging otherwise you risk merging in failing builds meaning all PRs after it will suffer failing builds as well. If a build is failing then there's a real issue that needs to be solved (maybe not related to Valet but dependencies/tests or otherwise).

PR #1350, opened 2023-01-25
Add artisan command

moisish:
This PR adds a "valet artisan" command which behaves in line with the existing "valet php" and "valet composer" commands. The "valet artisan" command is basically a shortcut for "valet php artisan"

### DIFF
--- a/cli/app.php
+++ b/cli/app.php
@@ -645,6 +645,15 @@ You might also want to investigate your global Composer configs. Helpful command
     ]);
 
     /**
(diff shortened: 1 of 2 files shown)


mattstauffer:
Thanks for the PR @moisish! I'm trying to avoid any new unnecessary expansions of the API, as I've had to make a lot of changes already in order to support isolation and Expose and some other new features. So while this is convenient, I'm going to not merge it. Thank you for your work on it though!

PR #1360, opened 2023-02-07
Add php isolation from link command

joelbutcher:
This PR  improves the `valet link` command by adding an `--isolate` option to allow users to link, secure and isolate a site in one command.

**Before**
```bash
valet link --secure && valet isolate
```

**After**
```bash
valet link --secure --isolate
```

Please note that the `--isolate` option requires a `.valetphprc` file in the projects root directory as I could not get `$this->runCommand`  to accept any additional arguments / options.

### DIFF
--- a/cli/app.php
+++ b/cli/app.php
@@ -175,7 +175,7 @@ if (is_dir(VALET_HOME_PATH)) {
     /**
      * Register a symbolic link with Valet.
      */
-    $app->command('link [name] [--secure]', function (OutputInterface $output, $name, $secure) {
+    $app->command('link [name] [--secure] [--phpVersion=]', function (OutputInterface $output, $name, $secure, $phpVersion = null) {
         $linkPath = Site::link(getcwd(), $name = $name ?: basename(getcwd()));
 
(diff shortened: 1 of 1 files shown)


cli/app.php
```diff
@@ -175,15 +175,26 @@ function (ConsoleCommandEvent $event) {
...
      */
-    $app->command('link [name] [--secure]', function (OutputInterface $output, $name, $secure) {
+    $app->command('link [name] [--secure] [--isolate]', function (OutputInterface $output, $name, $secure, $isolate) {
         $linkPath = Site::link(getcwd(), $name = $name ?: basename(getcwd()));
 
         info('A ['.$name.'] symbolic link has been created in ['.$linkPath.'].');
 
         if ($secure) {
-            $this->runCommand('secure '.$name);
+            $this->runCommand('secure');
```
mattstauffer, line 184: What's the benefit of dropping `$name` here?

joelbutcher: It's not passed to the underlying command in any way

mattstauffer:
Thanks @joelbutcher!

PR #1366, opened 2023-02-13
Use Pint for Code Styling

driesvints:
This PR sets up Pint to lint code and fail when code isn't styled according to Laravel's coding standards. It will only auto-commit when it's run on master.

This PR also includes the first run and fixes by Pint.


mattstauffer:
Thanks @driesvints!!

PR #1370, opened 2023-02-17
Update php and composer commands to allow passing in specific site

### LINKED ISSUE #1367: `valet php` can't be defined to run on a site linked to a name different than its directory
### Description:
If I want to run a php command such as: `valet php -S 0.0.0.0:3030 -t ./public` it will use the default installed /usr/bin/php rather than the isolated php version if my site is isolated.

It would be preferred if I could pick my php version or name the site when running php commands like this, for example:

`valet php --site=mycustomsite -S 0.0.0.0:3030 -t ./public` OR `valet php --version=php@7.4 -S 0.0.0.0:3030 -t ./public`

### Steps To Reproduce:
In a folder OUTSIDE of your normally parked folder, or on a subdomain, isolate a site with a custom name and then run `valet php -v`, you will see that it does not use the isolated php version. 

### Diagnosis

This is not related to an internal issue that would be solvable using `diagnose`

mattstauffer:
Closes #1367.

E.g., to run this command:

```bash
valet php -v
```

When you've linked a site using a name other than the directory name; for example:

```bash
cd /tmp
mkdir my-site
cd my-site
valet link different-name-than-folder
valet isolate different-name-than-folder php@8.0
```

This PR allows you to run this:

```bash
valet php --site=different-name-than-folder -v
```

### DIFF
--- a/cli/app.php
+++ b/cli/app.php
@@ -641,19 +641,21 @@ You might also want to investigate your global Composer configs. Helpful command
     /**
      * Proxy commands through to an isolated site's version of PHP.
      */
-    $app->command('php [command]', function (OutputInterface $output, $command) {
+    $app->command('php [--site=] [command]', function (OutputInterface $output, $command) {
         warning('It looks like you are running `cli/valet.php` directly; please use the `valet` script in the project root instead.');
     })->descriptions("Proxy PHP commands with isolated site's PHP executable", [
         'command' => "Command to run with isolated site's PHP executable",
+        '--site' => 'Specify the site to use to get the PHP version (e.g. if the site isn\'t linked as its directory name)',
     ]);
 
     /**
      * Proxy commands through to an isolated site's version of Composer.
      */
-    $app->command('composer [command]', function (OutputInterface $output, $command) {
+    $app->command('composer [--site=] [command]', function (OutputInterface $output, $command) {
         warning('It looks like you are running `cli/valet.php` directly; please use the `valet` script in the project root instead.');
     })->descriptions("Proxy Composer commands with isolated site's PHP executable", [
         'command' => "Composer command to run with isolated site's PHP executable",
+        '--site' => 'Specify the site to use to get the PHP version (e.g. if the site isn\'t linked as its directory name)',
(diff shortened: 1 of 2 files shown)


cli/app.php
```diff
@@ -641,19 +641,21 @@ function (ConsoleCommandEvent $event) {
     /**
      * Proxy commands through to an isolated site's version of PHP.
      */
-    $app->command('php [command]', function (OutputInterface $output, $command) {
+    $app->command('php [--site=] [command]', function (OutputInterface $output, $command) {
```
driesvints, line 644:
```suggestion
    $app->command('php [command] [--site=]', function (OutputInterface $output, $command) {
```

cli/app.php
```diff
@@ -641,19 +641,21 @@ function (ConsoleCommandEvent $event) {
...
     })->descriptions("Proxy PHP commands with isolated site's PHP executable", [
         'command' => "Command to run with isolated site's PHP executable",
+        '--site' => 'Specify the site to use to get the PHP version (e.g. if the site isn\'t linked as its directory name)',
     ]);
 
     /**
      * Proxy commands through to an isolated site's version of Composer.
      */
-    $app->command('composer [command]', function (OutputInterface $output, $command) {
+    $app->command('composer [--site=] [command]', function (OutputInterface $output, $command) {
```
driesvints, line 654:
```suggestion
    $app->command('composer [command] [--site=]', function (OutputInterface $output, $command) {
```

driesvints (changes requested):

cli/app.php
driesvints: ah okay. Yeah that's fine. Bit odd though : P

mattstauffer: Yah.. I'm still wondering whether I should just do the work to make it possible to pass the `--site=` anywhere. It's totally doable, it's just.. gonna be really ugly in Bash.

scrummitch:
Working locally for me, test output here:

https://github.com/laravel/valet/issues/1367#issuecomment-1435479974

PR #1382, opened 2023-03-17
`brew tap` no longer necessary with Homebrew 4

vinkla:
Tapping `homebrew/core` is no longer necessary with [Homebrew 4](https://brew.sh/2023/02/16/homebrew-4.0.0/). 

> - Using JSON files downloaded from [formulae.brew.sh](https://formulae.brew.sh/) for package installation rather than local `homebrew/core` and `homebrew/cask` taps.
> - Unless you are developing formulae or casks, you can `brew untap homebrew/core` and `brew untap homebrew/cask` to save some space.

If a user has upgraded to Homebrew 4, we could skip tapping the repository. This will save about 600MB.

```
Untapping homebrew/core...
Untapped 3 commands and 6580 formulae (6,934 files, 609.7MB).
```

There are some things to take into consideration: 

1. Either Laravel Valet enforce Homebrew 4 (which shouldn't be a problem if Valet runs `brew update`)
2. or check which version of Homebrew is installed.

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -8,7 +8,6 @@ use Illuminate\Support\Collection;
 class PhpFpm
 {
     public $taps = [
-        'homebrew/homebrew-core',
         'shivammathur/php',
     ];
 


mattstauffer:
I'm OK with code that skips `core` if someone's on Homebrew 4, but that's all I feel comfortable with unless we have reason to think 80-90% of people are on Homebrew 4

vinkla:
If I’m not wrong it will automatically tap core if it needs to. This has to be tested.

drbyte:
Hmmm... it's telling me it can't untap once installed. 
(I haven't had a chance to test it on a newer Mac w/o brew already installed.)

`brew --version`
> Homebrew 4.0.26-29-g9a482de
> Homebrew/homebrew-core (git revision fa810a09209; last commit 2023-07-01)
> Homebrew/homebrew-cask (git revision 225dfff9ac; last commit 2023-07-01)

`brew info`
> 193 kegs, 114,946 files, 4.1GB

`brew tap-info`
> 7 taps, 0 private, 6747 formulae, 5 commands, 16,743 files, 900.3MB

`brew untap homebrew/core`
> Error: Untapping homebrew/core is not allowed

`brew untap homebrew/cask`
> Error: Refusing to untap homebrew/cask because it contains the following installed formulae or casks:
> ...

PR #1409, opened 2023-05-02
Add driver for Radicle by roots.io

theutz:
The good folks at [roots](https://roots.io) released a new opinionated boilerplate called for WordPress that brings a lot of Laravel-esque goodness to WP development.

However, the current `BedrockValetDriver` is not compatible with it, and so Valet doesn't work out-of-the-box.

This PR seeks to fix that by adding a new `RadicleValetDriver` that should do the trick.

(Re-opened from #1408)


theutz:
Closing in favor of better efforts underway within the Roots community.

PR #1426, opened 2023-07-04
Supported more HTTP requests

Nationalcat:
The previous configuration could only handle 5 requests simultaneously, causing it to freeze if it was a single-page application or if the page required numerous AJAX calls. With the current configuration, it can now handle up to 50 HTTP requests simultaneously.


drbyte:
I'm inclined to suggest this is something for you to configure on your own if needed.

Forcing it on every Valet user will immediately demand a lot more RAM and CPU whenever Valet is running.

driesvints:
Thank you. Right now we have only one request for this which leads me to believe this isn't needed for most use cases. Feel free to configure this yourself on your local machine.

PR #1427, opened 2023-07-06
Update BedrockValetDriver for roots/radicle Support

leek:
[Radicle](https://roots.io/radicle/) is a new starter kit for WordPress created by the makers of Bedrock.
It uses `public/` as a document root instead of `web/`.

This update allows the BedrockValetDriver to support `web/` or `public/` for serving.


leek:
Just needed to upgrade Valet!

Issue #1386, opened 2023-03-20
Valet link with --isolate doesn't pass custom name


chris-ware:
### Description:

When running `valet link --isolate` with a custom name that doesn't match the directory name, such as `valet link foobar --isolate`, where the directory the site is in may be foobar.com, the isolate command doesn't seem to receive the custom name and ends up failing with the following validation error.

```
In Site.php line 182:
                                                                  
  The [foobar.com] site could not be found in Valet's site list.  
                                                                  

link [--secure] [--isolate] [--] [<name>]
```

### Steps To Reproduce:

1. Checkout a site
2. Add a .valetrc file with an isolatable php version
3. `valet link --isolate foobar`

### Diagnosis

<details>
<summary>sw_vers</summary>
<pre>ProductName:	macOS
ProductVersion:	12.5.1
BuildVersion:	21G83</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 4.0.0</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "loopback": "127.0.0.1",
    "paths": [
        "/Users/chrisware/.config/valet/Sites"
    ]
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "beyondcode/expose": "^2.2",
        "laravel/envoy": "^2.8",
        "laravel/forge-cli": "^1.3",
        "laravel/installer": "^4.2",
        "laravel/pint": "^1.1",
        "laravel/valet": "^4.0",
        "statamic/cli": "^2.4",
        "tightenco/takeout": "^2.1"
    },
    "config": {
        "allow-plugins": {
            "ergebnis/composer-normalize": true
        },
    }
}</pre>
</details>
<details>
<summary>composer global diagnose</summary>
<pre>Changed current directory to /Users/chrisware/.composer
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.32.1
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
[...]

PR #1428, opened 2023-07-16
Fix link command's `--isolate` argument with custom name

mcaskill:
Related to #1386.

Tested in laravel/valet v4.1.2 with [mnapoli/silly](https://github.com/mnapoli/silly) v1.8.2 and [symfony/console](https://github.com/symfony/console) v6.2.5.

~~🟡 Unfortunately, a bug introduced in mnapoli/silly v1.8.1 breaks all arguments in `runCommand()` method preventing this pull request from resolving the isolation issue. See mnapoli/silly#69 for details and mnapoli/silly#70 for pull request.~~

🟢 Bug resolved in [mnapoli/silly v1.8.3](https://github.com/mnapoli/silly/releases/tag/1.8.3).

Passing a custom hostname to the `link` command with the `--isolate` argument will result in either a mismatched site being isolated or the site not being found:

```sh
~/Projects/example.com/www/_codebase
$ valet link example --secure --isolate
A [example] symbolic link has been created in [~/.config/valet/Sites/example].
Restarting nginx...
The [example.test] site has been secured with a fresh TLS certificate.
Found '_codebase/.valetrc' or '_codebase/.valetphprc' specifying version: php@8.1

In Site.php line 182:

  The [_codebase] site could not be found in Valet's site list.


link [--secure] [--isolate] [--] [<name>]
```

The reason for this is because the `isolate` command defaults to using the current directory's _name_ which won't match the custom hostname.

This issue does not affect the `secure` command because it resolves the hostname from the current directory's _full path_.

This pull request passes the hostname (regardless if custom or the current directory's name) to both sub-commands to ensure consistency of operations.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "php": "^7.1|^8.0",
         "illuminate/collections": "^8.0|^9.0|^10.0",
         "illuminate/container": "~5.1|^6.0|^7.0|^8.0|^9.0|^10.0",
-        "mnapoli/silly": "^1.0",
+        "mnapoli/silly": "^1.5",
         "symfony/console": "^3.0|^4.0|^5.0|^6.0",
         "symfony/process": "^3.0|^4.0|^5.0|^6.0",
         "guzzlehttp/guzzle": "^6.0|^7.4",
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,9 @@
             "tests/Drivers/BaseDriverTestCase.php"
         ]
     },
+    "conflict": {
+        "mnapoli/silly": ">=1.8.1 <1.8.3"
+    },
     "require": {
         "php": "^7.1|^8.0",
         "illuminate/collections": "^8.0|^9.0|^10.0",
--- a/tests/CliTest.php
+++ b/tests/CliTest.php
@@ -259,17 +259,17 @@ class CliTest extends BaseApplicationTestCase
     {
         [$app, $tester] = $this->appAndTester();
 
-        $cwd  = getcwd();
+        $cwd = getcwd();
         $name = 'tighten';
         $host = $name.'.test';
 
         $customPhpVersion = '82';
-        $phpRcVersion     = '8.2';
-        $fullPhpVersion   = 'php@8.2';
+        $phpRcVersion = '8.2';
+        $fullPhpVersion = 'php@8.2';
 
-        $brewMock  = Mockery::mock(Brew::class);
+        $brewMock = Mockery::mock(Brew::class);
         $nginxMock = Mockery::mock(Nginx::class);
-        $siteMock  = Mockery::mock(RealSite::class);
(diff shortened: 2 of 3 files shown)


drbyte:
👍

drbyte:
👍  Looks good to me. Thanks for all your work on this!

composer.json
```diff
@@ -34,11 +34,14 @@
             "tests/Drivers/BaseDriverTestCase.php"
         ]
     },
+    "conflict": {
+        "mnapoli/silly": ">=1.8.1 <1.8.3"
+    },
```
driesvints, line 39: Do we really need the conflict here? Can't we bump the minimum version?

mcaskill: We could. Would `^1.8.3` be sufficient?

drbyte: Unfortunately `^1.7` would be the very minimum matching set of requirements for compatibility with Valet in its present state, without triggering errors. 
Actually, `^1.5` is safest, for compatibility with older PHP that may be present on a given Mac.

Once we completely drop any backwards compatibility pre-7.4, then we could bump to `1.8.3`.

mcaskill: Makes sense. So, we will keep the `^1.5` requirement that I had bumped. Do I remove the conflict entry?

drbyte: I suggest leaving it as-is, but it's not my call. Deferring to Dries ;)

mcaskill: The conflict entry is there to avoid accidentally installing 1.8.1 or 1.8.2—or not updating from those two versions—in order for this `--isolate` PR to work.

driesvints: The conflict is fine. No worries. Ideally I want to avoid those as much as possible but nbd.

tests/CliTest.php
```diff
@@ -253,6 +254,70 @@ public function test_link_command_with_secure_flag_secures()
...
+        $name = 'tighten';
+        $host = $name.'.test';
+
+        $customPhpVersion = '82';
+        $phpRcVersion     = '8.2';
+        $fullPhpVersion   = 'php@8.2';
+
+        $brewMock  = Mockery::mock(Brew::class);
+        $nginxMock = Mockery::mock(Nginx::class);
+        $siteMock  = Mockery::mock(RealSite::class);
```
driesvints, line 271:
```suggestion
        $cwd = getcwd();
        $name = 'tighten';
        $host = $name.'.test';

        $customPhpVersion = '82';
        $phpRcVersion = '8.2';
        $fullPhpVersion = 'php@8.2';

        $brewMock = Mockery::mock(Brew::class);
        $nginxMock = Mockery::mock(Nginx::class);
        $siteMock = Mockery::mock(RealSite::class);
```

mcaskill:
I've rebased the latest changes.

mattstauffer:
Thank you so much @mcaskill, and thanks as always @drbyte and @driesvints!

Issue #1395, opened 2023-03-27
Can't install Valet if volume name has an space character


tytoalba99:
### Description:

Can't install Valet if volume name has an space character

### Steps To Reproduce:

Create a user in a external volume and name it with a space between: Ex: "John SSD"

### Diagnosis

<details>
<summary>sw_vers</summary>
<pre>ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H15</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>/Volumes/John SSD/john/.composer/vendor/laravel/valet
/Volumes/John SSD/john/.composer/vendor/laravel/valet
Laravel Valet 4.0.0</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "loopback": "127.0.0.1",
    "paths": []
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "laravel/valet": "^4.0"
    }
}</pre>
</details>
<details>
<summary>composer global diagnose</summary>
<pre>Changed current directory to /Volumes/John SSD/john/.composer
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.24.3
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.5.5
PHP version: 8.2.4
PHP binary path: /usr/local/Cellar/php/8.2.4/bin/php
OpenSSL version: OpenSSL 1.1.1t  7 Feb 2023
cURL version: 8.0.1 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1t
zip: extension present, unzip present, 7-Zip not available</pre>
</details>
<details>
<summary>composer global outdated</summary>
<pre>Changed current directory to /Volumes/John SSD/john/.composer
[...]

PR #1431, opened 2023-07-20
Add quotes around $PHP

datashaman:
If the PHP executable path has a space in it, it breaks valet. For example, when using the new Laravel Herd which stores its config and binaries under `~/Library/Application Support/Herd/...`.

May help with https://github.com/laravel/valet/issues/1395


drbyte:
👍

PR #1434, opened 2023-07-26
Only delete DNSMasq resolver when Herd is not installed

mpociot:
This PR ensures that the DNSMasq resolver configuration file only gets deleted when Laravel Herd is not installed.
Right now this introduces a bug when a user:

1. Installs Herd
2. Uninstalls Valet

As Valet then deletes the DNSMasq resolver that Herd still uses.

### DIFF
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -43,8 +43,13 @@ class DnsMasq
         $this->brew->stopService('dnsmasq');
         $this->brew->uninstallFormula('dnsmasq');
         $this->cli->run('rm -rf '.BREW_PREFIX.'/etc/dnsmasq.d/dnsmasq-valet.conf');
-        $tld = $this->configuration->read()['tld'];
-        $this->files->unlink($this->resolverPath.'/'.$tld);
+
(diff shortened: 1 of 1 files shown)


drbyte:
Looks good to me. 👍

RobertBoes:
Shouldn't Herd just create the file if it doesn't exist and needs it?

cli/Valet/DnsMasq.php
```diff
@@ -43,8 +43,13 @@ public function uninstall(): void
         $this->brew->stopService('dnsmasq');
         $this->brew->uninstallFormula('dnsmasq');
         $this->cli->run('rm -rf '.BREW_PREFIX.'/etc/dnsmasq.d/dnsmasq-valet.conf');
-        $tld = $this->configuration->read()['tld'];
-        $this->files->unlink($this->resolverPath.'/'.$tld);
+
+        // As Laravel Herd uses the same DnsMasq resolver, we should only
+        // delete if, if Herd is not installed.
```
mateusjunges, line 48:
```suggestion
        // delete it if Herd is not installed.
```

PR #1435, opened 2023-08-11
WIP: Merge Herd fork back in

mattstauffer:

### DIFF
--- a/cli/Valet/Context/Context.php
+++ b/cli/Valet/Context/Context.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Valet\Context;
+
+abstract class Context
+{
+    public static function assign()
+    {
+        if (static::isHerd()) {
+            return new Herd();
+        }
+
+        return new Standalone();
+    }
+
+    public static function isHerd()
+    {
+        return true; // @todo implement
+    }
+
+    public static function isStandalone()
+    {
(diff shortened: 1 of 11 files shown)


mattstauffer:
Current version of this introduces two concepts: the `Context` class, and its two children, `Herd` and `Standalone`.

You can ask at any point for an instance of `Context`:

```php
$context = resolve(Context::class);

if ($context->isHerd()) {
    // Do something
}
```

Or you can ask for the *currently bound* context:

```php
echo sprintf(
    "You are currently using %s",
    resolve(Context::class)->name(), // If you're on Herd, returns an instance of `Herd`, so name is "Laravel Herd"
);
```

cli/Valet/Context/Context.php
```diff
@@ -0,0 +1,27 @@
...
+        if (static::isHerd()) {
+            return new Herd();
+        }
+
+        return new Standalone();
+    }
+
+    public static function isHerd()
+    {
+        return true; // @todo implement
```
adrum, line 18: This should work in the release version: `basename($_SERVER['SCRIPT_NAME']) == 'herd.phar'`

Not sure what the solution is for non-release version.

mattstauffer:
Closing due to a lack of interest. @mpociot let me know if this is ever something you want to pursue in the future.

PR #1437, opened 2023-08-25
Add support for proxying multiple domains at once

RobertBoes:
This PR adds support for proxying multiple domains at once with a comma separated list of domains, such as `valet proxy my-domain,sub.my-domain http://127.0.0.1:8000`

The change is backwards compatible, since I explode the URL by comma, if there's no comma present it works as before.

Reason for this change is working locally with Octane can be a bit frustrating. We have an app that uses multiple domains and we use Valet to proxy those domains to Octane. However, currently we'd have to run these commands:
```sh
valet proxy domain http://127.0.0.1:8000
valet proxy sub.domain http://127.0.0.1:8000
valet proxy api.domain http://127.0.0.1:8000
```
This not only takes a long time, Valet also restarts nginx after each command, prompting for the sudo password every time.
With this change it can be done in a single command, is much faster and only restarts nginx once;
```sh
valet proxy domain,sub.domain,api.domain http://127.0.0.1:8000
```

It also includes the `unproxy` command
```sh
valet unproxy domain,sub.domain,api.domain
```
And this will also work if you proxied `domain,api.domain` and then run `valet unproxy api.domain`, as it will only unproxy the `api.domain` and keep the `domain` proxy

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -768,31 +768,34 @@ class Site
         }
 
         $tld = $this->config->read()['tld'];
-        if (! ends_with($url, '.'.$tld)) {
-            $url .= '.'.$tld;
+
+        foreach (explode(',', $url) as $proxyUrl) {
+            if (! ends_with($proxyUrl, '.'.$tld)) {
+                $proxyUrl .= '.'.$tld;
+            }
+
+            $siteConf = $this->replaceOldLoopbackWithNew(
+                $this->files->getStub($secure ? 'secure.proxy.valet.conf' : 'proxy.valet.conf'),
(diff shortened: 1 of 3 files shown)


drbyte:
Nice! Looks good to me.

SanderMuller:
This PR would improve our QoL by a ton. 

Currently we have to run the valet proxy 5 times (once for each (sub)domain) when starting Octane and run valet unproxy 5 times again when stopping octane. Each command triggers a security prompt on Mac and takes some time to execute.

Reducing this to one proxy/unproxy call per time would be really neat!

mattstauffer:
Fantastic--what a great PR! Thank you!

PR #1438, opened 2023-08-27
Drop Mailhog and Redis from default logs list, since Valet doesn't install them

mattstauffer:
In the original PR (https://github.com/laravel/valet/pull/757), we included Mailhog and Redis, but it's always bothered me because they're not actually installed by default.

This PR removes them and only has default log items that are actually installed.


drbyte:
Given that anyone who wants to tail other logs can simply add those keys to their own valet config file, I have no objection to this. And running `valet log` provides instructions on how to do that too.

PR #1448, opened 2023-09-21
PHP 8.3

driesvints:


driesvints:
@drbyte @mattstauffer might need your help here a bit at fixing the build for PHP 8.3

drbyte:
@driesvints I think if you add `'php@8.3'` between `'php'` and `'php@8.2'` in the top of `Brew.php` it should be good.

After 8.3 is officially released we can then also update the LATEST_PHP_VERSION constant a few lines below, but I wouldn't do that right now.

PR #1451, opened 2023-10-06
handle deprecated http2_push_preload conf for nginx >= 1.25.1

OpJePl44tsm4n:
Updated the config so it will work for fresh installs.

removed http2 and http2_push_preload because they are deprecated in latest Nginx versions.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -631,8 +631,11 @@ class Site
     public function buildSecureNginxServer(string $url, string $siteConf = null): string
     {
         if ($siteConf === null) {
+            $nginxVersion = ltrim(exec('nginx -v'), 'nginx version: nginx/');
+            $configFile = version_compare($nginxVersion, '1.25.1', ">=") ? 'secure.valet.conf' : 'secure.valet-legacy.conf';
+
             $siteConf = $this->replaceOldLoopbackWithNew(
-                $this->files->getStub('secure.valet.conf'),
+                $this->files->getStub($configFile),
                 'VALET_LOOPBACK',
                 $this->valetLoopback()
             );
@@ -774,8 +777,11 @@ class Site
                 $proxyUrl .= '.'.$tld;
             }
 
+            $nginxVersion = ltrim(exec('nginx -v'), 'nginx version: nginx/');
+            $configFile = version_compare($nginxVersion, '1.25.1', ">=") ? 'secure.proxy.valet.conf' : 'secure.proxy.valet-legacy.conf';
+
             $siteConf = $this->replaceOldLoopbackWithNew(
-                $this->files->getStub($secure ? 'secure.proxy.valet.conf' : 'proxy.valet.conf'),
+                $this->files->getStub($secure ? $configFile : 'proxy.valet.conf'),
                 'VALET_LOOPBACK',
                 $this->valetLoopback()
             );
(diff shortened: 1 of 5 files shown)


drbyte:
I agree that this is the correct fix.

Unfortunately it only works if the Nginx version is 1.25.1 or greater, which is quite new. The changed syntax is not backward-compatible. (Ref: [nginx docs](https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2) )

> 2023-08-15 | nginx-1.25.2 mainline version has been released.
2023-06-13 | nginx-1.25.1 mainline version has been released.

If someone is using (or is forced to use) an older Nginx version, they will get the following anytime they run `valet secure` or re-secure their sites with `valet tld`, and Valet will refuse to work until the old syntax is restored to all the "secured" Sites configs.

```
  Nginx cannot start; please check your nginx.conf [1: nginx: [warn] protocol options redefined for 127.0.0.1:443 in ~/.config/valet/Nginx/foo.test:9
  nginx: [emerg] unknown directive "http2" in ~/.config/valet/Nginx/foo.test:15
  ].
```

@mattstauffer how do you feel about incorporating nginx-version-detection so we can handle version-specific stub file handling?

OpJePl44tsm4n:
@drbyte thank you for your response, I was thinking about that. Without a mayor version bump, probably a version check would be needed, since installing Valet requires the latest nginx version by default. 

I'll take a look and will update the PR.

drbyte:
At least the old syntax is not throwing Fatal errors and preventing Nginx from starting.
Unfortunately the new syntax does throw fatal errors on older Nginx versions. 😢

OpJePl44tsm4n:
@drbyte true but fresh installs will require lates version of nginx so now it's broken as is. 
option is to cap the nginx version in the install command and update this part. 

```
$this->brew->installOrFail('nginx', []);
```

my suggestion is to add a check in Site.php to check the nginx version and require 
secure.valet.outdated.config for the older versions. This would make it work for all versions. 

I'll make a fix so you can check it.

cli/Valet/Site.php
```diff
@@ -631,8 +631,11 @@ public function buildCertificateConf(string $path, string $url): void
     public function buildSecureNginxServer(string $url, string $siteConf = null): string
     {
         if ($siteConf === null) {
+            $nginxVersion = ltrim(exec('nginx -v'), 'nginx version: nginx/');
```
drbyte, line 634: Since `ltrim()` technically doesn't support passing a "phrase" (it treats it as a bunch of characters, not a phrase), perhaps simply `str_replace()` might be better?

drbyte:
@mattstauffer @driesvints I think this is ready for merge. 

Of course, feel free to test on your local machine before+after upgrading nginx via `brew`

cli/Valet/Site.php
drbyte: Looks good. Thx. 👍

drbyte: @OpJePl44tsm4n perhaps also rework the Title of this PR slightly to incorporate "http2" into it?
Maybe something like "handle deprecated http2_push_preload conf for nginx >= 1.25.1" ?

mattstauffer:
@drbyte thank you so much for reviews and back-and-forths! Again, apologies for my delay here. @OpJePl44tsm4n thank you for this PR! 

Testing now, and will merge if it doesn't bonk things for me.

PR #1460, opened 2023-12-05
Update "fetch-share-url" to pull correct URL for HTTPS tunnels, not just HTTP

rana01645:
Every time I'm using `valet fetch-share-url` it's giving this error

`There is no Ngrok tunnel established for ***********.test`

During deep diving, I found that my tunnel is using HTTPS protocol which is why it's unable to get the shared URL!

image

This pull request will solve this issue!

### DIFF
--- a/cli/Valet/Ngrok.php
+++ b/cli/Valet/Ngrok.php
@@ -51,15 +51,14 @@ class Ngrok
     }
 
     /**
-     * Find the HTTP tunnel URL from the list of tunnels.
+     * Find the HTTP/HTTPS tunnel URL from the list of tunnels.
      */
     public function findHttpTunnelUrl(array $tunnels, string $domain): ?string
     {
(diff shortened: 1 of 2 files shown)


mattstauffer:
@rana01645 This is very helpful!

Do you know *why* you're getting an HTTPS tunnel instead of HTTP, by chance?

mattstauffer:
To be honest, I don't know why we want the HTTP version locally, but I've run into this problem before too, so I'm going to merge it with a hope and a prayer that there's nothing *bad* about pulling in the HTTPS version when that's all we have.

rana01645:
@mattstauffer I also don't know why I'm getting only the HTTPS, we use `fetch-share-url` to automate the local dev setup using a bash command, and suddenly the automation fails due to this error!

It could be something related to ngrok maybe!

PR #1465, opened 2023-12-27
Allow renew() to be called without parameters

drbyte:
A reasonable default is set here, consistent with the default set in other places.

Fixes #1464
Updates #1461

@driesvints Can you review and consider releasing a patch, since this is a breaking bug that occurs on `valet install` (which people run not only to install Valet, but also after upgrading it)

PR #1466, opened 2023-12-28
certificate renew issue in `valet install`; no value for expireIn provided

rana01645:
`valet install` is throwing this error as no default value is provided for `Site::renew();`
image


drbyte:
Seems to be a duplicate of #1465

driesvints:
Thanks. We fixed this.

PR #1469, opened 2024-02-09
Allow specifying php version to restart

N-Silbernagel:
This PR adds the ability to specify aPHP version to restart. This reduces the time it takes to restart PHP on system with multiple PHP versions installed when change an ini value or troubleshooting.

Let me know if this is of interest to the project and if I should add similar functionality for starting and stopping php.


mattstauffer:
Thanks @N-Silbernagel!

PR #1471, opened 2024-02-19
Update KirbyValetDriver.php

Werbschaft:
This makes it possible to open PHP files again, see this discussion. The solution from @rasteiner works perfectly. You can also have a look @pedroborges ?

https://forum.getkirby.com/t/use-php-and-kirby-in-css-file/30724


mattstauffer:
Thanks @Werbschaft!

PR #1477, opened 2024-03-22
Add values() call to reset keys after modifying paths

adriaanzon:
This fixes an issue where running commands like `valet link foo` would cause the `paths` config array in ~/.config/valet/config.json to turn into an object with non-sequential keys. This in turn caused phpmon to fail with "Laravel Valet configuration file invalid or missing", when it tried to parse the config file.


mattstauffer:
@adriaanzon sorry for the delay, but can you give an example of what caused this to happen? I can't reproduce it.

mattstauffer:
Eh. Even if I can't repro it, it shouldn't have a negative effect because it's just enforcing that this does what we already expect it's going to do.

PR #1478, opened 2024-04-01
Ignore invalid paths

ahmedash95:
### Why?

While removing a directory that is parked in Valet. I encountered an error originating from Herd's Valet, though the issue lies with Valet itself. And from a user's perspective, I believe Valet should remain operational even if one of the paths is invalid

image

#### Suggestion
Introducing a command such as `valet check` would be beneficial. It could report on Valet's status, confirming if all paths are correct or if any path is invalid.

### DIFF
--- a/cli/Valet/Server.php
+++ b/cli/Valet/Server.php
@@ -160,6 +160,10 @@ class Server
         $domain = static::domainFromSiteName($siteName);
 
         foreach ($this->config['paths'] as $path) {
+            if (! is_dir($path)) {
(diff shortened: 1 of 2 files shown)


drbyte:
I agree with this code change.

For cleanup, perhaps the `valet paths` command could offer to cleanup any invalid paths when it displays the ones it has registered. I'd support a PR or discussion for implementing that.

driesvints:
Thanks!

PR #1486, opened 2024-05-14
Adding support for Nette Framework.

antonL95:
Adding support for Nette Framework default directory structure which expects that he main index.php file is in the www/ folder instead of public/ folder


mattstauffer:
@antonL95 Can you update this to have a narrower `serves` function that's less likely to trigger for other projects?

antonL95:
@mattstauffer I updated the `servers` function to be much narrower. Since Nette framework uses (by default) .neon files for configuration it should be a pretty good giveaway that the project uses Nette.

mattstauffer:
Fantastic, thanks @antonL95!

Issue #1487, opened 2024-06-20
NET::ERR_CERT_AUTHORITY_INVALID


rabol:
### Description

I have just installed the latest version of Valet

I then did a ```Valet install``` and now I get this error on all .test sites

NET::ERR_CERT_AUTHORITY_INVALID

### Steps To Reproduce

install v 4.7.0
do a valet install

It will then renew all certificates

### Diagnosis

<details>
<summary>sw_vers</summary>
<pre>ProductName:		macOS
ProductVersion:		14.5
BuildVersion:		23F79</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 4.7.0</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "loopback": "127.0.0.1",
    "paths": [
        "/Users/rabol/code/web"
    ],
    "share-tool": "expose"
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "laravel/valet": "^4.0",
        "laravel/installer": "^5.0" 
    }
}</pre>
</details>
<details>
<summary>composer global diagnose</summary>
<pre>Changed current directory to /Users/rabol/.composer
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.39.3
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.5.8 => 2.7.7)
Composer version: 2.5.8
PHP version: 8.3.8
PHP binary path: /opt/homebrew/Cellar/php/8.3.8/bin/php
OpenSSL version: OpenSSL 3.3.1 4 Jun 2024
cURL version: 8.8.0 libz 1.2.12 ssl (SecureTransport) OpenSSL/3.3.1
zip: extension present, unzip present, 7-Zip not available</pre>
</details>
[...]

PR #1488, opened 2024-06-21
Require Trusting CA when securing sites

adrum:
When trying to reproduce https://github.com/laravel/valet/issues/1487, I was able to get a Site in this state by following these steps:

1. Fresh install of brew and Valet on macOS 14.5 (via [VirtualBuddy](https://github.com/insidegui/VirtualBuddy))
2. Install valet v4.6.0
3. secure a site
4. Manually remove the CA from Keychain Access.app
5. Site should still be valid, since its own cert is trusted.
6. Upgrade Valet to v4.7.0
7. Run `valet install`
8. This calls renew, which also calls secure on every site. Since the CA is not trusted, the user will be prompted for their password to trust the CA in the System Keychain. Instead of approving, **Click "Cancel"**
9. This will yield the NET::ERR_CERT_AUTHORITY_INVALID error, since the old site cert was removed from the System Keychain.

This PR aims to prevent this state by doing the following:

1. Prevent users from proceeding with the secure command if they cancel the Trust CA operation.
2. Within the `secure` command, moves the `createCa` invocation above the `unsecure` command to prevent the old site's certificates from being removed if the Trust CA command is canceled. This prevents the site from being in an unpredictable state. If the command fails, it should leave the site alone. Otherwise, the site will be unsecured unexpectedly. 

cc @driesvints

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -476,7 +476,7 @@ class Site
         $phpVersion = $this->customPhpVersion($url);
 
         // Create the CA if it doesn't exist.
-        // If the user cancels the trust operation, the old certificate will be not removed.
+        // If the user cancels the trust operation, the old certificate will not be removed.
         $this->files->ensureDirExists($this->caPath(), user());
         $caExpireInDate = (new \DateTime())->diff(new \DateTime("+{$caExpireInYears} years"));
         $this->createCa($caExpireInDate->format('%a'));
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -475,17 +475,18 @@ class Site
         // Extract in order to later preserve custom PHP version config when securing
         $phpVersion = $this->customPhpVersion($url);
 
-        $this->unsecure($url);
-
+        // Create the CA if it doesn't exist.
+        // If the user cancels the trust operation, the old certificate will be not removed.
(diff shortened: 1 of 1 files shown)


drbyte:
Logic seems sound to me.

driesvints:
Thanks @adrum

mattstauffer:
@adrum This is a pretty edge-case-y thing. I'll likely merge it anyway because the code isn't that complex, but is this really something we expect users to have to deal with? What's a practical situation in which someone is manually deleting the CA, and then also choosing not to trust the CA?

rabol:
Just a small note:

when one use the 'trust' option of Valet, one is not asked to confirm the creation of certificates
```
trust              Add sudoers files for Brew and Valet to make Valet commands run without passwords
```

In my case I'm not asked for any password when securing a local site

adrum:
@mattstauffer I agree with you that this is an edge case. 

I recall when implementing https://github.com/laravel/valet/pull/1463 my CA was not trusted in my machine, so I'm guessing it was due to an older version of Valet when I first set up my machine that never trusted it? Or I manually removed it, but I cannot verify that. (Nor does it matter, aside from there might be others out there without a trusted CA).

At the very least, it will help ensure those without a trusted CA will at least experience an error with a description of how to recover if they skip the Keychain Access GUI prompt.

Issue #1496, opened 2024-09-24
Need to renew expired self-signed Laravel root certificate


wturrell:
### Description

TL;DR: my certificates for the individual *.test domains are fine, but suddenly I have a "Laravel Valet CA Self Signed CN" with an expiry of 1 July 2021, so no sites will load.

I've been reading https://github.com/laravel/valet/issues/1487 …

Opening Keychain Access confirms the certificate is expired, as does running:

```
security verify-cert -c ~/.config/valet/CA/LaravelValetCASelfSigned.pem
(Cert Verify Result: CSSMERR_TP_CERT_EXPIRED)
```

I have upgraded Valet from 4.7.1 to 4.8.0.

I followed this bit of the instructions:

```
sudo security delete-certificate -c "Laravel Valet CA Self Signed CN" /Library/Keychains/System.keychain -- This will remove the cert from your System Keychain.
sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/.config/valet/CA/LaravelValetCASelfSigned.pem -- This will add it back
```

All that does is (correctly) remove the old cert, but then it puts it back again. How can I actually renew it?

I tried `valet renew` (as well as `unsecure` and `secure`) but this didn't help.

Thanks!

### Steps To Reproduce

-

### Diagnosis

<details>
<summary>sw_vers</summary>
<pre>ProductName:		macOS
ProductVersion:		14.6.1
BuildVersion:		23G93</pre>
</details>
<details>
<summary>valet --version</summary>
<pre>Laravel Valet 4.8.0</pre>
</details>
<details>
<summary>cat ~/.config/valet/config.json</summary>
<pre>{
    "tld": "test",
    "paths": [
        "/Users/wt/.config/valet/Sites",
        "/Users/wt/Sites/mprss",
        "/Users/wt/Sites/credobase/htdocs",
        "/Users/wt/Sites/markpack"
    ],
    "loopback": "127.0.0.1"
}</pre>
</details>
<details>
<summary>cat ~/.composer/composer.json</summary>
<pre>{
    "require": {
        "laravel/installer": "^5.6",
        "laravel/valet": "^4.0",
        "drupal/coder": "^8.3",
        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
        "statamic/cli": "^2.6"
    },
    "require-dev": {
        "squizlabs/php_codesniffer": "^3.5"
    },
    "config": {
        "allow-plugins": {
[...]

PR #1498, opened 2024-11-15
Add the ability to renew the Certificate Authority certificate

adrum:
This PR adds the ability to renew the Certificate Authority certificate optionally when running the `valet renew --ca` command. Additionally, it adds the ability to view the CA certificate's expiration date through `valet secured --ca`.

This should resolve the issue highlighted in #1496. It may also resolve https://github.com/laravel/valet/issues/1487, but I don't think I was able to verify if the CA expired. 

My comment [here](https://github.com/laravel/valet/issues/1496#issuecomment-2477759775) summarizes the scenario pretty well how users may experience this issue, but I would wager to guess it should be an increasingly uncommon scenario, given new CAs should have 20 years of validity.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -437,9 +437,9 @@ class Site
     /**
      * Get all of the URLs with expiration dates that are currently secured.
      */
-    public function securedWithDates(): array
+    public function securedWithDates($ca = false): array
     {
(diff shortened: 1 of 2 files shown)


mattstauffer:
Thanks @adrum!

PR #1499, opened 2024-11-17
Add support for PHP 8.4

SupianIDz:
Include PHP 8.4 in the list of supported PHP versions in both Valet Brew and GitHub Actions test matrix. This ensures compatibility and testing for the latest PHP release.

---
This fixes the `Unable to determine linked PHP when parsing '8.4'` issue when using Valet using PHP version 8.4

Screenshot 2024-11-17 at 20 54 10

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: ['8.0', 8.1, 8.2, 8.3]
+        php: ['8.0', 8.1, 8.2, 8.3, 8.4]
 
     name: PHP ${{ matrix.php }}
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center"><img src="/art/logo.svg"></p>
 
 <p align="center">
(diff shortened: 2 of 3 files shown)


drbyte:
Agreed.

We can update the `LATEST_PHP_VERSION` constant below it after 8.4 is officially released.

/cc @mattstauffer

SupianIDz:
PHP 8.4 is out now, so when will this be merged?

cc @mattstauffer

drbyte:
NOTE: As a temporary workaround, you can edit your `~/.composer/vendor/laravel/valet/cli/Valet/Brew.php` and add PHP 8.4 to the top of the supported-versions array.

Later, after this is merged, you can remove the edit. Or when you run `composer update` it will ask you if you want to "stash" or "replace" your edits. Choose "Replace" (or whatever means to delete your temporary edits).

RobertBoes:
@mattstauffer Right, that makes sense, sorry I worded it a bit shitty 😅

Thanks for the clarification, I was under the impression it just used Valet (since that's present in the binary file of Herd) and Valet contains a bit of code to account for Herd.

mattstauffer:
@RobertBoes that's totally understandable! 

That was the original plan, but seems to not be the direction the Herd team chose to go in the end. So I closed my PR to make that possible and they are now entirely separate codebases.

PR #1510, opened 2024-12-18
fix(entry-script): use correct php binary for proxy commands

saibotk:
These are the only instances, where the valet cli script is executed with the default php binary instead of the evaluated binary.

We now use the `$PHP` executable to ensure that the valet cli is always called with the correct binary even when proxying composer / php.


mattstauffer:
Thank you!

PR #1513, opened 2025-02-28
Add PHPUnit 12 support, and upgrade phpunit-polyfills

drbyte:
Added `DataProvider` attribute, to support newer PHPUnit, in order to support latest `yoast/phpunit-polyfills`.

Closes #1512 which only did part of the update


mattstauffer:
Thanks @drbyte!

PR #1514, opened 2025-04-15
fix(php-fpm): correctly reconfigure all versions

saibotk:
This previously only reapplied the configuration for the linked version. Because it is recommended to run valet install again after a brew upgrade, it is necessary to update all utilized PHP versions.

On brew upgrades, some versions received the default `www.conf` again. This would leave the user with a broken setup until they either call `valet use php@X.X` for all the versions or isolate a site again with the version.

This commit fixes the behavior by rerunning the configuration step on all php versions. This is more consistent, because in the next line all utilized PHP versions are also restarted.

Seems like this also solves issues like #1410 or at least lets them only rerun a single `valet install`, which is something users would do in a regular troubleshooting session.

### DIFF
--- a/cli/Valet/PhpFpm.php
+++ b/cli/Valet/PhpFpm.php
@@ -27,8 +27,9 @@ class PhpFpm
 
         $this->files->ensureDirExists(VALET_HOME_PATH.'/Log', user());
 
-        $phpVersion = $this->brew->linkedPhp();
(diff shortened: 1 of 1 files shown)


drbyte:
Agreed. 👍 

/cc @mattstauffer

mattstauffer:
OOO great find--thanks so much!

PR #1516, opened 2025-05-02
Update CraftValetDriver for compatbility with craftcms/security-patches

adrum:
When applying https://github.com/craftcms/security-patches to a Craft project, the fix for [CVE-2024-56145](https://github.com/advisories/GHSA-2p6p-9rc9-62j9) breaks sites served by Valet (and Herd). 

This fix allows these sites to work when using Valet (and Herd).

### DIFF
--- a/cli/Valet/Drivers/Specific/CraftValetDriver.php
+++ b/cli/Valet/Drivers/Specific/CraftValetDriver.php
@@ -193,6 +193,10 @@ class CraftValetDriver extends ValetDriver
         $_SERVER['PHP_SELF'] = $scriptName;
         $_SERVER['DOCUMENT_ROOT'] = $sitePath.'/'.$frontControllerDirectory;
 
+        if (isset($_SERVER['argv'])) {
+            unset($_SERVER['argv']);
+        }
+
         return $indexPath;
     }
 }


mattstauffer:
@adrum Not sure if you have the details, but can you explain how `argv` being set breaks anything? I read through the CVE and can't easily follow why that would be the case. Thanks!

adrum:
@mattstauffer Of course! When utilizing the security patches plugin, they are just throwing a 400 anytime it's set: 

https://github.com/craftcms/security-patches/blob/main/src/Extension.php#L45

adrum:
Here's the proper fix they applied: https://github.com/craftcms/cms/compare/3.9.13...3.9.14

I think the https://github.com/craftcms/security-patches package is a quick solution that blocks the vulnerability in a "big hammer" way.

mattstauffer:
Ahh, that's so helpful. So if you have this security patches thing set, then it forces this error every time. And future versions don't even do anything with `argv` I assume, so we can just merge this as-is.

I'm going to merge, but could you do me a sanity check and grep through the codebase to make sure there aren't any other references to `argv` in the codebase that we may be breaking with this? Seems unlikely...

Thanks so much!

mattstauffer:
@adrum brilliant. Thank you!

PR #1521, opened 2025-06-18
Create php-memory-limits.ini only one time

mrodespin:
Change how the php-memory-limits.ini is managed because when changed between PHP versions, all the configuration added previously is overriden and it is a waste of time if you work with several PHP versions during the day


mattstauffer:
Seems reasonable to me! Failing tests are un-related, it's because of a change made in the latest PHPUnit.

PR #1525, opened 2025-07-04
Fix PHP PATH_INFO

francescom:
The PHP PATH_INFO is a well established although not often used mechanism, supported by most webserver, and even when it is not supported it is patchable so that it works.

You can add data after the base url and before the GET parameters. When you call a page like this

path/to/page.php/some/path/info?parameter=value

PHP will put the string 'some/path/info' into $_SERVER['PATH_INFO].

Without this patch Valet tries to find the file path/to/page.php/some/path/info on disk, without finding it and returning a 404.

The patch removes the PATH_INFO string from $_SERVER['REQUEST_URI'] at the very beginning of server.php so Valet can find the correct file, then recovers the original $_SERVER['REQUEST_URI'] just before the final require.

Although the PATH_INFO mechanism is not often known and used, it is sometimes useful and there is no reason not to follow PHP on this, by failing to include the correct file.

### DIFF
--- a/server.php
+++ b/server.php
@@ -1,5 +1,22 @@
 <?php
 
+
+// FIX PHP PATH_INFO
+
+$originalRequestUri = $_SERVER['REQUEST_URI'];
+$path = parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH);
+
+// Needs patch
+if (substr($path, -strlen($_SERVER['PATH_INFO'])) == $_SERVER['PATH_INFO']) {
+    // Cut away PHP PATH_INFO from REQUEST_URI to find the correct file later
+    $_SERVER['REQUEST_URI'] =
+        substr($path, 0, -strlen($_SERVER['PATH_INFO'])) // Remove PATH_INFO: page.php/remove/additional/path
+        . substr($_SERVER['REQUEST_URI'], strlen($path)); // Restore GET parameters '?param=what&ever=floats'
+}
+unset($path); // Don't leave global garbage around
+
+// END FIX PHP PATH_INFO
+
 require_once './cli/includes/require-drivers.php';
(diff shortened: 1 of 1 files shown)


mattstauffer:
Hey @francescom, as far as I can tell `PATH_INFO` is not usually set on default Nginx configurations; you have to do some work to manually set it in Nginx, and it's really only mainly used in Apache.

Valet uses Nginx, in part with the intention to mimic the most common deploy environments for Laravel, so I'm not _super_ interested in patching it to be a non-standard Nginx installation--I think that sort of thing makes it more likely people will get a mismatch between their local and their deploy setting.

Is there any reason this is not something you can fix in userland, by creating a custom driver?

Thanks!

francescom:
You can restore standard PHP support for PHP_INFO in Valet by either putting the attached LocalValetDriver.php class in the website root, or by putting PathInfoValetDriver.php in the Valet Drivers folder

[LocalValetDriver.php](https://github.com/user-attachments/files/26237679/LocalValetDriver.php)

[PathInfoValetDriver.php](https://github.com/user-attachments/files/26237680/PathInfoValetDriver.php)

_The correct folder is int the class comments:_

-    macOS (Herd):    ~/.config/herd/config/valet/Drivers/
-    macOS (Valet):   ~/.valet/Drivers/
-    Linux (Valet):   ~/.valet/Drivers/
-    Windows (Herd):  C:\Users\<user>\.config\herd\config\valet\Drivers\
-    WSL (Herd Win):  /mnt/c/Users/<user>/.config/herd/config/valet/Drivers/

If the Drivers/ folder doesn't exist, create it.

mattstauffer:
Closing this as the preferred option is for people who need it to create the custom driver. Thanks!

PR #1527, opened 2025-07-14
fix(php): ignore warnings in CLI parsing

saibotk:
When running PHP with extensions like imagick, sometimes the extension needs a rebuild after `brew upgrade`. Otherwise, it will throw a warning:
```
Warning: Version warning: Imagick was compiled against ImageMagick version 1809 but version 1810 is loaded. Imagick will run but may behave surprisingly in Unknown on line 0
```

This warning is also printed, within the shell executions in Valet. But we use the output of those PHP CLI invocations. With those warnings being printed out, the parsing on the output fails and Valet cannot be used until the pecl extension is rebuilt.

With this patch, all CLI PHP invocations use ` -d error_reporting=1`. This suppresses warnings, but will still report errors.

It makes valet more robust in case a system package update causes warnings.

### DIFF
--- a/find-usable-php.php
+++ b/find-usable-php.php
@@ -4,7 +4,7 @@ $minimumPhpVersion = '8.0';
 
 // First, check if the system's linked "php" is 8+; if so, return that. This
 // is the most likely, most ideal, and fastest possible case
-$linkedPhpVersion = shell_exec('php -r "echo phpversion();"');
+$linkedPhpVersion = shell_exec('php -d error_reporting=1 -r "echo phpversion();"');
 
 if (version_compare($linkedPhpVersion, $minimumPhpVersion) >= 0) {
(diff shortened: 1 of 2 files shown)


mattstauffer:
Looks good to me initially. @drbyte any concerns before I merge?

drbyte:
This is probably the best-available approach. We can't automate the re-install of all possible extensions without adding a lot of complexity due to unpredictability. And suppression of warnings (to prevent undesired excess verbage) but not errors (where we do want a failure to be reported) makes the best sense, particularly in the use-cases where this PR has introduced changes (where Valet is asking PHP for some information that's largely unaffected by and immaterial to any warnings PHP itself may be encountering).

So @mattstauffer  I'll give it a yes. 👍

PR #1531, opened 2025-09-15
fix(TLS): create OpenSSL strict-mode compliant certs

Tugzrida:
As mentioned in #1530, I recently ran into some issues with Valet certs not passing OpenSSL's X509 strict mode, which was enabled by default in Python 3.13. This made it difficult for python apps to communicate with php apps run under Valet during development.

This PR adds the necessary X509 extensions to both the CA and per-site leaf certificates to satisfy strict mode.

Note that the reference to the extension config file has been removed from the CSR creation command as the AKID extension can't be added without the CA certificate present (which it isn't when running this command). This has no effect on the final certificate as the CSR extensions were being overwritten when the certificate was signed anyway.

I have tested the migration pathway of this change under VirtualBuddy, and securing sites after this change continues to work regardless of whether the CA certificate was generated before or after this change.

The only thing to note is if the CA cert was generated prior to this change (and therefore doesn't have a SKID extension), generated leaf certificates will have a blank AKID, which shouldn't impact their validity to non-strict mode clients, however they will not pass strict mode validation until the CA is deleted and re-created with the new extensions.

If desired, logic could be added to detect the presence of the SKID extension in the CA certificate and forcibly re-create it if not found, however I didn't think this was wise as a default as it would disrupt the trust of other Valet site certificates until they too were re-created, which would likely be unexpected behaviour.

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -570,7 +570,7 @@ class Site
         ));
 
         $this->cli->runAsUser(sprintf(
-            'openssl req -new -newkey rsa:2048 -days %s -nodes -x509 -subj "/C=/ST=/O=%s/localityName=/commonName=%s/organizationalUnitName=Developers/emailAddress=%s/" -keyout "%s" -out "%s"',
+            'openssl req -new -newkey rsa:2048 -days %s -nodes -x509 -subj "/C=/ST=/O=%s/localityName=/commonName=%s/organizationalUnitName=Developers/emailAddress=%s/" -keyout "%s" -out "%s" -addext "basicConstraints=critical,CA:TRUE" -addext "keyUsage=critical,digitalSignature,keyCertSign" -addext "subjectKeyIdentifier=hash"',
             $caExpireInDays, $oName, $cName, 'rootcertificate@laravel.valet', $caKeyPath, $caPemPath
         ));
         $this->trustCa($caPemPath);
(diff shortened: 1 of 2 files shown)


cli/Valet/Site.php
```diff
@@ -649,7 +649,7 @@ public function createPrivateKey(string $keyPath): void
     public function createSigningRequest(string $url, string $keyPath, string $csrPath, string $confPath): void
     {
         $this->cli->runAsUser(sprintf(
-            'openssl req -new -key "%s" -out "%s" -subj "/C=/ST=/O=/localityName=/commonName=%s/organizationalUnitName=/emailAddress=%s%s/" -config "%s"',
+            'openssl req -new -key "%s" -out "%s" -subj "/C=/ST=/O=/localityName=/commonName=%s/organizationalUnitName=/emailAddress=%s%s/"',
             $keyPath, $csrPath, $url, $url, '@laravel.valet', $confPath
```
drbyte, line 653: Should `, $confPath` be removed as well, since the previous line removed the `%s` which referenced it?

cli/Valet/Site.php
```diff
@@ -570,7 +570,7 @@ public function createCa(int $caExpireInDays): void
         ));
 
         $this->cli->runAsUser(sprintf(
-            'openssl req -new -newkey rsa:2048 -days %s -nodes -x509 -subj "/C=/ST=/O=%s/localityName=/commonName=%s/organizationalUnitName=Developers/emailAddress=%s/" -keyout "%s" -out "%s"',
+            'openssl req -new -newkey rsa:2048 -days %s -nodes -x509 -subj "/C=/ST=/O=%s/localityName=/commonName=%s/organizationalUnitName=Developers/emailAddress=%s/" -keyout "%s" -out "%s" -addext "basicConstraints=critical,CA:TRUE" -addext "keyUsage=critical,digitalSignature,keyCertSign" -addext "subjectKeyIdentifier=hash"',
```
drbyte, line 573: I'm not an expert in this subject matter, but while looking at this change I notice here you added `basicConstraints=critical,CA:TRUE`, but in the `openssl.conf` file it's `basicConstraints = critical,CA:FALSE`. The TRUE vs FALSE stood out. 
Was that intentional? Is it moot? Just curious. And commenting "just in case" it matters.

cli/Valet/Site.php
Tugzrida: Thanks for this catch, fixed

drbyte:
Looks good to me 👍

mattstauffer:
Thanks @Tugzrida and @drbyte!

PR #1536, opened 2025-10-28
Create seperate dnsmasq configs for every site or proxy

nforst:
I have added that a separate dnsmasq config is created for each site or proxy. The exact explanation is in this feature request

https://github.com/laravel/valet/discussions/1534

### DIFF
--- a/tests/CliTest.php
+++ b/tests/CliTest.php
@@ -228,7 +228,7 @@ class CliTest extends BaseApplicationTestCase
         [$app, $tester] = $this->appAndTester();
 
         $site = Mockery::mock(RealSite::class);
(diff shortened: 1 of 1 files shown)


drbyte:
You are correct that you can add multiple `.conf` files into the `dnsmasq.d` directory for dnsmasq to pick up and let you customize.

The problem I see with this PR in its current state (unless I'm misreading the code) is that by making it per-host-only you've removed the generic `*.test` (or whatever configured tld) resolver that allows you to use `http://foo.test` by simply creating a `foo` directory in your `Sites` directory (all the other parked paths you've added in the valet config file).

PR #1554, opened 2026-06-18
Added support to determine php versions using composer.json

indykoning:
This PR adds support to fall back to the php version defined in the composer.json
Due to using composer/semver all combination of php requirements are possible (though the match is not exact as we only care about the x.x version, no patch versions)

Requirements like `^8.3`, `8.3`, `<8.5`, `>=8.4` are all supported.
Priority is still given to the valetrc files, and we pick the lowest matching version so `^8.3` could not suddenly switch to 8.6 once it comes out.
This is the same way [rectorphp/rector](https://github.com/rectorphp/rector/blob/a6a70cce35f29cb34a7874a518caf48b7f8d5fc5/src/Configuration/RectorConfigBuilder.php#L388) handles php constraints.

docs: https://github.com/laravel/docs/pull/11301

### DIFF
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -2,6 +2,7 @@
 
 namespace Valet;
 
+use Composer\Semver\Semver;
 use DateTime;
 use DomainException;
 use Illuminate\Support\Collection;
@@ -1163,5 +1164,43 @@ class Site
         $valetRc = $this->valetRc($siteName, $cwd);
 
         return PhpFpm::normalizePhpVersion(data_get($valetRc, 'php'));
+    }
+
+    /**
+     * Get PHP version from composer.json for a site.
+     */
+    public function phpComposerVersion(string $siteName, ?string $cwd = null): ?string
+    {
+        if ($cwd) {
+            $path = $cwd.'/composer.json';
+        } elseif ($site = $this->parked()->merge($this->links())->where('site', $siteName)->first()) {
+            $path = data_get($site, 'path').'/composer.json';
+        } else {
+            return null;
+        }
+
+        if (! $this->files->exists($path)) {
+            return null;
+        }
+
+        $composer = json_decode($this->files->get($path), true);
+        $constraint = data_get($composer, 'require.php', data_get($composer, 'require.php-64bit', data_get($composer, 'config.platform.php')));
+
+        if (empty($constraint)) {
+            return null;
+        }
+
+        // Find the lowest supported PHP version that satisfies the constraint (We do not suddenly want to raise the php version if a new one comes out)
+        $phpVersion = $this->brew->supportedPhpVersions()->reverse()->first(function ($formula) use ($constraint) {
+            if ($formula === 'php') {
+                return false;
+            }
+
+            $version = substr($formula, 4);
+
+            return Semver::satisfies($version, $constraint);
+        });
+
+        return $phpVersion ? PhpFpm::normalizePhpVersion($phpVersion) : null;
     }
 }
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -1186,7 +1186,7 @@ class Site
         $composer = json_decode($this->files->get($path), true);
         $constraint = data_get($composer, 'require.php', data_get($composer, 'require.php-64bit', data_get($composer, 'config.platform.php')));
 
-        if (empty($constraint)  || Semver::satisfies(substr($this->brew->linkedPhp(), 4), $constraint)) {
+        if (empty($constraint) || Semver::satisfies(substr($this->brew->linkedPhp(), 4), $constraint)) {
             return null;
         }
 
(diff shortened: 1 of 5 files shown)


cli/Valet/Site.php
```diff
@@ -1164,4 +1165,42 @@ public function phpRcVersion(string $siteName, ?string $cwd = null): ?string
...
+        }
+
+        if (! $this->files->exists($path)) {
+            return null;
+        }
+
+        $composer = json_decode($this->files->get($path), true);
+        $constraint = data_get($composer, 'require.php', data_get($composer, 'require.php-64bit', data_get($composer, 'config.platform.php')));
+
+        if (empty($constraint)) {
```
indykoning: If we wish to use the currently linked php version if it matches we could add
```suggestion
        if (empty($constraint)  || Semver::satisfies(substr($this->brew->linkedPhp(), 4), $constraint)) {
```

cli/app.php
```diff
@@ -705,7 +709,7 @@ function (ConsoleCommandEvent $event) {
         );
 
         if (! $phpVersion) {
-            $phpVersion = Site::phpRcVersion($site ?: basename(getcwd()));
+            $phpVersion = Site::phpRcVersion($site ?: basename(getcwd())) ?: Site::phpComposerVersion($site ?: basename(getcwd()));;
```
drbyte: Minor nit: You added a stray `;` to the end of this line, resulting in `;;`

indykoning: You're right, fixed!

cli/Valet/Site.php
drbyte: Seems sensible to me: no need to downgrade to a lower-detected one if the linked one satisfies already.

drbyte:
Looks good to me. 
Probably good to add tests. Maybe cover things like: simple `^8.3`, range like <8.5/>=8.4, missing/invalid composer.json, even invalid JSON.
I also note a minor styling point: `else if` will probably get reformatted by pint to `elseif` on merge.

Nice addition IMO.

indykoning:
I've added tests for missing and broken composer.json files
- One for when the current php version already matches the constraint
- One checking it goes for the lowest supported version when your version doesn't match the constraint
- A regular compound constraint
- A constraint your current version exceeds
- No supported versions matching the constraint

drbyte:
Tests look good. Thanks.
While reviewing, one edge-case came up: `Semver::satisfies()` throws `UnexpectedValueException` on a constraint string it can't parse (ie: a malformed non-semver version), instead of degrading to `null`. Perhaps it should wrap in a try/catch/return-null around the Semver::satisfies calls?

indykoning:
Good catch! While Composer should and does catch the malformed constraint we don't want Valet to crash because of it.
I agree that we should degrade to `null`, the second `Semver::satisfies` call in the loop shouldn't need the try/catch since it's using the same constraint, which now gets short-circuited earlier.

I've also added a test for it

drbyte:
Agreed.

Looks good to me.

//cc: @mattstauffer

drbyte:
@indykoning I ran this through Codex for further review, and it pointed out 2 concerns:

1. the PR code description and code comments/docs claim that exact patch constraints like 8.3.11 are supported, but it will actually fail with unpredicted results if that 3-level semver constraint is read from composer.json or require-php. So perhaps we should remove or qualify that claim? The rest of Valet only handles 2-level constraints, but now we're reading composer files that might have 3-level info, which we're not normalizing, and `satisfies()` fails in that case.

2. `phpComposerVersion()` returns null for an already satisfied linked PHP. Its suggested fix:

The Composer resolver should not return `null` when the currently linked PHP already satisfies the project constraint. That makes `valet isolate` and `valet link --isolate` think no Composer PHP version was found, even though there is a valid version to pin.

Recommended adjustment:

```php
if (empty($constraint)) {
    return null;
}

$linkedPhp = $this->brew->linkedPhp();

try {
    if (Semver::satisfies(substr($linkedPhp, 4), $constraint)) {
        return PhpFpm::normalizePhpVersion($linkedPhp);
    }
} catch (UnexpectedValueException $e) {
    return null;
}
```

This keeps `phpComposerVersion()` behaving as a resolver: if it can determine a usable PHP formula, it returns one. It also avoids a regression in bare `valet use`: if linked `php@8.4` already satisfies `^8.2`, returning the lowest matching version like `php@8.2` would cause Valet to downgrade global PHP unnecessarily. Returning the linked formula lets `valet use` hit its existing “already using” guard, while `isolate` and `link --isolate` receive a concrete version they can pin.

The existing test for “linked PHP satisfies composer constraint” should change from expecting `null` to expecting the linked formula, for example `php@8.4`.

indykoning:
I've updated the description, we cannot determine if the current php version matches a patch version. Nor would it make sense to do that.

I've updated the code to return the currently linked php version, a new case that might happen is a DomainException being thrown if no current php version could be determined. We return null in that case.

drbyte:
Nice catch on the `DomainException` issue. Should a test exercise `linkedPhp()` throwing, similarly?

Thanks for your work on this.

mattstauffer:
Thanks so much @indykoning! I'm going to add the last test @drbyte mentioned after merging so we can get this merged. And thanks as always @drbyte for the review!

PR #1559, opened 2026-08-11
fix(nginx): listen on [::1] so the IPv6 .test DNS answer from #1545 serves

### LINKED ISSUE #1558: .test sites return "404 page not found" when the OS resolves ::1 first — #1545 added IPv6 DNS but nginx stubs are IPv4-only
# Issue: `.test` sites 404 over IPv6 — #1545 added a `::1` DNS answer but nginx stubs only listen on `127.0.0.1`

## Title
`.test` sites return `404 page not found` when the OS resolves `::1` first — #1545 added IPv6 DNS but nginx stubs are IPv4-only

## Description

> _Disclaimer: this issue was filed by Claude Code, but the behavior and the
> suggested fix were verified by the user on their own machine._

#1545 ("Add IPV6 lookup for Safari Happy Eyeballs", commit `e559ad6`) changed
`DnsMasq::createDnsmasqTldConfigFile()` so dnsmasq now answers every `.test`
host with **both** an A record (`127.0.0.1`) and an AAAA record (`::1`):

```
address=/.test/127.0.0.1
address=/.test/::1        # added in #1545
listen-address=127.0.0.1
```

However, the generated nginx site configs still `listen` on IPv4 only — the
stubs in `cli/stubs/site.valet.conf` and `cli/stubs/secure.valet.conf` have no
`[::1]` listener. So when the OS follows Happy Eyeballs / RFC 6724 and connects
to `::1` first, nginx has nothing bound on the IPv6 loopback for that vhost. The
request falls through to Valet's catch-all server and returns:

```
404 page not found
```

The result: on machines where the resolver returns the AAAA (`::1`) record ahead
of the A record, **every `.test` site returns 404 in the browser** after #1545,
even though the site is correctly linked/secured and works fine over IPv4.
#1545 effectively shipped the IPv6-DNS half without the IPv6-listener half.

## Steps to reproduce

1. Valet 4.12.0 (any build containing #1545).
2. `valet secure my-app` (also reproduces with a plain parked/linked site).
3. Open `https://my-app.test/` in a browser, or curl with default resolution:

```
$ curl -skI https://my-app.test/
HTTP/2 404               # body: "404 page not found"

$ curl -skI --resolve my-app.test:443:127.0.0.1 https://my-app.test/
HTTP/2 200               # IPv4 works fine

$ curl -g -skI --resolve 'my-app.test:443:[::1]' https://my-app.test/
HTTP/2 404               # nginx isn't listening on ::1

$ dscacheutil -q host -a name my-app.test
name: my-app.test
ipv6_address: ::1        # returned first → browser uses it
name: my-app.test
ip_address: 127.0.0.1
```

## Expected

Since #1545 makes `.test` resolve to `::1`, nginx should also serve the vhost on
`[::1]`, so sites load regardless of which address family the OS picks.

## Suggested fix

Add an `[::1]` listener next to each `127.0.0.1` listener in the two nginx stubs,
e.g.:

```diff
 server {
     listen 127.0.0.1:443 ssl;
+    listen [::1]:443 ssl;
     #listen VALET_LOOPBACK:443 ssl; # valet loopback
```
(and the `:80` / `:60` server blocks likewise, in both `site.valet.conf` and
`secure.valet.conf`).

I patched the stubs locally, re-ran `valet secure`, and all three curl probes
above (default, forced IPv4, forced `[::1]`) then return **200** — the 404 is
gone over both stacks. Happy to open a PR if the approach looks right.

## Open question for maintainers

Should the IPv6 listener honor a custom `valet loopback` value? The current
`VALET_LOOPBACK` mechanism assumes IPv4, and a bracketed IPv6 form would need
separate handling. The simplest fix always adds a plain `[::1]` alongside
`127.0.0.1`, independent of the custom-loopback line — but you may want the two
to stay coupled. This is the main design decision, which is why I'm opening an
issue rather than going straight to a PR.

Possibly related: #1503 (`VALET_LOOPBACK` not updated in secure confs).

## Environment

- Valet: v4.12.0 (Composer global install)
- macOS: 26.2 (build 25C56)
- nginx: 1.27.3 (Homebrew)
- PHP: 8.4
- Verified `cli/Valet/DnsMasq.php` (installed) emits the `::1` line at the
  `createDnsmasqTldConfigFile()` method; `cli/stubs/{site,secure}.valet.conf`
  contain only `listen 127.0.0.1:...`.

Martin-Welte:
Fixes #1558.

## Problem

#1545 ("Add IPV6 lookup for Safari Happy Eyeballs") made dnsmasq answer every `.test` host with both `127.0.0.1` (A) and `::1` (AAAA):

```
address=/.test/127.0.0.1
address=/.test/::1
listen-address=127.0.0.1
```

But the nginx site/proxy stubs still `listen` on IPv4 only. When the OS follows Happy Eyeballs / RFC 6724 and connects to `::1` first, nginx has nothing bound there, so the request falls through to the catch-all and returns `404 page not found` for every site.

## Fix

Add a `listen [::1]:PORT;` line next to each `listen 127.0.0.1:PORT;` line in all four stubs (`site`, `secure`, `proxy`, `secure.proxy`), so nginx serves the same vhost over IPv6 loopback.

The new `[::1]` lines carry no `# valet loopback` comment, so neither `replaceLoopback()` nor `replaceOldLoopbackWithNew()` matches them — the custom-loopback mechanism is unaffected.

## Testing

Extended `test_can_install_nginx_site_config_for_specific_php_version` to assert the rendered config listens on both `127.0.0.1:80` and `[::1]:80`. Neither listen line was covered before, which is how the IPv4-only gap went unnoticed. Confirmed the assertion fails if the `[::1]` line is removed from the stub.

- Full suite green (283 tests)
- `pint` clean
- Verified end-to-end locally: with the patched stubs, a secured `.test` site returns `200` over both IPv4 and `[::1]`, where it previously 404'd over IPv6.

## Note for reviewers

With a custom loopback set (`valet loopback <ip>`), the rendered config now contains both the custom `listen <ip>:PORT;` line and the always-on `listen [::1]:PORT;` line. This is additive and matches #1545's intent of always answering IPv6, but if you'd prefer the `[::1]` listener to be gated by config instead of always-on, happy to adjust.

### DIFF
--- a/tests/SiteTest.php
+++ b/tests/SiteTest.php
@@ -989,7 +989,9 @@ class SiteTest extends TestCase
                 Mockery::on(function ($argument) {
                     return preg_match('/^# '.ISOLATED_PHP_VERSION.'=php@8.0/', $argument)
                         && preg_match('#fastcgi_pass "unix:.*/valet80.sock#', $argument)
-                        && strpos($argument, 'server_name site2.test www.site2.test *.site2.test;') !== false;
+                        && strpos($argument, 'server_name site2.test www.site2.test *.site2.test;') !== false
+                        && strpos($argument, 'listen 127.0.0.1:80;') !== false
+                        && strpos($argument, 'listen [::1]:80;') !== false;
                 }),
             ]);
 
--- a/cli/stubs/proxy.valet.conf
+++ b/cli/stubs/proxy.valet.conf
@@ -2,6 +2,7 @@
 
 server {
     listen 127.0.0.1:80;
+    listen [::1]:80;
     #listen VALET_LOOPBACK:80; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
@@ -47,6 +48,7 @@ server {
 
 server {
     listen 127.0.0.1:60;
+    listen [::1]:60;
     #listen VALET_LOOPBACK:60; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
--- a/cli/stubs/secure.proxy.valet-legacy.conf
+++ b/cli/stubs/secure.proxy.valet-legacy.conf
@@ -2,6 +2,7 @@
 
 server {
     listen 127.0.0.1:80;
+    listen [::1]:80;
     #listen VALET_LOOPBACK:80; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
@@ -9,6 +10,7 @@ server {
 
 server {
     listen 127.0.0.1:443 ssl http2;
+    listen [::1]:443 ssl http2;
     #listen VALET_LOOPBACK:443 ssl http2; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
--- a/cli/stubs/secure.proxy.valet.conf
+++ b/cli/stubs/secure.proxy.valet.conf
@@ -2,6 +2,7 @@
 
 server {
     listen 127.0.0.1:80;
+    listen [::1]:80;
     #listen VALET_LOOPBACK:80; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
@@ -9,6 +10,7 @@ server {
 
 server {
     listen 127.0.0.1:443 ssl;
+    listen [::1]:443 ssl;
     #listen VALET_LOOPBACK:443 ssl; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
--- a/cli/stubs/secure.valet-legacy.conf
+++ b/cli/stubs/secure.valet-legacy.conf
@@ -1,5 +1,6 @@
 server {
     listen 127.0.0.1:80;
+    listen [::1]:80;
     #listen VALET_LOOPBACK:80; # valet loopback
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     return 301 https://$host$request_uri;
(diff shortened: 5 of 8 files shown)


tests/SiteTest.php
```diff
@@ -989,7 +989,9 @@ public function test_can_install_nginx_site_config_for_specific_php_version()
                 Mockery::on(function ($argument) {
                     return preg_match('/^# '.ISOLATED_PHP_VERSION.'=php@8.0/', $argument)
                         && preg_match('#fastcgi_pass "unix:.*/valet80.sock#', $argument)
-                        && strpos($argument, 'server_name site2.test www.site2.test *.site2.test;') !== false;
+                        && strpos($argument, 'server_name site2.test www.site2.test *.site2.test;') !== false
+                        && strpos($argument, 'listen 127.0.0.1:80;') !== false
```
Martin-Welte, line 993: IPv4 was also not checked yet, i added both checks here now.

Martin-Welte:
noticed the legacy files should probably also contain the fix. Open question would be the valet.conf file.

drbyte:
Looks good to me. 

/cc @mattstauffer

drbyte:
Thinking about it further, yes, for thoroughness, I guess we probably should do `valet.conf` as well, to catch the "default" site.

drbyte:
One caveat/note: this change only affects "future" site-configs that get created. Or re-secured. I think that's fine. Most people for whom IPv4-only is working fine, won't need/miss IPv6 ... until they do. And then they can update/resecure their sites then as needed. **Not a blocker IMO.**

Martin-Welte:
yes, but also every fresh installation, right? This would make it hard for first time users, if the first thing they encounter is a 404 issue with DNS masq.

drbyte:
Yes, fresh installs would be provisioned from the outset, too.

mattstauffer:
Pushing a few minor changes right after merging because I don't have permissions to push to this PR, but this is great--thanks so much @Martin-Welte! And thanks as always for the review @drbyte!

PR #1562, opened 2026-08-19
Un-set variables to keep them out of user project scope

### LINKED ISSUE #1546: Variable $path on my project
### Description

When I call the $path variable in an empty file, it already has the value "./cli/Valet/Drivers/ValetDriver.php" filled in.

### Steps To Reproduce

Run valet for the folder, create an index.php file and call die($path); in it.

### Diagnosis

as a solution in the require-drivers.php file, call unset($path);

mattstauffer:
Fixes #1546, closes #1560.

@wakqasahmed thanks for your PR; but if we're going to unset them, we should unset them all. 

@drbyte can you think of any reason folks we know will have any issue with this? I consider these more a private API than a public API, so I don't really think of this as a breaking change, but just want to make sure there's not a practical use case we've heard of that I'm forgetting.

### DIFF
--- a/cli/includes/require-drivers.php
+++ b/cli/includes/require-drivers.php
@@ -8,3 +8,6 @@ foreach (scandir('./cli/Valet/Drivers') as $file) {
         require_once $path;
     }
 }
+
+// Unset our loop variables to keep them from leaking into the user's app.
+unset($file, $path);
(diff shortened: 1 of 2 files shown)


drbyte:
@mattstauffer From a cursory visual review, I think this makes sense.

And I agree: this is a private API. We don't need to be leaking things into global scope. If something needs whatever comes from this, it can glean it on its own.

mattstauffer:
@drbyte thanks my dude :)

PR #1568, opened 2026-08-24
Use PHP 8 features in CLI code

webpatser:
Follow-up to #1567 (require PHP 8.0). The CLI already guarantees PHP 8.0+ at runtime (`find-usable-php.php` enforces a `8.0` minimum), and parts of the CLI already use `str_contains` and constructor property promotion. This brings the rest of the CLI code in line, in two mechanical commits.

## Summary

- Replaces `strpos` comparisons with `str_contains`/`str_starts_with` in CLI code, only where the semantics are identical (`=== 0`, `!== 0`, `=== false`, `!== false`). Three call sites were deliberately left alone because they are not equivalent: the truthy checks in `PhpFpm::isolate()` and `Expose::findHttpTunnelUrl()`, and the offset use in `Status::jsonFromCli()`.
- Adds native property types to the untyped properties in the CLI classes (`Diagnose`, `DnsMasq`, `Ngrok`, `PhpFpm`, `Status`, `Valet`).

**Scope note:** everything loaded by `server.php` (`cli/Valet/Server.php`, `cli/Valet/Drivers/**`, `cli/includes/require-drivers.php`, `cli/includes/compatibility.php`) is untouched. That code runs under the site's FPM PHP without the Composer autoloader, and site isolation supports down to `php@7.1`, so it intentionally keeps using `strpos` and stays PHP 7-parseable.

The three `switch` statements in the service commands were also left as-is: their `case ''` relies on loose comparison to match an omitted (`null`) service argument, which a strict `match` would silently break.

## Test plan

- Full test suite passes after each commit: 293 tests, 561 assertions.
- `php cli/valet.php --version` and `php cli/valet.php paths` behave unchanged.
- `git diff --name-only` confirms no server-context file is modified.

### DIFF
--- a/cli/Valet/Diagnose.php
+++ b/cli/Valet/Diagnose.php
@@ -8,7 +8,7 @@ use Symfony\Component\Console\Output\ConsoleOutput;
 
 class Diagnose
 {
-    public $commands = [
+    public array $commands = [
         'sw_vers',
         'valet --version',
         'cat ~/.config/valet/config.json',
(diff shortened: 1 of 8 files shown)


mattstauffer:
Wonderful!

PR #1569, opened 2026-08-26
Add missing share server block to secure proxy stubs

webpatser:
## Summary

`valet share` tunnels to the plaintext server block on port 60. That block exists in `secure.valet.conf`, `secure.valet-legacy.conf`, and `proxy.valet.conf`, but both secure proxy stubs end without it, so sites created with `valet proxy --secure` cannot be shared.

- Adds the port-60 server block to `secure.proxy.valet.conf` and `secure.proxy.valet-legacy.conf`, mirroring the one in `proxy.valet.conf` (including the `X-Robots-Tag` header the share port sets everywhere else)
- Adds a test asserting both stubs contain the block and that `replaceOldLoopbackWithNew()` activates its custom loopback listener, matching the coverage added in #1561

## Test plan

- `./vendor/bin/phpunit` green (294 tests)
- Rendered the stub with all `VALET_*` tokens substituted and verified `nginx -t` reports `syntax is ok` on nginx 1.31.4

### DIFF
--- a/cli/stubs/secure.proxy.valet-legacy.conf
+++ b/cli/stubs/secure.proxy.valet-legacy.conf
@@ -57,3 +57,38 @@ server {
         deny all;
     }
 }
+
+server {
+    listen 127.0.0.1:60;
+    listen [::1]:60;
+    #listen VALET_LOOPBACK:60; # valet loopback
+    server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
+    root /;
+    charset utf-8;
+    client_max_body_size 128M;
+
+    add_header X-Robots-Tag 'noindex, nofollow, nosnippet, noarchive';
+
+    location /VALET_STATIC_PREFIX/ {
+        internal;
+        alias /;
+        try_files $uri $uri/;
+    }
+
+    access_log off;
+    error_log "VALET_HOME_PATH/Log/VALET_SITE-error.log";
+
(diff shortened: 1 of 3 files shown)


mattstauffer:
Could you catch up for the right number for `client_max_body_size` ? Thanks!

webpatser:
@mattstauffer good catch, that block was copied before the 512M bump landed. Pushed a follow-up setting `client_max_body_size 512M` in both secure proxy stubs so it matches the rest.

mattstauffer:
Thanks!! Unfortunately, we've gone stale *again* – we dropped the legacy stubs entirely, which makes this un-mergeable.

Could you rebase your branch onto the latest `master` and adjust for the legacy removal? Specifically:

1. Drop the changes to `cli/stubs/secure.proxy.valet-legacy.conf` entirely — that file no longer exists on `master`
2. Update the new test so it only iterates over `secure.proxy.valet.conf`

Thank you!

webpatser:
Hey @mattstauffer, rebased onto `main` and force-pushed. I dropped all the changes to `secure.proxy.valet-legacy.conf` since that stub is gone now, and the test only checks `secure.proxy.valet.conf`. Tests are green locally. Should be good to go, thanks for the nudge!

mattstauffer:
Wonderful, thanks @webpatser!

PR #1570, opened 2026-08-26
Match client_max_body_size to Valet's PHP upload limits

webpatser:
## Summary

`php-memory-limits.ini` sets `upload_max_filesize = 512M` / `post_max_size = 512M` and even carries the comment that changes need to be reflected in nginx with `client_max_body_size`. But only the top-level `nginx.conf` and the HTTPS block of `secure.valet.conf` say 512M; every other server block (plain HTTP sites, isolated sites, proxies, and the share blocks) still caps at 128M.

Net effect: an upload between 128M and 512M succeeds on `https://site.test` but returns 413 on `http://site.test`, even though PHP would happily accept it.

- Sets `client_max_body_size 512M;` consistently across all server blocks in the site, proxy, and legacy stubs

## Test plan

- Rendered the stubs with all `VALET_*` tokens substituted and verified `nginx -t` reports `syntax is ok` on nginx 1.31.4

### DIFF
--- a/cli/stubs/proxy.valet.conf
+++ b/cli/stubs/proxy.valet.conf
@@ -7,7 +7,7 @@ server {
     server_name VALET_SITE www.VALET_SITE *.VALET_SITE;
     root /;
     charset utf-8;
-    client_max_body_size 128M;
+    client_max_body_size 512M;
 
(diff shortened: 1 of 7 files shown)


mattstauffer:
Thank you! Great catch!

PR #1571, opened 2026-08-26
Drop legacy nginx stubs for pre-1.25.1 versions

webpatser:
## Summary

The `-legacy` stubs exist for nginx older than 1.25.1 (released June 2023) and differ from their modern twins by three lines. They still carry `http2_push_preload`, which was removed entirely in nginx 1.25.1, so a stale legacy site config is a hard `nginx -t` error (unknown directive) on any current nginx. Homebrew has shipped nginx >= 1.25.1 for over three years, and the duplication is exactly how the two stub variants drift apart over time (see #1569).

- Removes `secure.valet-legacy.conf` and `secure.proxy.valet-legacy.conf`
- Removes the per-site `exec('nginx -v')` version sniffing in `Site.php`. Beyond being dead weight, that check resolved `nginx` from `$PATH` (not the Homebrew binary Valet actually runs) and fell back to the legacy stub when nginx wasn't installed yet, writing a config the current nginx then rejects

Same spirit as #1567: trim compatibility paths nothing supported still needs.

## Test plan

- `valet secure` and `valet proxy --secure` paths still resolve their stubs via `Filesystem::getStub()`, so the custom-stub override in `~/.config/valet/stubs/` keeps working


mattstauffer:
Brilliant, thanks!

PR #1573, opened 2026-08-26
Modernize gzip configuration

webpatser:
## Summary

The `gzip_types` list in `nginx.conf` predates a few MIME registrations and misses types common in today's Laravel apps:

- `.xml` files are emitted as `text/xml` by nginx's own `mime.types`, but the list only had `application/xml`, so static XML (sitemaps, feeds) was never compressed
- Adds `text/javascript` (nginx has flip-flopped between it and `application/javascript` across releases; listing both covers either mapping), `application/wasm`, `application/ld+json`, and `application/manifest+json`
- Swaps the legacy names for their registered successors: `application/x-font-ttf` -> `font/ttf`, `font/opentype` -> `font/otf`, `application/x-web-app-manifest+json` -> `application/manifest+json`, and drops `text/x-component` (IE HTC behaviors)

Also enables `gzip_static`, which is compiled into the Homebrew nginx build (`--with-http_gzip_static_module`) and serves precompressed `.gz` siblings when a build step emits them, at zero cost otherwise.

## Test plan

- Rendered `nginx.conf` with tokens substituted, `nginx -t` reports `syntax is ok` on nginx 1.31.4

### DIFF
--- a/cli/stubs/nginx.conf
+++ b/cli/stubs/nginx.conf
@@ -20,6 +20,7 @@ http {
     ssi on;
 
     gzip  on;
+    gzip_static on;
     gzip_comp_level 5;
     gzip_min_length 256;
     gzip_proxied any;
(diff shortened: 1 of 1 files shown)


mattstauffer:
I think the only changes in here that I would change, as they do help with parity toward production, are the `text/xml` and `application/wasm`. The rest takes us further from production parity or adds nothing.

webpatser:
@mattstauffer fair enough. Trimmed the PR down to just adding `text/xml` and `application/wasm` to `gzip_types`; everything else is reverted so the list stays where it was.

mattstauffer:
Thanks!

PR #1574, opened 2026-08-26
Make dnsmasq authoritative for the Valet TLD

webpatser:
## Summary

The generated `tld-*.conf` only sets `address=` entries, which cover A and AAAA lookups. Any other record type for a `.test` name is not answered locally and gets forwarded to the upstream servers from `/etc/resolv.conf`. Safari and current Chrome send an HTTPS (type 65) query alongside A/AAAA on every navigation, so each fresh `.test` lookup leaks to public resolvers and waits on a round trip (or a timeout, if the upstream is unreachable) before the browser settles. Same spirit as #1545, which fixed the AAAA half of this.

- Adds `local=/tld/` so dnsmasq answers every record type for the Valet TLD authoritatively, with an immediate local answer and nothing forwarded upstream
- Switches `address=/.test/` to the canonical `address=/test/` form, matching the `local=` syntax (dnsmasq treats both the same; the docs use the dotless form)

Scoped deliberately to the per-TLD config file: no changes to the shared `dnsmasq.conf` or the `dnsmasq.d` layout, so setups where the same dnsmasq serves other tools (Herd, custom forwarding) are untouched.

## Test plan

- `./vendor/bin/phpunit` green, including the updated `DnsMasqTest` content assertion
- `dnsmasq --test` reports `syntax check OK` on the generated config (dnsmasq 2.93)

### DIFF
--- a/cli/Valet/DnsMasq.php
+++ b/cli/Valet/DnsMasq.php
@@ -110,8 +110,9 @@ class DnsMasq
         $loopback = $this->configuration->read()['loopback'];
 
         $this->files->putAsUser($tldConfigFile,
-            'address=/.'.$tld.'/'.$loopback.PHP_EOL
-            .'address=/.'.$tld.'/::1'.PHP_EOL // IPV6 loopback prevents Safari "Happy Eyeballs" slow load
+            'address=/'.$tld.'/'.$loopback.PHP_EOL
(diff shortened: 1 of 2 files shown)


mattstauffer:
Great update!

PR #1578, opened 2026-09-09
Detect PHP installed from non-core Homebrew taps

drbyte:
Follow-up to [#1577](https://github.com/laravel/valet/discussions/1577), where `valet use` tries to install a PHP version that is already installed from `shivammathur/php`.

Valet's install logic is already tap-agnostic in intent — `ensureInstalled()` short-circuits on `installed()` before it ever taps or applies the `shivammathur/php/` prefix, and `brew link` / `brew unlink` / the FPM config paths all use bare formula names, which are the same whatever tap a keg came from. The gap is upstream of that: **both of Valet's "is it installed?" probes assume homebrew/core.**

### `Brew::installed()`

Runs `brew info <formula> --json=v2` with an unqualified name. Homebrew resolves that against homebrew/core first, so the core formula can report an empty `installed` array while a keg installed from another tap occupies the rack. `ensureInstalled()` then proceeds to `brew install php@8.4` from core.

It now falls back to checking for a non-empty keg in `$(brew --prefix)/Cellar/<formula>`, which is tap-agnostic. Casks are unaffected — the fallback only applies to formulae.

### `Brew::installedPhpFormulae()`

Matches `brew list --formula` output against `SUPPORTED_PHP_VERSIONS` with an exact `contains()`, so a fully-qualified `shivammathur/php/php@8.4` never matches and `hasInstalledPhp()` returns `false` — which sends `PhpFpm::install()` off to install `php` on top of a perfectly good PHP. Tap prefixes are now stripped.

### Notes

- No behaviour change for the ordinary homebrew/core path: the Cellar check is only consulted when the JSON says not-installed.
- This deliberately does **not** touch which tap Valet installs *from* when a version is genuinely missing — `LIMITED_PHP_VERSIONS` still routes to `shivammathur/php`.
- 11 new tests in `BrewTest`. The pre-existing `CliTest` failures on this branch also fail on `master` in the same environment.

@dpletiko — if you're able to test this against your `shivammathur/php` install, that would be very helpful:

```bash
composer global config repositories.valet vcs https://github.com/drbyte/valet
composer global require laravel/valet:dev-tap-agnostic-php-detection
```

Then `valet use php@8.4` (or whichever version you have from the tap) should link the existing install rather than trying to install anything. To go back afterwards:

```bash
composer global config --unset repositories.valet
composer global require laravel/valet
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cli/Valet/Brew.php
+++ b/cli/Valet/Brew.php
@@ -51,20 +51,20 @@ class Brew
 
         // should be a json response, but if not installed then "Error: No available formula ..."
         if (starts_with($result, 'Error: No')) {
-            return false;
+            return $this->hasInstalledKeg($formula);
         }
 
         $details = json_decode($result, true);
 
         if (! empty($details['formulae'])) {
-            return ! empty($details['formulae'][0]['installed']);
+            return ! empty($details['formulae'][0]['installed']) || $this->hasInstalledKeg($formula);
         }
 
         if (! empty($details['casks'])) {
             return ! is_null($details['casks'][0]['installed']);
         }
 
-        return false;
+        return $this->hasInstalledKeg($formula);
     }
 
     /**
@@ -102,7 +102,7 @@ class Brew
     {
         return collect(
             explode(PHP_EOL, $this->cli->runAsUser('brew list --formula | grep php'))
-        );
+        )->map(fn ($formula) => $this->formulaName($formula));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mattstauffer:
Thanks for the PR! I don't love the structure of the returns. What if `installed` was simplified?

Something like this:

```php
public function installed()
{
    return $this->hasInstalledCore() || $this->hasInstalledKeg();
}
```

... or something similar. Then we're not falling back to `hasInstalledKeg` three different times.

drbyte:
Good call — that reads much better. Pushed in 164723c.

`installed()` is now just:

```php
public function installed(string $formula): bool
{
    return $this->hasInstalledFormulaOrCask($formula) || $this->hasInstalledKeg($formula);
}
```

I named the first one `hasInstalledFormulaOrCask` rather than `hasInstalledCore`, because `brew info` resolves against any tapped tap and not only homebrew/core — so "core" would be a little misleading about what it actually answers. Happy to rename if you prefer something shorter.

Its body is byte-identical to what `installed()` had before this PR, so the diff against `master` is now just the new one-line `installed()` plus the two new helpers.

One behaviour change worth flagging, since the `||` removes the early return: **casks now consult the Cellar fallback too.** Previously the cask branch returned straight away. In practice a cask and a keg sharing a name isn't a case that arises for anything Valet manages, and "there's a keg on disk" is a correct answer to "is this installed" regardless — but it is a real difference from the three-branch version, so I've swapped the test that asserted the old early-return for one covering the not-installed cask path, plus a short-circuit test proving the Cellar check is skipped when Homebrew already says yes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mattstauffer:
Thanks @drbyte!

xpletiko:
@drbyte Thanks, nice one. 
Cloned the main branch and tried version switching and after a small hickup verified that versions actually switch.

The small hickup was from the usage of `runAsUser` command (`CommandLine`). 

The `sudo -u user()` invocation, even though it uses the same user, loses the `XDG_CONFIG_HOME` variable from the env (at least for me - don't even recall why it exists). 
That, in part, breaks the `valet use` command because brew will check the trusted taps/formulae. Since the `XDG_CONFIG_HOME` variable is unset now (after user switching - even to the same user) it will lookup a different path for the `trust.json` trusted entries and fail the command since the entries are now empty/different/missing.

> Trust non-official tap formulae, casks or commands so Homebrew may load them.
Trusted entries are stored in ${XDG_CONFIG_HOME}/homebrew/trust.json if
$XDG_CONFIG_HOME is set or ~/.homebrew/trust.json otherwise.

Re-trusting all entries with the same prefix as the valet user switching fixes the `valet use` command since now both paths will have the same trusted entries:
`sudo -u "$(whoami)" brew trust shivammathur/php` 
or formulae
`sudo -u "$(whoami)" brew trust  shivammathur/php/php@8.4`

PR #1579, opened 2026-09-18
Allow valet which to accept a site name or directory path

wppoland:
### Summary
Previously, `valet which` only checked the current working directory (`getcwd()`). If a developer was working in another folder or needed to check which Valet driver serves another project, they had to navigate to that directory first.

This PR updates `valet which [site]` to optionally accept:
- A registered site name (from parked or linked sites, e.g. `valet which my-site` or `valet which my-site.test`)
- An explicit directory path (e.g. `valet which /path/to/my-site`)
- Omitting the argument preserves full backward compatibility by checking the current working directory as before.

### Changes
- Added `Site::getSitePath(string $siteName): ?string` to cleanly look up paths for site names and strip the configured `.tld` when provided.
- Refactored internal lookups in `Site.php` (`getSiteUrl`, `valetRc`, `phpRcVersion`, `phpComposerVersion`) to use the new `getSitePath` helper instead of repeating collection lookups.
- Updated `valet which [site]` in `cli/app.php` to resolve the path for the given site or folder.
- Added comprehensive unit tests in `tests/CliTest.php` and `tests/SiteTest.php`.


drbyte (commented):
Thanks for this — consolidating the four `parked()->merge($this->links())->where('site', …)` lookups into one helper is a good change. A few things to sort out before it can go in, roughly in order of importance.

### 1. Directory check runs before the registered-site lookup

`cli/app.php` checks `Filesystem::isDir($site)` first, so a bare site name is resolved as a path relative to the current directory before Valet's site list is consulted. If cwd happens to contain a subdirectory with the same name as a registered site, `valet which my-site` silently reports the driver for that unrelated subdirectory. (I confirmed this with an empty `my-best-site/` in cwd alongside the parked fixture.)

The PR description lists "registered site name" first, which is the order I'd expect: look the name up in the site list, and only fall back to treating it as a path if that fails (or treat it as a path up front only when it contains a `/` or is `.`/`..`).

### 2. TLD stripping now lives in three places, and `getSiteUrl` double-strips

`getSitePath()` copies the `.`/`./` and `.tld` handling from `getSiteUrl()` rather than extracting it, so:

- `getSiteUrl()` still strips the TLD itself, then calls `getSitePath()`, which strips it again. That's a behaviour change: `getSiteUrl('my-best-site.test.test')` now returns `my-best-site.test.test` (matched against the bare `my-best-site`), where `master` throws `The [my-best-site.test] site could not be found`. `getSiteUrl` feeds `PhpFpm::isolateDirectory()`/`unIsolateDirectory()`, so a typo like `valet isolate --site=foo.test.test` would now write an nginx config for a hostname Valet never serves instead of failing.
- `which` in `app.php` re-reads the config and strips the TLD a third time (`str_ends_with` here, `ends_with()` helper everywhere else) just to recover the name that `getSitePath()` already computed internally.

Suggestion: one `normalizeSiteName(string $name): string` on `Site` that handles `.`/`./` and strips the TLD once, used by `getSiteUrl`, `getSitePath` and `which`. `getSiteUrl` then becomes roughly

```php
$name = $this->normalizeSiteName($directory);

if (! $this->getSitePath($name)) {
    throw new DomainException("The [{$name}] site could not be found in Valet's site list.");
}

return $name.'.'.$tld;
```

and `which` can derive its display name from the same normaliser (or from `Site::host($path)`, see 5) instead of parsing the TLD itself.

### 3. `test_it_can_get_site_path` shells out to `sudo`

The new `SiteTest` test calls the real `Site::link()`, which goes through `CommandLine::runAsUser()` → `sudo -u <user> ln -s …`. It's the only test in the suite that invokes `sudo`. On a CI runner with passwordless sudo it passes, but on a contributor's machine without a cached sudo timestamp the Symfony `Process` has no TTY, `sudo` can't prompt, `runCommand()`'s no-op `$onError` swallows the failure, and the two `my-linked-site` assertions fail with `null`. When it does succeed it leaves `tests/config/valet/Sites/my-linked-site` and a prepended `Sites` entry in `config.json` behind for every later test in the process (`SiteTest::tear_down` only clears `tests/output`).

Creating the link directly (`symlink()` or `Filesystem::symlink()` plus `Configuration::prependPath()`), or mocking `CommandLine`, avoids all of that.

### 4. `valet which <name>` now inherits `parked()`/`links()` side effects

Resolving a name via `getSitePath()` walks `parked()->merge($this->links())`, which runs `getSites()` and therefore `Brew::linkedPhp()` for every site plus `ensureDirExists()` for every configured path. The no-argument `valet which` never touched Brew; with an argument it now aborts with `Homebrew PHP appears not to be linked` when PHP is unlinked (which is exactly how the new `test_which_command_with_site_name` fails on a machine without Homebrew PHP linked) and creates any missing parked directories as a side effect. A name lookup only needs the `Sites/` symlinks and the parked directory names, so a lighter resolver that doesn't go through `getSites()`/`getPhpVersion()` would be preferable.

### 5. Smaller points

- In the directory branch, `$siteName = basename($path)`, but `Site::host($path)` already exists for exactly this ("the real hostname for the given path, checking links") and is what `which-php` uses. For a linked site the two differ: with the fixture linked as `aliased`, `valet which aliased` prints `The [aliased] site…` while `valet which /path/to/my-best-site` prints `The [my-best-site] site…` and passes a different `$siteName` to `ValetDriver::assign()` for the same project.
- `getSitePath()` returns `getSites()`'s `path`, which for links is the raw `readlink()` target, not `realpath()`. `valet link` always writes an absolute target, but a hand-made relative symlink in `~/.config/valet/Sites` would make `which` evaluate the driver checks relative to the caller's cwd. Worth a `realpath()` before handing it to `assign()`.
- No CLI-level test exercises the `.test` form (`valet which my-best-site.test`), which is the only input that hits the TLD strip in `app.php`, nor the cwd-shadowing case from point 1.

Also a heads-up that the `tests` workflow hasn't run on this PR yet, so none of the new tests have been exercised in CI.

wppoland:
Thank you for the detailed review and edge-case catches @drbyte!

All points addressed in 5747206:
1. **Lookup precedence:** `cli/app.php` now checks registered sites via `Site::getSitePath($site)` first before falling back to directories (unless an explicit path containing `/`, `.`, or `..` was provided).
2. **TLD stripping & normalisation:** Extracted `Site::normalizeSiteName(string $name): string` which handles `.`/`./` and single-pass TLD stripping. Used consistently across `getSiteUrl`, `getSitePath`, and `which`.
3. **No sudo in tests:** `test_it_can_get_site_path` now creates the symlink directly via `$files->symlink()` and cleans it up in a `finally` block, avoiding any subprocess or sudo prompts.
4. **Lightweight resolver:** `Site::getSitePath()` now checks `$this->sitesPath()` links and configured parked paths directly on the filesystem before falling back, avoiding calls to `parked()`, `links()`, `getSites()`, or `Brew`.
5. **Additional polish & tests:**
   - Used `Site::host($path)` for directory-based naming.
   - Ensured `realpath()` resolution on paths.
   - Added tests for the `.test` form (`my-best-site.test`) and for preferring registered sites over cwd subdirectories with the same name.

All 311 unit tests pass cleanly.

drbyte:
While looking at the changes I had a few concerns (including with my last suggestion 😬 ), so I asked Claude to confirm or refute them for me.  I'll post the feedback below...

drbyte (commented):
Thanks for turning this round so quickly — you addressed every point, and two of the fixes are better than what I suggested.

## What landed well

- **`$isExplicitPath`** — I proposed the `str_contains($site, '/')` heuristic but *not* `$site === '..'`, and that guard turns out to be load-bearing: without it `getSitePath('..')` resolves to a real directory and `valet which ..` reports a driver for something inside `~/.config/valet`. Good catch on a hole in my suggestion.
- **`Site::host()` in the no-argument branch too.** I only flagged the directory branch; the `basename(getcwd())` case was wrong the same way and I missed it.
- **The sudo removal is idiomatic, not just compliant** — `Filesystem::symlink` + `ensureDirExists` + `$site->sitesPath('name')` with `try`/`finally`, rather than a raw `symlink()`. The new tests clean up after themselves too.
- **The direct-filesystem `getSitePath()` goes beyond the ask and measurably works.** `SiteTest::test_it_can_read_valet_rc_files` fails on `master` and passes on this branch on a machine with no linked Homebrew PHP — you fixed an unrelated pre-existing test as a side effect.
- `normalizeSiteName()` is the right abstraction with the right name. Everything below is about how many places call it, not the helper.

## The double-TLD strip: my earlier comment was under-specified, sorry

I handed you `$name = $this->normalizeSiteName($directory); if (! $this->getSitePath($name))` and never said that `getSitePath()` has to *stop* normalizing. You implemented the snippet faithfully, and because `getSitePath()` still normalizes, the two compose:

`getSiteUrl('my-best-site.test.test')` -> `'my-best-site.test'` -> `getSitePath()` strips again -> `'my-best-site'` -> found -> returns `'my-best-site.test.test'`.

Verified on 5747206; `master` throws `The [my-best-site.test] site could not be found` here. `valet isolate --site=foo.test.test` therefore writes an nginx config for a hostname Valet never serves.

## Two things came in with the rewrite

Both only show up if you run the resolvers side by side, so I don't think reading would have caught either.

**1. Parked-path precedence is inverted.** `array_reverse(...)` + return-on-first-match makes the *last* configured path win. `parked()` reverses and then `merge`s — later merges overwrite — so the *first* path wins there; and `Server::sitePath()`, the code that actually serves the request, loops forward and returns the first match. With two parked paths `pa` then `pb`, each containing `dupe/`:

```
getSitePath('dupe')        -> pb/dupe
parked() / Server::sitePath('dupe') -> pa/dupe
```

Nginx serves `pa/dupe`, so `valetRc()` / `phpRcVersion()` / `phpComposerVersion()` now read `.valetrc` and `composer.json` out of the directory Valet *doesn't* serve — that reaches `valet use` and `valet isolate` auto-detection, not just `which`.

**2. `isDir($base.'/'.$siteName)` accepts path segments,** so `getSitePath()` no longer proves the name is a registered site — which is the one thing `getSiteUrl()` uses it for. Verified: `getSitePath('..')` returns a parked path's parent, `getSitePath('')` returns the `Sites` directory itself once it exists, and `getSiteUrl('..')` returns `'...test'` where `master` threw — so `valet isolate --site=..` would write an nginx config named `...test`.

## Why the legacy lookup probably couldn't just be deleted

I suspect you kept `parked()->merge($this->links())` as a safety net. Two things about it:

- It can only run when the new scan already failed, so it never rescues a hit — but it runs on every *miss*, which is exactly the `getSites()` -> `getPhpVersion()` -> `Brew::linkedPhp()` path the rewrite was meant to avoid. Your own new assertion trips over it: `assertNull($site->getSitePath('non-existent-site'))` errors with `Homebrew PHP appears not to be linked` (`SiteTest:1190` -> `Site.php:229` -> `getSites` -> `getPhpVersion` -> `Brew::linkedPhp`), so `test_it_can_get_site_path` is a new failure versus `master` on any machine without a linked brew PHP. In real use `valet which <typo>` aborts with that exception instead of printing your "could not find a site or directory" warning.
- But deleting it breaks `test_gets_site_url_from_directory` — that test mocks `parked()`/`links()` and passes `paths => []`, so **every one of its assertions resolves through the fallback**. Which means the rewrite currently has no `getSiteUrl()` coverage at all, and the test that looks like coverage is pinning the code path you were removing. That's a real blocker, not an oversight, and it's why I wrote the test change rather than asking you to.

## Patch

Verified on this branch: full suite has nothing newly broken, `test_it_can_get_site_path` and `test_it_can_read_php_rc_version` now pass, all six `which` tests stay green (five of yours, one rewritten below), `vendor/bin/pint --test` passes, and it syntax-checks on 8.0/8.1/8.2/8.3/8.5. Normalize once at the public boundary, iterate `paths` in the server's order, reject names that aren't site names, and only trust a link whose target is a directory:

```diff
diff --git a/cli/Valet/Site.php b/cli/Valet/Site.php
index 5461811..9529469 100644
--- a/cli/Valet/Site.php
+++ b/cli/Valet/Site.php
@@ -171,7 +171,7 @@ public function getSiteUrl(string $directory): string
         $tld = $this->config->read()['tld'];
         $name = $this->normalizeSiteName($directory);
 
-        if (! $this->getSitePath($name)) {
+        if (! $this->resolveSitePath($name)) {
             throw new DomainException("The [{$name}] site could not be found in Valet's site list.");
         }
 
@@ -201,34 +201,45 @@ public function normalizeSiteName(string $siteName): string
      */
     public function getSitePath(string $siteName): ?string
     {
-        $siteName = $this->normalizeSiteName($siteName);
+        return $this->resolveSitePath($this->normalizeSiteName($siteName));
+    }
+
+    /**
+     * Resolve an already-normalized site name to its directory.
+     */
+    protected function resolveSitePath(string $siteName): ?string
+    {
+        if ($siteName === '' || $siteName === '.' || $siteName === '..' || str_contains($siteName, '/')) {
+            return null;
+        }
 
         // 1. Check linked sites in ~/.config/valet/Sites
-        $linkPath = $this->sitesPath().'/'.$siteName;
+        $linkPath = $this->sitesPath($siteName);
+
         if ($this->files->isLink($linkPath)) {
-            return $this->files->realpath($this->files->readLink($linkPath));
-        }
+            $target = $this->files->readLink($linkPath);
 
-        if ($this->files->isDir($linkPath)) {
+            if ($this->files->isDir($target)) {
+                return $this->files->realpath($target);
+            }
+        } elseif ($this->files->isDir($linkPath)) {
             return $this->files->realpath($linkPath);
         }
 
-        // 2. Check parked paths
-        $config = $this->config->read();
-        foreach (array_reverse($config['paths']) as $path) {
+        // 2. Check parked paths, in the same order the server resolves them
+        foreach ($this->config->read()['paths'] as $path) {
             if ($path === $this->sitesPath()) {
                 continue;
             }
 
             $candidate = $path.'/'.$siteName;
+
             if ($this->files->isDir($candidate)) {
                 return $this->files->realpath($candidate);
             }
         }
 
-        $site = $this->parked()->merge($this->links())->where('site', $siteName)->first();
-
-        return data_get($site, 'path');
+        return null;
     }
 
     /**
```

<details>
<summary>Tests (four in <code>SiteTest</code>, plus a rewrite of the shadowing test) — click to expand</summary>

I mutation-tested these rather than trusting that they pass: gutting the parked loop, gutting the link branch, re-adding the second `normalizeSiteName()` call, dropping the name guard, re-adding `array_reverse`, and dropping the `isDir($target)` check each make one of them fail. (Dropping the `$path === $this->sitesPath()` skip is caught by nothing, because it's behaviourally neutral — the link branch already checked that directory.)

`test_gets_site_url_from_directory` moves onto `Filesystem` mocks, keeping all eight of its original assertions, with `site2`'s link target deliberately placed *outside* the parked paths so the link lookup is actually pinned (my first attempt wasn't — the parked loop rescued it).

```diff
diff --git a/tests/SiteTest.php b/tests/SiteTest.php
index 4c64d7c..150b946 100644
--- a/tests/SiteTest.php
+++ b/tests/SiteTest.php
@@ -774,6 +774,10 @@ public function test_gets_site_url_from_directory()
 
         swap(Configuration::class, $config);
 
+        $files = Mockery::mock(Filesystem::class);
+
+        swap(Filesystem::class, $files);
+
         $siteMock = Mockery::mock(Site::class, [
             resolve(Brew::class),
             resolve(Configuration::class),
@@ -784,32 +788,29 @@ public function test_gets_site_url_from_directory()
         swap(Site::class, $siteMock);
 
         $config->shouldReceive('read')
-            ->andReturn(['tld' => 'test', 'loopback' => VALET_LOOPBACK, 'paths' => []]);
+            ->andReturn(['tld' => 'test', 'loopback' => VALET_LOOPBACK, 'paths' => ['/Users/name/code']]);
 
-        $siteMock->shouldReceive('parked')
-            ->andReturn(collect([
-                'site1' => [
-                    'site' => 'site1',
-                    'secured' => '',
-                    'url' => 'http://site1.test',
-                    'path' => '/Users/name/code/site1',
-                ],
-            ]));
+        // site1 and portal.test-site are parked; site2 is linked to a directory
+        // outside the parked paths, so only the link lookup can resolve it.
+        $files->shouldReceive('isLink')->andReturnUsing(function ($path) {
+            return $path === VALET_HOME_PATH.'/Sites/site2';
+        });
 
-        $siteMock->shouldReceive('links')->andReturn(collect([
-            'site2' => [
-                'site' => 'site2',
-                'secured' => 'X',
-                'url' => 'http://site2.test',
-                'path' => '/Users/name/code/site2',
-            ],
-            'portal.test-site' => [
-                'site' => 'portal.test-site',
-                'secured' => 'X',
-                'url' => 'http://portal.test-site.test',
-                'path' => '/Users/name/code/portal.test-site',
-            ],
-        ]));
+        $files->shouldReceive('readLink')
+            ->with(VALET_HOME_PATH.'/Sites/site2')
+            ->andReturn('/Users/name/elsewhere/site2');
+
+        $files->shouldReceive('isDir')->andReturnUsing(function ($path) {
+            return in_array($path, [
+                '/Users/name/code/site1',
+                '/Users/name/code/portal.test-site',
+                '/Users/name/elsewhere/site2',
+            ]);
+        });
+
+        $files->shouldReceive('realpath')->andReturnUsing(function ($path) {
+            return $path;
+        });
 
         $siteMock->shouldReceive('host')->andReturn('site1');
 
@@ -826,6 +827,145 @@ public function test_gets_site_url_from_directory()
 
         $this->assertEquals('portal.test-site.test', $site->getSiteUrl('portal.test-site'));
         $this->assertEquals('portal.test-site.test', $site->getSiteUrl('portal.test-site.test'));
+
+        $this->assertEquals('/Users/name/code/site1', $site->getSitePath('site1'));
+        $this->assertEquals('/Users/name/elsewhere/site2', $site->getSitePath('site2.test'));
+    }
+
+    public function test_it_rejects_site_names_that_are_not_valet_sites()
+    {
+        $config = Mockery::mock(Configuration::class);
+
+        swap(Configuration::class, $config);
+
+        $files = Mockery::mock(Filesystem::class);
+
+        swap(Filesystem::class, $files);
+
+        $siteMock = Mockery::mock(Site::class, [
+            resolve(Brew::class),
+            resolve(Configuration::class),
+            resolve(CommandLine::class),
+            resolve(Filesystem::class),
+        ])->makePartial();
+
+        swap(Site::class, $siteMock);
+
+        $config->shouldReceive('read')
+            ->andReturn(['tld' => 'test', 'loopback' => VALET_LOOPBACK, 'paths' => ['/Users/name/code']]);
+
+        $files->shouldReceive('isLink')->andReturn(false);
+        $files->shouldReceive('isDir')->andReturnUsing(function ($path) {
+            // Everything that exists on this fake disk: the parked site, and the
+            // directories a traversing name would reach.
+            return in_array($path, [
+                '/Users/name/code/site1',
+                '/Users/name/code',
+                VALET_HOME_PATH.'/Sites',
+                VALET_HOME_PATH,
+            ]);
+        });
+        $files->shouldReceive('realpath')->andReturnUsing(function ($path) {
+            return $path;
+        });
+
+        $site = resolve(Site::class);
+
+        // A doubled TLD must not resolve to the bare site, and a name that walks
+        // out of the site directories is not a site at all.
+        foreach (['site1.test.test', '..', '', 'code/site1', './site1'] as $notASite) {
+            $this->assertNull($site->getSitePath($notASite), "[{$notASite}] should not resolve to a path.");
+
+            try {
+                $site->getSiteUrl($notASite);
+
+                $this->fail("Expected [{$notASite}] to be rejected by getSiteUrl().");
+            } catch (DomainException $e) {
+                $this->assertStringContainsString('could not be found', $e->getMessage());
+            }
+        }
+    }
+
+    public function test_it_resolves_duplicate_site_names_in_parked_path_order()
+    {
+        $config = Mockery::mock(Configuration::class);
+
+        swap(Configuration::class, $config);
+
+        $files = Mockery::mock(Filesystem::class);
+
+        swap(Filesystem::class, $files);
+
+        $siteMock = Mockery::mock(Site::class, [
+            resolve(Brew::class),
+            resolve(Configuration::class),
+            resolve(CommandLine::class),
+            resolve(Filesystem::class),
+        ])->makePartial();
+
+        swap(Site::class, $siteMock);
+
+        $config->shouldReceive('read')->andReturn([
+            'tld' => 'test',
+            'loopback' => VALET_LOOPBACK,
+            'paths' => ['/Users/name/first', '/Users/name/second'],
+        ]);
+
+        $files->shouldReceive('isLink')->andReturn(false);
+        $files->shouldReceive('isDir')->andReturnUsing(function ($path) {
+            return in_array($path, ['/Users/name/first/dupe', '/Users/name/second/dupe']);
+        });
+        $files->shouldReceive('realpath')->andReturnUsing(function ($path) {
+            return $path;
+        });
+
+        $site = resolve(Site::class);
+
+        // The server resolves the first registered path first (see Server::sitePath),
+        // so a duplicated site name must resolve the same way here.
+        $this->assertEquals('/Users/name/first/dupe', $site->getSitePath('dupe'));
+    }
+
+    public function test_it_ignores_a_link_whose_target_is_missing()
+    {
+        $config = Mockery::mock(Configuration::class);
+
+        swap(Configuration::class, $config);
+
+        $files = Mockery::mock(Filesystem::class);
+
+        swap(Filesystem::class, $files);
+
+        $siteMock = Mockery::mock(Site::class, [
+            resolve(Brew::class),
+            resolve(Configuration::class),
+            resolve(CommandLine::class),
+            resolve(Filesystem::class),
+        ])->makePartial();
+
+        swap(Site::class, $siteMock);
+
+        $config->shouldReceive('read')->andReturn([
+            'tld' => 'test',
+            'loopback' => VALET_LOOPBACK,
+            'paths' => ['/Users/name/code'],
+        ]);
+
+        $files->shouldReceive('isLink')->andReturnUsing(function ($path) {
+            return $path === VALET_HOME_PATH.'/Sites/blog';
+        });
+        $files->shouldReceive('readLink')->andReturn('/Users/name/deleted/blog');
+        $files->shouldReceive('isDir')->andReturnUsing(function ($path) {
+            return $path === '/Users/name/code/blog';
+        });
+        $files->shouldReceive('realpath')->andReturnUsing(function ($path) {
+            return $path;
+        });
+
+        $site = resolve(Site::class);
+
+        // A stale link must not mask a parked site of the same name.
+        $this->assertEquals('/Users/name/code/blog', $site->getSitePath('blog'));
     }
 
     public function test_it_throws_getting_nonexistent_site()
```

And your shadowing test, which currently can't fail — I checked by forcing `$isExplicitPath = true`, i.e. restoring the directory-first precedence it's named after, and it stayed green: an empty shadow directory and the fixture both resolve to `BasicValetDriver` with the same site name. Giving the shadow directory a `public/index.php` makes the two resolutions differ. This version also fixes a smaller thing — the current `if (! file_exists) mkdir` / `finally { if (file_exists) rmdir }` pair will delete a pre-existing `my-best-site/` in the working directory that the test didn't create (and `rmdir` on a non-empty one raises a warning, which `convertWarningsToExceptions="true"` turns into an error):

```diff
diff --git a/tests/CliTest.php b/tests/CliTest.php
index 36cee66..5c96958 100644
--- a/tests/CliTest.php
+++ b/tests/CliTest.php
@@ -574,20 +574,28 @@ public function test_which_command_prefers_registered_site_over_cwd_subdirectory
 
         Configuration::addPath(__DIR__.'/fixtures/Parked/Sites');
 
-        $dummySubdir = getcwd().'/my-best-site';
-        if (! file_exists($dummySubdir)) {
-            mkdir($dummySubdir);
-        }
+        // A directory in the working directory that shadows the registered site's
+        // name, with contents that would select a *different* driver.
+        $shadowPath = __DIR__.'/output/shadow-'.uniqid();
+        mkdir($shadowPath.'/my-best-site/public', 0755, true);
+        touch($shadowPath.'/my-best-site/public/index.php');
+
+        $originalCwd = getcwd();
+        chdir($shadowPath);
 
         try {
             $tester->run(['command' => 'which', 'site' => 'my-best-site']);
             $tester->assertCommandIsSuccessful();
 
-            $this->assertStringContainsString('The [my-best-site] site is served by [', $tester->getDisplay());
+            // The registered site is a plain directory, so resolving it gives the
+            // basic driver; the shadowing directory would have given the public one.
+            $this->assertStringContainsString(
+                'The [my-best-site] site is served by [Valet\Drivers\BasicValetDriver]',
+                $tester->getDisplay()
+            );
         } finally {
-            if (file_exists($dummySubdir)) {
-                rmdir($dummySubdir);
-            }
+            chdir($originalCwd);
+            exec('rm -rf '.escapeshellarg($shadowPath));
         }
     }
```

</details>

## Caveats on the above, so you're not surprised

- **Behaviour change:** `valet which my-best-site/` and `valet which ./my-best-site` now report "could not find" instead of resolving as names (today they resolve and print `The [./my-best-site] site ...`). I think treating a slash as "this is a path" is right, but it's your call — `rtrim($site, '/')` before dispatch keeps the trailing-slash case working.
- With the guard in place, `which`'s third branch is unreachable (both inputs above fall through to the `else`), so it and the first branch can be collapsed. I left that out of the patch to keep it focused.
- `getSitePath()` is public. In-repo callers are the only ones I can see, but anything third-party passing a path rather than a name would now get `null`.
- I ran the suite on 8.4 and the new tests on 8.5, not on 8.0-8.2 — the installed PHPUnit needs >= 8.3 and there's no committed lock file, so CI resolves a different PHPUnit there. Syntax is checked across the matrix; behaviour is verified on 8.4/8.5.
- The seven `CliTest` failures I still see locally (`test_link_command`, `test_parked_command`, the isolate ones) fail on `master` here too — they need a linked Homebrew PHP, which my older machine doesn't have. I'm inferring they're green on CI from `master` being green, not observing it.

